PR #5, opened 2021-12-23
Hotfix/fix up variant editing screens

alecritson:
### Changes

- On the option selector, if you have no options available it would suggest there was no options due to the search term, which is incorrect. This should now clarify why there are no options a bit better.
- The "add option" button could be easily missed on the option selector, it's now next to the search box.


glennjacobs:
Can we also update the placeholder text as we discussed? And hide the translation button if only one language set?

image

PR #39, opened 2022-01-07
Feature - Account area

alecritson:
This PR adds a new screen for the current staff member to be able to edit their own details.

### Functionality

- Ability to update their firstname, lastname and email address
- Adds a new `password` field type so the user can toggle visibility as they are typing
- The new password requires their current password to be provided in order to change (if forgotten they should do a password reset)

### Additional points

- Updated default Gravatar to use the mystery user


glennjacobs:
Looks good, but,...

If I update my name or email, the name and Gravatar doesn't update in the bottom left of the UI until refreshing the page. I guess it should...

alecritson:
I think we'd need to create a component in the bottom of the UI with a computed property for the avatar which would react to the change since they're not connected in any way 🤔

alecritson:
@glennjacobs Pushed some updates to hopefully resolve the comments above.

glennjacobs (approved):

PR #60, opened 2022-01-11
Create edit form for tags

briavers:
This should fix https://github.com/getcandy/getcandy/issues/57 and add some functionality. I've based myself of Currency and Attribute items to try and follow the general flow. 

This is my first PR to an open source project, so please be gentle but correct.

If you require some tests I'm okay with writing them, just let me know


packages/admin/src/Http/Livewire/Pages/Settings/Tags/TagShow.php
```diff
@@ -18,7 +18,7 @@ public function render()
     {
         return view('adminhub::livewire.pages.settings.tags.show')
             ->layout('adminhub::layouts.settings', [
-                'title' => 'View Tag',
+                'title' => 'Edit Tag',
```
glennjacobs, line 21: This would need to go into the language files, thanks.

glennjacobs:
Thanks. We'll give this a proper review tomorrow. 👍

glennjacobs:
I've pulled this down and tested, and it looks good. Just the missing translation required I believe.

alecritson:
On the tag field-type we are forcing the value to be uppercase when it's added. Should we have something on the model / an observer to ensure the same?

glennjacobs:
> On the tag field-type we are forcing the value to be uppercase when it's added. Should we have something on the model / an observer to ensure the same?

I thought about that, but I don't think it's required. Changing the case doesn't break anything and some might want to preferably have tags cased a certain way.

glennjacobs:
@briavers are you able to sort the translation so we can merge it? Thanks.

briavers:
I'll update the translations tomorrow

glennjacobs (approved):

glennjacobs:
@briavers thanks for the contribution 👍 . We've merged into main and are performing final testing before we tag the next release.

Issue #61, opened 2022-01-13
Problem if I using UUIDs or GUIDs in my User model


devmansurov:
How I can use getcandy if I use UUIDs or GUIDs in my User model?

glennjacobs:
At the moment, you can't. However, you raise a good point and I can't imagine you'd be the only person with this set-up.

We'll take a look and see what can be done.

glennjacobs:
I've created the PR mentioned above. Be good if you could review it and let us know if it fixes the issue for you.

devmansurov:
Add foreign key customization [here](https://github.com/getcandy/core/blob/main/database/migrations/2021_10_01_090000_create_orders_table.php#L14) too

`$table->userForeignKey();`

devmansurov:
You must add `userForeignKey()` everywhere where use `user_id`

PR #62, opened 2022-01-13
Add the ability to change foreign key to users table

glennjacobs:
This is a fix for #61 where a Laravel app is using a different field type for the `users` table primary key. E.g. `uuid` or `int` as opposed to `bigint`.

You can now set in the config which field type to use. Automatically detecting the field type was considered, but came with some risks and additional packages to install, which wasn't deemed worth it.


packages/core/src/GetCandyServiceProvider.php
```diff
@@ -184,4 +171,43 @@ protected function registerObservers(): void
...
+            foreach ($columns as $column) {
+                $this->decimal("{$column}_value", 10, 4)->default(0)->nullable()->index();
+                $this->string("{$column}_unit")->default('mm')->nullable();
+            }
+        });
+
+        Blueprint::macro('userForeignKey', function ($field_name = 'user_id') {
+            $userModel = config('auth.providers.users.model');
+
+            if (config('getcandy.database.users_id_type') == 'uuid') {
```
alecritson, line 199: For readability would it be better to do something like:

```php
$type = config('getcandy.database.users_id_type', 'bigint')

if ($type == 'uuid') {
}
```

alecritson (approved):

alecritson (approved):

PR #70, opened 2022-01-18
Use start and end dates for channel associations

alecritson:
This PR looks to change the way models are scheduled against channels.

Currently for customer groups we have a `starts_at` and `ends_at` column to determine when the scheduling should start and end. For channels, this isn't the case and relies on a `published_at` column.

This PR will add the `starts_at` and `ends_at` columns so they get the same treat and adds consistency across models that can be scheduled.


docs/src/upgrading.md
```diff
@@ -30,6 +30,27 @@ If you're using Meilisearch, run the following
 php artisan getcandy:meilisearch:setup
 ```
 
+## 2.0-beta6
```
glennjacobs, line 33: Should be "Unreleased" really, but will let it slide as we're looking to do a release tomorrow.

glennjacobs: I checked, and the next release would be 2.0-beta7 - so this is wrong.

glennjacobs (approved):

PR #82, opened 2022-01-21
Change slugs to use underscore https://github.com/getcandy/getcandy/issues/81

itcyborg:
Update how handle is generated when creating an attribute. This is to prevent issues when parsing the attributes in javascript without adding additional mapping using php.
case : suppose we want to add an attribute named "warranty information", the handle generated will be 'warranty-information'. this commit attempts to fix this by replacing '-' with '_'


alecritson:
Looks like this is doing the same as #83 but just on different files. Going to close this and suggest the changes go on to that PR so it's all in one place to review.

itcyborg:
This is ok. Was working on a tab and completely forgot about vscode online.


Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows

PR #83, opened 2022-01-21
Change slugs to use underscore

itcyborg:
Update how handle is generated when creating an attribute. This is to prevent issues when parsing the attributes in javascript without adding additional mapping using php.
case : suppose we want to add an attribute named "warranty information", the handle generated will be 'warranty-information'. this commit attempts to fix this by replacing '-' with '_'

### DIFF
--- a/packages/admin/src/AdminHubServiceProvider.php
+++ b/packages/admin/src/AdminHubServiceProvider.php
@@ -68,6 +68,7 @@ use Illuminate\Support\Facades\Event;
 use Illuminate\Support\Facades\Gate;
 use Illuminate\Support\ServiceProvider;
 use Livewire\Livewire;
+use Illuminate\Support\Str;
 
 class AdminHubServiceProvider extends ServiceProvider
 {
@@ -130,6 +131,11 @@ class AdminHubServiceProvider extends ServiceProvider
             RouteMatched::class,
             [SetStaffAuthMiddlewareListener::class, 'handle']
         );
+
+        // Handle generator
+        Str::macro('handle',function($string){
+            return Str::slug($string, '_');
+        });
 
         $this->app->singleton(\GetCandy\Hub\Editing\ProductSection::class, function ($app) {
             return new \GetCandy\Hub\Editing\ProductSection();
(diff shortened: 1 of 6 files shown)


itcyborg:
The snake helper does not change the delimiter if it already exists.
image

itcyborg:
The suggestion can work if the words are separated by spaces.

glennjacobs:
It's likely we may want to use this style of "slugification" in other parts of the system or even addons down the line. 

I suggest we look to add our own string helper to avoid duplication.

e.g. 

```php
      Str::macro('snakeSlug', function (string $value) {
            Str::slug($value, '_')
      });
```

alecritson:
What if it was named something like:

```php
Str::handle($handle);
```

itcyborg:
> What if it was named something like:
> 
> ```
> Str::handle($handle);
> ```

This would make more sense.

glennjacobs:
Yeah, that name sounds good.

packages/admin/src/AdminHubServiceProvider.php
```diff
@@ -131,6 +132,11 @@ public function boot()
             [SetStaffAuthMiddlewareListener::class, 'handle']
         );
 
+        // Handle generator
+        Str::macro('handle', function ($string) {
```
alecritson, line 136: I feel like this should be registered in the core. Other areas of GetCandy will likely benefit from this, not just admin hub code.

alecritson (changes requested):
I appreciate your effort on this and it is looking good. 

Can we get some basic tests in for the macro to make sure it's returning what we expect based on a bunch of different strings that could be passed.

packages/admin/src/AdminHubServiceProvider.php
itcyborg: I have updated as requested. The helper is now in getcandy core package. I've also added some unit test for the helper function. Hope this is well. let me know if further changes are required.

alecritson:
Thanks for sorting @itcyborg I'll get this merged ready for the upcoming release!

alecritson (approved):

PR #86, opened 2022-01-24
feat[Docs]: Add main site link to top navigation

poppabear8883:


glennjacobs:
I can see why this might be handy to have, but at the same time, we have more content coming soon to the docs which may need to use the space available in the top menu. So I'll leave this open until we decide.

poppabear8883:
Fair enough

glennjacobs:
Closing this for the moment. I'm going to be reworking the docs, so it's something I'll consider then. Thanks for the suggestion though.

PR #87, opened 2022-01-24
patch[Docs]: Remove duplicated editLinks prop

poppabear8883:


alecritson:
Thanks for spotting this and doing a PR!

PR #89, opened 2022-01-25
Bug: Don't assume user has a customer associated yet

ryanmitchell:
Installed into an existing Laravel app and the trying to `CartSession::current()` I got an error on this line as my user (logged in) had no customer associated with it (yet).

This fix should cover that offer.

### DIFF
--- a/packages/core/src/GetCandyServiceProvider.php
+++ b/packages/core/src/GetCandyServiceProvider.php
@@ -94,9 +94,10 @@ class GetCandyServiceProvider extends ServiceProvider
         $this->registerObservers();
         $this->registerAddonManifest();
         $this->registerBlueprintMacros();
-        
+
         config('auth.providers.users.model')::resolveRelationUsing('customers', function ($userModel) {
             $prefix = config('getcandy.database.table_prefix');
+
             return $userModel->belongsToMany(
                 Customer::class,
                 "{$prefix}customer_user"
--- a/packages/core/src/GetCandyServiceProvider.php
+++ b/packages/core/src/GetCandyServiceProvider.php
@@ -29,7 +29,6 @@ use GetCandy\Models\CartLine;
 use GetCandy\Models\Channel;
 use GetCandy\Models\Collection;
 use GetCandy\Models\Currency;
-use GetCandy\Models\Customer;
 use GetCandy\Models\Language;
 use GetCandy\Models\OrderLine;
 use GetCandy\Models\Url;
@@ -94,15 +93,6 @@ class GetCandyServiceProvider extends ServiceProvider
         $this->registerObservers();
         $this->registerAddonManifest();
         $this->registerBlueprintMacros();
-
-        config('auth.providers.users.model')::resolveRelationUsing('customers', function ($userModel) {
-            $prefix = config('getcandy.database.table_prefix');
-
-            return $userModel->belongsToMany(
-                Customer::class,
-                "{$prefix}customer_user"
-            )->withTimestamps();
-        });
 
         if (!$this->app->environment('testing')) {
             $this->registerStateListeners();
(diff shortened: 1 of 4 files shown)


glennjacobs:
Thanks for the contribution, we'll take a look.

glennjacobs:
@alecritson as this is in the core, do we need a test added to ensure it can't crop up again?

alecritson:
@glennjacobs Likely would a test and also i'm thinking wouldn't the `customers` relationship also return a `Collection` instance so that check would always pass?

Should it be `$user->customers->count()` ?

glennjacobs:
> @glennjacobs Likely would a test and also i'm thinking wouldn't the `customers` relationship also return a `Collection` instance so that check would always pass?
> 
> Should it be `$user->customers->count()` ?

I think you're correct, as we can't be sure the `->customers` relationship has been loaded at this point. 

@ryanmitchell are you OK to update and add in a suitable test?

joelwmale:
@alecritson @glennjacobs @ryanmitchell I'd recommend updating the chick to be this:

```
$this->customerGroups = $cart->user && $cart->user->customers->isNotEmpty() ?
```

ryanmitchell:
It wasn’t a collection for me - it was NULL which is why I took the approach I did. 

Happy to update as you see fit - but i don’t think the collection count approach solves it.

glennjacobs:
I think the best thing to do here is to include tests to prove it works in all scenarios, and of course, we'll need those to merge anyhow.

My understanding is that the recommended approach to check if an unloaded relationship exists is to use `->count()` or `->exists()`. I've not tried `->isNotEmpty()` before, but that's a Collection method not an Eloquent method.

As we're dealing with the Cart here, it's very important we get this fix right, so we can't make any assumptions.

ryanmitchell:
No problem. Leave it with me and I'll pick it up early next week.

glennjacobs:
> No problem. Leave it with me and I'll pick it up early next week.

FYI, we'll be preparing beta8 on Monday and it would be good to have this fix included.

ryanmitchell:
Ok, I found the actual cause of the issue is that the User -> customers relation wasn't being defined. 

Therefore I modified the service provider to define a dynamic relation, as otherwise you'd be expecting every install to do that which doesn't seem right.

alecritson:
@ryanmitchell We have a trait with that relationship already set up. In case we need to add any further relations and without putting them in the SP.

So, am I right in now thinking your user model doesn't use this: https://github.com/getcandy/getcandy/blob/main/packages/core/src/Base/Traits/GetCandyUser.php

ryanmitchell:
Yes - I don't think it's in the docs (or at least I didn't see it) so I didn't add the trait.

Is a trait the best approach? Seeing as it's so integral to the order process I do feel it should be automatically added.

glennjacobs:
image

Whoopsie!

ryanmitchell:
At least it makes sense of why you weren't seeing it and I was. I'll revert the SP changes but the tests are still useful I think.

I'll also update the user stub to use that trait.

glennjacobs:
We've patched the docs, thanks for highlighting that 
https://docs.getcandy.io/installation.html#add-the-getcandyuser-trait

joelwmale:
> At least it makes sense of why you weren't seeing it and I was. I'll revert the SP changes but the tests are still useful I think.
> 
> I'll also update the user stub to use that trait.

If @glennjacobs or @alecritson can merge in my fix for the failing test - my fix has this trait added to the user stub, which might resolve this issue also?

alecritson:
> > At least it makes sense of why you weren't seeing it and I was. I'll revert the SP changes but the tests are still useful I think.
> > I'll also update the user stub to use that trait.
> 
> If @glennjacobs or @alecritson can merge in my fix for the failing test - my fix has this trait added to the user stub, which might resolve this issue also?

Hey @joelwmale Your PR for the fix has been merged into the next release :)

PR #92, opened 2022-01-26
Bug: dashboard placed_at throws error when null

ryanmitchell:
This PR wraps it in `optional()`


alecritson:
We haven't used optional in the code before this. What's your thoughts on using `$order->placed_at?->format('jS F Y h:ma')` instead? @glennjacobs @ryanmitchell

glennjacobs:
> We haven't used optional in the code before this. What's your thoughts on using `$order->placed_at?->format('jS F Y h:ma')` instead? @glennjacobs @ryanmitchell

Have we used `?` in other views then?

alecritson:
Not so much in other views but it has been used in code in GetCandy.

alecritson:
After chatting we're going to go with `optional`

PR #103, opened 2022-01-27
Hotfix - Stop charts duplicating on the dashboard

### LINKED ISSUE #100: [Hub] Dashboard UI is having issues with the charts
When you first load the dashboard the `Sale Performance` chart is bugged. Simply resizing the window seems to cause it to fix itself ?

chrome_3B7jCKbv8O

https://user-images.githubusercontent.com/7263458/151368568-cde0d3aa-cd20-4d16-926c-dee533a28d67.mp4

ApplicationFrameHost_wL2P8HvGSo

alecritson:
Fixes #100 

The upgrade with Alpine means that `init` gets called automatically so we're in fact calling it twice at the moment, resulting in a weird graph duplication visual.

### DIFF
--- a/packages/admin/resources/views/livewire/components/reporting/apex-charts.blade.php
+++ b/packages/admin/resources/views/livewire/components/reporting/apex-charts.blade.php
@@ -1,15 +1,14 @@
 <div
     style="width: 100%; height: 100%;"
     x-data="{
-    init() {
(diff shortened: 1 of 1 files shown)


itcyborg:
Thanks for raising this, thought it was ma customizations that was causing it

PR #106, opened 2022-01-27
feat[Core]: Add base_price accessor to Product model

poppabear8883:
Usage example: `$product->base_price->formatted`

```blade
<product-card name="{{$product->translateAttribute('name')}}"
                        price="{{$product->base_price->formatted}}"
                        image=""
></product-card>
```

The way it currently is used in the hub looks something like this and this PR can make this approach cleaner:
```blade
@php
    $price = $variant->basePrices->first(fn($price) => $price->currency->default);
@endphp
{{ $price->price->formatted }}
```

### DIFF
--- a/packages/core/src/Models/Product.php
+++ b/packages/core/src/Models/Product.php
@@ -118,6 +118,17 @@ class Product extends BaseModel implements SpatieHasMedia
     }
 
     /**
+     *
+     * @return int
+     */
+    public function getBasePriceAttribute()
+    {
+        return $this->variants->first()
+            ->basePrices->first(fn($price) => $price->currency->default)
+            ->price->value;
+    }
+
+    /**
      * Associate a product to another with a type.
      *
      * @param mixed  $product
--- a/packages/core/src/Models/Product.php
+++ b/packages/core/src/Models/Product.php
@@ -118,13 +118,14 @@ class Product extends BaseModel implements SpatieHasMedia
     }
 
     /**
+     * Returns the base price of the product.
      *
      * @return int
      */
     public function getBasePriceAttribute()
     {
         return $this->variants->first()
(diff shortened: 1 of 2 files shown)


glennjacobs:
OK, so you're looking for a convenient way to get the base price for a product (not a variant). I don't think this is the correct approach. I will discuss with @alecritson tomorrow and make suggestions.

poppabear8883:
I was under the impression that a `product` with `no variants` still uses a `variant` under the hood for the price so I was assuming that the variants collection would return the base variant and that's what would own the base price. Now that I have `tinkered` with it I can see that the "under the hood variant" is not included in the products variants collection.

I'm trying to reverse engineer the *hub* to see what approaches your team has taken.

I am also noticing that when i initially create a product with variants it does ask for a `Base Price` ... then I save the product and then I can edit the variant pricing, but the base price is no longer there ? Is this by design ?

glennjacobs:
> I am also noticing that when i initially create a product with variants it does ask for a `Base Price` ... then I save the product and then I can edit the variant pricing, but the base price is no longer there ? Is this by design ?

Sounds odd. I'll review with Alec in the morning. But regardless, I'm not keen on pricing functionality on something that isn't purchasable. We also have the Pricing Manager coming in beta8 which will no doubt likely cover your requirements and more.

poppabear8883:
> But regardless, I'm not keen on pricing functionality on something that isn't purchasable.

Its not really "pricing functionality". Its just an Accessor ... to help display data in a more elegant way.

> We also have the Pricing Manager coming in beta8 which will no doubt likely cover your requirements and more.

Awesome, I'm ok with just closing this PR if you find that there is no need for it.

glennjacobs:
Closing as I believe the new Pricing Manager functionality will make this requirement redundant.

PR #108, opened 2022-01-27
Feature: Fix tests when running from core

joelwmale:
Tests are currently failing when running from within the `core` directory. This PR fixes the single test that fails.


alecritson:
@joelwmale Thanks for sorting this out. Looks cleaner, will get it merged into the release branch.

PR #121, opened 2022-02-01
Prevent pricing saving twice

alecritson:
When creating a product, we need to save the pricing initially in case we're generating variants as well. However this was being called twice so this PR adds an additional check to prevent that.


glennjacobs (approved):

glennjacobs:
Do we want to consider a test for this?

glennjacobs (approved):

PR #122, opened 2022-02-01
Prevent duplicate cart lines

alecritson:
This PR looks to tweak the cart line check when adding an item.

Currently if either one is null but not the other, it doesn't match and results in a duplicate cart line being added. This PR defaults both to an empty array `[]` when there is no value.


glennjacobs:
Should we not have a test for this fix?

glennjacobs (approved):

PR #124, opened 2022-02-01
Description should not be required

alecritson:
Tweaked the installer so description field is no longer required on install.


alecritson:
After a bit of discussion on Discord. If we're not making it required, should this be a system attribute?

glennjacobs:
> After a bit of discussion on Discord. If we're not making it required, should this be a system attribute?

Let's just make it a normal attribute that isn't required. It will be a good example of a standard attribute then.

glennjacobs (approved):

PR #127, opened 2022-02-01
feat[Core]: Add collections relationship to product model

### LINKED ISSUE #126: Product model missing collections() BelongsToMany relationship
We should be able to get the collections the product belongs to.

poppabear8883:
Adds `$product->collections` relationship

closes #126

### DIFF
--- a/packages/core/src/Models/Product.php
+++ b/packages/core/src/Models/Product.php
@@ -107,6 +107,16 @@ class Product extends BaseModel implements SpatieHasMedia
     }
 
     /**
+     * Return the product collections relation.
+     *
+     * @return \Illuminate\Database\Eloquent\Relations\BelongsToMany
+     */
+    public function collections()
+    {
+        return $this->belongsToMany(Collection::class, 'getcandy_collection_product');
+    }
+
+    /**
      * Return the associations relationship.
      *
      * @return \Illuminate\Database\Eloquent\Relations\HasMany
--- a/packages/core/tests/Unit/Models/ProductTest.php
+++ b/packages/core/tests/Unit/Models/ProductTest.php
@@ -3,11 +3,13 @@
 namespace GetCandy\Tests\Unit\Models;
 
 use GetCandy\Models\Channel;
+use GetCandy\Models\Collection;
 use GetCandy\Models\CustomerGroup;
 use GetCandy\Models\Product;
 use GetCandy\Models\ProductAssociation;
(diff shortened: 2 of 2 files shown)


packages/core/src/Models/Product.php
```diff
@@ -106,6 +106,16 @@ public function variants()
...
     }
 
+    /**
+     * Return the product collections relation.
+     *
+     * @return \Illuminate\Database\Eloquent\Relations\BelongsToMany
+     */
+    public function collections()
+    {
+        return $this->belongsToMany(Collection::class, 'getcandy_collection_product');
```
glennjacobs, line 116: Table prefix shouldn't be hardcoded.

poppabear8883: yep I'm fixing this now

poppabear8883: Ref: 311bc8c

glennjacobs:
@alecritson do we need to consider the pivot table on this?

alecritson:
I think we should include the pivot on this in case the developer needs to get the position of the product in relation to the collection. Something like `->withPivot(['position']);` Then I think we should be good to get this merged in for the next release 🙏  Thanks for the contribution!

packages/core/tests/Unit/Models/ProductTest.php
```diff
@@ -437,4 +439,18 @@ public function can_only_remove_associations_of_a_certain_type()
...
+    public function can_have_collections_relationship()
+    {
+        $collection = Collection::factory()->create();
+        $product = Product::factory()->create();
+        $product->collections()->sync($collection);
+
+        $this->assertInstanceOf(EloquentCollection::class, $product->collections);
+        $this->assertCount(1, $product->collections);
+        $this->assertInstanceOf(Collection::class, $product->collections->first());
+        $this->assertArrayHasKey('pivot', $product->collections->toArray()[0]);
```
alecritson, line 453: Is there a reason why you went for array access over something like:

```php
$collection->pivot->position
```

poppabear8883: Well the assertion is `ArrayHasKey`.

I suppose I could have just asserted `NotNull`.

I can update this to be consistent with other assertions.

alecritson: Yeah I think that would be better if that's okay, thank you!

PR #135, opened 2022-02-04
Bug fix: don't assume `id` is the user key

ryanmitchell:
Use getKey() instead

### DIFF
--- a/packages/core/src/Listeners/CartSessionAuthListener.php
+++ b/packages/core/src/Listeners/CartSessionAuthListener.php
@@ -39,7 +39,7 @@ class CartSessionAuthListener
 
         if (! $currentCart) {
             // Does this user have a cart?
-            $userCart = Cart::whereUserId($event->user->id)->latest()->first();
+            $userCart = Cart::whereUserId($event->user->getKey())->latest()->first();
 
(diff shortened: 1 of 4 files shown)


alecritson:
Thanks @ryanmitchell , does this open more questions... Since the `user_id` column is constrained, if it's not the `id` column on the `users` table then `whereUserId()` would be incorrect to no? 🤔

ryanmitchell:
So the use case is logging in via Statamic using the user:login form - it provides a `Statamic\Auth\Eloquent\User` instance which is an abstraction - its not a `App\Models\User` instance.

getKey() works in this scenario and under conditions too.

alecritson:
Thanks, I think it'd be good to change to `getKey` although there are some other places that would need to be changed I think where we use `$user->id` so I think these should be updated to be consistent. I'm happy to do that and then merge this when done or it can be part of this PR?

When I had a look there isn't many places that I could see, let me know what you'd prefer.

image

ryanmitchell:
No problem - I've updated them all

alecritson (approved):
Thank you for sorting this! I'll get this merged then if you want to use it before the release you can target the main branch.

PR #141, opened 2022-02-07
replaced strpos and magic accessor

f4bio:
* replaced `strpos()` with `str_contains()` as suggested as of php8.0 (some explanation [here](https://blakey.co/blog/goodbye-strpos-and-strstr:-str_contains-in-php8))
* using "correct" route name accessor `getName()`

### DIFF
--- a/packages/admin/src/Listeners/SetStaffAuthMiddlewareListener.php
+++ b/packages/admin/src/Listeners/SetStaffAuthMiddlewareListener.php
@@ -9,17 +9,17 @@ class SetStaffAuthMiddlewareListener
     /**
      * Handle the event.
      *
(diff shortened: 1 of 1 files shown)


packages/admin/src/Listeners/SetStaffAuthMiddlewareListener.php
```diff
@@ -9,17 +9,17 @@ class SetStaffAuthMiddlewareListener
     /**
      * Handle the event.
      *
-     * @param  \Illuminate\Routing\Events\RouteMatched  $event
+     * @param  RouteMatched  $event
```
alecritson, line 12: Is there a reason the fully qualified name was removed?

alecritson (approved):

PR #143, opened 2022-02-07
Feature/env setup

f4bio:
[ff](https://github.com/getcandy/getcandy/pull/142) 

---
_(I hope I used that correctly?! :D)_


glennjacobs:
You are changing multiple things in a single PR. Please, if you wish to contribute, keep PRs to a single theme and include a description of what the changes are and why you think they should be considered.

However, I would always advise opening an issue before spending time on a PR, as there's a fair chance that your hard work may be rejected.

Issue #144, opened 2022-02-08
Configurable models


edcoreweb:
What's your opinion on making the models configurable, or offering a way to extend / overwrite models?

Ran into a issue where I needed to add a new relation on the product model, and the way things are now, does not make it very easy to do that.

Most of the packages list the models in the config file, so they can be replaced if needed. Is that something that is planned, and would you accept a PR, making models configurable?

glennjacobs:
We suggest https://laravel.com/docs/8.x/eloquent-relationships#dynamic-relationships

edcoreweb:
Hmm, that makes sense for relations indeed, but for other methods? Maybe it would be a good idea to make the models macroable? What do you think?

glennjacobs:
Interesting idea. A quick Google found these packages which could potentially be used. 

https://github.com/spatie/macroable
https://github.com/javoscript/laravel-macroable-models

glennjacobs:
I think if we implemented the Spatie package and included the trait in the `GetCandy\Base\BaseModel` that could work nicely.

Happy for you to do a PR, as long as it includes suitable tests ;-)

glennjacobs:
Merged into `main` :-) So you can now use this by either requiring `dev-main` or using the monorepo.

And of course it will be included in the next beta release.

edcoreweb:
Great! Thanks a lot!

glennjacobs:
> Great! Thanks a lot!

Thanks for the contribution!

PR #145, opened 2022-02-08
Add macro support to all models

edcoreweb:
See #144 

Laravel already has a built in Macroable trait, used that instead of the **spatie/macroable**, as no extra dependencies are needed.


glennjacobs:
@alecritson should we consider some docs to go alongside this?

alecritson:
Docs would be useful, just where the best place is?

PR #149, opened 2022-02-09
Fix return types in doc blocks

KKSzymanowski:
This fixes most of the issues with `@return` annotations in doc blocks.

One place I don't know how to fix is `GetCandy\FieldTypes\Dropdown`.
The `value` property is declared as `string`.
The constructor parameter which is assigned to the `value` property is declared as `int|float`.
The `getValue()` method which returns the `value` property is declared as `array`.
What should be the type of the `Dropdown::$value` property?


alecritson:
Hi @KKSzymanowski Thanks for this. Is there any reason why you removed the fully qualified class names from the doc blocks? My concern with this is that some systems might not pick up the alias and it's also not immediately obvious when looking just at the comment what is expected.

With regards to the drop-down, looks like there are some hang-ups in there. It should be returning a `string|int`.

KKSzymanowski:
If you're referring to these changes:
image
I assumed that if the Collection class has already been imported I can just use the class name instead of the fully qualified name.
I can revert these changes if you want.

alecritson:
> If you're referring to these changes: image I assumed that if the Collection class has already been imported for the type hint in the argument list I can just use the class name instead of the fully qualified name. I can revert these changes if you want.

That's it yeah, I think using the full name is better as it means we don't have to import classes we might not actually be using and also some IDE's might not see/use the alias

KKSzymanowski:
I ran a different inspection in PhpStorm and found a lot of `@param` annotations which had nothing to do with the method signature so I fixed them here as well.
I also fixed the Dropdown according to your suggestion and stuck everywhere with the FQCNs.

This turned out to be much bigger of a PR than I initially expected but it shouldn't change behaviour in any way.

alecritson (approved):

PR #157, opened 2022-02-09
Fix rich text attribute creation in getcandy:install command

KKSzymanowski:
The install command creates default attributes with the wrong configuration.
For rich text it is
```json
{
    "type": "richtext"
}
```
instead of
```json
{
    "richtext": "1"
}
```

and for regular text fields it is:
```json
{
    "type": "text"
}
```
instead of
```json
{
    "richtext": null
}
```

### DIFF
--- a/packages/admin/resources/views/field-types/text/settings.blade.php
+++ b/packages/admin/resources/views/field-types/text/settings.blade.php
@@ -4,5 +4,5 @@
   :error="$errors->first('attribute.configuration.richtext')"
   :disabled="!!$attribute->system"
 >
-  <x-hub::input.toggle :disabled="!!$attribute->system" wire:model="attribute.configuration.richtext" id="fieldType" />
+  <x-hub::input.toggle :disabled="!!$attribute->system" :on-value="true" :off-value="false" wire:model="attribute.configuration.richtext" id="fieldType" />
 </x-hub::input.group>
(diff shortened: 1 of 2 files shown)


KKSzymanowski:
I went with the same initial configuration to that which comes out of the attributes settings. In my opinion the `richtext` configuration should be either `true` or `false` or maybe `1` or `0`, but the Admin Hub saves it as `'1'` or `null` so that's why I did it this way.

alecritson:
Thanks @KKSzymanowski I agree they should be `true`/`false` so looks like it's not working as intended if it's doing `"1"`/`null`. I'll take a look unless you're happy to add that as part of this PR.

KKSzymanowski:
Looks like it's enough to explicitly define `on-value` and `off-value` for the rich text toggle in 
`packages/admin/resources/views/field-types/text/settings.blade.php`:

```diff
<x-hub::input.group
  label="Richtext"
  for="richtext"
  :error="$errors->first('attribute.configuration.richtext')"
  :disabled="!!$attribute->system">
-    <x-hub::input.toggle :disabled="!!$attribute->system" wire:model="attribute.configuration.richtext" id="fieldType" />
+    <x-hub::input.toggle :disabled="!!$attribute->system" :on-value="true" :off-value="false" wire:model="attribute.configuration.richtext" id="fieldType" />
</x-hub::input.group>
```

Shouldn't `true`/`false` be the default `on-value` and `off-value` for the toggle, though?

alecritson (approved):
Thanks for sorting this out, will get it ready for the next release.

PR #158, opened 2022-02-10
Bug: don't assume unit quantity is 1

ryanmitchell:
When removing tiered pricing items don't assume the base unit is 1, as its editable

(unless I've misunderstood how you see unit_quantity working)

### DIFF
--- a/packages/admin/src/Actions/Pricing/UpdateTieredPricing.php
+++ b/packages/admin/src/Actions/Pricing/UpdateTieredPricing.php
@@ -42,7 +42,7 @@ class UpdateTieredPricing
                 }
             }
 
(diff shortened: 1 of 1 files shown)


glennjacobs:
> I think I'm following you, so ...
> 
> If we have a product that has a unit quantity of say 10, that means we can only buy in blocks of 10 (ie 10, 20, 30 for clarity). Is that how you see it working?
> 
> So then to apply different tiered pricing for 30, you would enter a lower limit of 3 (as thats 3 x 10 units). Is this correct?

Sounds like you got it. Some systems call it a "factor".

An example from one of our clients is screws. A screw costs less than a penny, so we can't give a single item a proper price. So we instead set the unit quantity to 100 or 1000 and then we can add a price in that will work.

Closing as I don't think this is needed now.

PR #171, opened 2022-02-15
Only calculate unit price if it doesn't already exist

ryanmitchell:
@alecritson here is what I was thinking... basically only calculate the unit price if the unit price is not already an instance of `Price`.

This allows cart line modifiers to set a price at the `calculating()` stage and still benefit from the other calculations such as tax etc

### DIFF
--- a/packages/core/src/Actions/Carts/CalculateLine.php
+++ b/packages/core/src/Actions/Carts/CalculateLine.php
@@ -33,23 +33,27 @@ class CalculateLine
         $purchasable = $cartLine->purchasable;
         $cart = $cartLine->cart;
         $unitQuantity = $purchasable->getUnitQuantity();
+                
+        if (! ($unitPrice = $cartLine->unitPrice) instanceof Price) {
 
-        $priceResponse = Pricing::currency($cart->currency)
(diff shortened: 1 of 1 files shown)


glennjacobs:
Seems OK. I can't think of a better approach right now. Would be good to have a comment to explain why the IF statement is there, and of course include some tests to demonstrate overriding the price and it calculating correctly.

ryanmitchell:
Yep - if you're happy with it as is then I'll do both of those things later today

glennjacobs:
@ryanmitchell is this ready for review?

ryanmitchell:
Yep!

alecritson:
Think this looks alright now. My only last question @glennjacobs is should we be documenting this as it could be a handy thing to know?

glennjacobs:
> > Think this looks alright now. My only last question @glennjacobs is should we be documenting this as it could be a handy thing to know?
> 
> Maybe something could be added here? https://docs.getcandy.io/extending/cart-modifiers.html#overview

I don't think that should hold back this PR. We can tweak the docs later.

alecritson (approved):

PR #176, opened 2022-02-16
Hub - Product associations

alecritson:
this PR looks to add Product associations to the admin hub.

There should be a new panel when editing/creating a product that allows you to select both direct and inverse associations.

When selecting inverse associations, the idea is that you will then be adding the current product you are editing as a direct association to the selected products.

Depending on what type of association you select from the dropdown when adding products, they will all get added with that type but you should be free to change each one as you see fit.


glennjacobs (approved):

glennjacobs:
@alecritson I merged the Collections PR, but this has caused conflicts on this PR, doh!

glennjacobs (approved):

PR #177, opened 2022-02-16
Added AddressObserver and ensures that only one default and billing shipping address per customers exists

nicolalazzaro:
- Ensures that only one default shipping address per customers exists.
- Ensures that only one default billing address per customers exists.

### DIFF
--- a/packages/core/tests/Unit/Observers/AddressObserverTest.php
+++ b/packages/core/tests/Unit/Observers/AddressObserverTest.php
@@ -17,82 +17,56 @@ class AddressObserverTest extends TestCase
     /** @test */
     public function can_only_have_one_shipping_default_per_customer()
     {
-        $customerA = Customer::factory()->create();
+        $customer = Customer::factory()->create();
 
-        $addressADefault = Address::factory()->create([
-            'customer_id' => $customerA->id,
+        $addressA = Address::factory()->create([
+            'customer_id' => $customer->id,
             'shipping_default' => true,
         ]);
 
-        $this->assertTrue($addressADefault->shipping_default);
+        $this->assertTrue($addressA->shipping_default);
 
-        $addressANewDefault = Address::factory()->create([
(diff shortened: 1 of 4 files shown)


nicolalazzaro:
> Can you add some tests for this?

Done

packages/core/tests/Unit/Observers/AddressObserverTest.php
```diff
@@ -0,0 +1,98 @@
...
+            'customer_id' => $customerA->id,
+            'shipping_default' => true,
+        ]);
+
+        $this->assertFalse($addressADefault->refresh()->shipping_default);
+        $this->assertTrue($addressANewDefault->shipping_default);
+
+        $customerB = Customer::factory()->create();
+
+        $addressBDefault = Address::factory()->create([
```
alecritson, line 39: Looks like we're repeating the same test but for a different customer. Was this meant to be a test for updating an existing record since I can't see those checks?

nicolalazzaro: Forgive me, I did not quite understand your request. Anyway yes we are testing another client as you could read. I am adding assertions for the updating event, is that what you are asking?

alecritson: Yes I think so. The current test looks like it's duplicating the create assertions, which I think is not required. I think it would be good instead if we test updating an existing address.

nicolalazzaro: Okay, that is acceptable. I will update the test as soon as possible.

nicolalazzaro: @alecritson do you think is enough now?

alecritson (approved):
Thanks @nicolalazzaro, appreciate the contribution. Merged into main so it's ready for the next release :)

PR #178, opened 2022-02-16
Use register instead of boot for container bindings

edcoreweb:
This is a issue, because if a service provider is loaded before the package service provider, no bindings are available.

Laravel calls register first on all providers and then boot, as such it's recommended all bindings are registered in the register method.


alecritson:
Thanks, @edcoreweb Do you think we should move bits from the `AdminHubServiceProvider` also?

edcoreweb:
@alecritson Sure, moved the admin ones as well.

alecritson (approved):

PR #180, opened 2022-02-17
Allign some comments

nicolalazzaro:


packages/core/src/Models/Product.php
```diff
@@ -153,6 +153,11 @@ public function dissociate($product, $type = null)
         Dissociate::dispatch($this, $product, $type);
     }
 
+    /**
+     * Returns the indexable data for the collection.
```
alecritson, line 157: Should this read:

> Returns the indexable data for the product.

As we're on the product model?

packages/core/src/Models/Customer.php
```diff
@@ -39,7 +39,7 @@ protected static function newFactory(): CustomerFactory
     }
 
     /**
-     * Get the indexable data array for the model.
+     * Returns the indexable data for the collection.
```
alecritson, line 42: I assume this should be `...for the customer` ?

alecritson (approved):

PR #189, opened 2022-02-22
Feat/cart improvements

alecritson:
This PR looks to add a `createOrder` method to the `CartSession` facade.

The reasoning for this is when you want to create an order from a cart, you may wish for the cart session to forget the cart.

Usage:

```php
\GetCandy\Facades\CartSession::createOrder();
```

This will create the order and remove the cart id from the session.

```php
\GetCandy\Facades\CartSession::createOrder(
    forget: false
);
```

This will create the order but retain the cart id in the session.

---

There are some extra tests in this that I felt were missing and should help with the included methods.

### DIFF
--- a/docs/src/getcandy/orders.md
+++ b/docs/src/getcandy/orders.md
@@ -41,7 +41,15 @@ You can either create an order directly, or the recommended way is via a `Cart` 
 $order = \GetCandy\Models\Order::create([/** .. */]);
 
 // Recommended way
-$cart = Cart::first()->getManager()->createOrder();
+$order = Cart::first()->getManager()->createOrder();
+```
+
+If you are using the `CartSession` you can create a cart via the facade, this will then handle removing the cart from the session if you want it to.
+
+```php
+$order = CartSession::createOrder(
+    forget: true
+);
 ```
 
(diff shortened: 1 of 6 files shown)


docs/src/getcandy/orders.md
```diff
@@ -41,7 +41,15 @@ You can either create an order directly, or the recommended way is via a `Cart`
 $order = \GetCandy\Models\Order::create([/** .. */]);
 
 // Recommended way
-$cart = Cart::first()->getManager()->createOrder();
+$order = Cart::first()->getManager()->createOrder();
+```
+
+If you are using the `CartSession` you can create a cart via the facade, this will then handle removing the cart from the session if you want it to.
```
glennjacobs, line 47: "create an order" do you mean?

docs/src/getcandy/orders.md
```diff
@@ -41,7 +41,15 @@ You can either create an order directly, or the recommended way is via a `Cart`
...
+
+```php
+$order = CartSession::createOrder(
+    forget: true
```
glennjacobs, line 51: Isn't `forget` true as default? If so, should we not present the typical usage of this and also describe the option for retaining the cart?

glennjacobs (commented):
Some comments left

glennjacobs (approved):

PR #206, opened 2022-03-02
Fixed translation issue in product variations

lucasvmds:
In the Hub, when selecting an option to create product variants, and the default language is different from "en" an error is displayed.
The same error was also displayed when loading the details of the product or its variants.

This happened because in the templates there were attributes set to "en" and not to translate the values.


alecritson:
Thanks for this @lucasvmds i'll get it merged and ready for the next release 🙏

alecritson (approved):

PR #207, opened 2022-03-03
Feat - Disable variants hub

alecritson:
This PR looks to add a new config option for the hub to disable product variants:

```
'disable_variants' => false,
```

By default variants are enabled, but setting this to `true` will hide variant generation components in the hub, remove the link from the side menu and also will not try and generate them when you save a product.

The `GenerateVariants` job will now also throw an exception if it is called and variants are disabled.

```php
GetCandy\Hub\Exceptions\VariantsDisabledException
```

### DIFF
--- a/docs/src/upgrading.md
+++ b/docs/src/upgrading.md
@@ -32,7 +32,7 @@ php artisan getcandy:meilisearch:setup
 
 ## [Unreleased]
 
-The is a new configraton option under `getcandy-hub/products.php` to disable product variants. This is useful if your storefront will never need to generate different product options and you don't want staff members to be able to do it accidentally.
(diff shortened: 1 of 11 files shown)


docs/src/upgrading.md
```diff
@@ -32,6 +32,22 @@ php artisan getcandy:meilisearch:setup
 
 ## [Unreleased]
 
+The is a new configraton option under `getcandy-hub/products.php` to disable product variants. This is useful if your storefront will never need to generate different product options and you don't want staff members to be able to do it accidentally.
```
glennjacobs, line 35: "There"

glennjacobs (commented):
Just need to check the grammar/spelling in the docs.

glennjacobs (approved):

PR #209, opened 2022-03-06
Fix meilisearch and add command

lucasvmds:
This PR has three points:

1. Correction of Meilisearch installation command;
2. Adjustment of the data being indexed;
3. Added command to index data from multiple models.

Fix Meilisearch install command:
When executing the command getcandy:meilisearch:setup, when creating a new index, an excption is thrown, this happened when creating and already fetching the index soon after. The Collections model was also added, which uses meilsearch to query collections in the product registration.

Adjusting the data being indexed:
Standardization in the names of model indexes, as well as the storage of data for each language in the hub, which I think is essential for multi-language stores.

Added command to index data from multiple models:
With the idea of ​​being able to be installed in a project that is already running, I believe that a command that indexes the existing data for several models is very useful.


alecritson:
Thanks for the contribution @lucasvmds we will take a look at this as soon as possible.

packages/core/src/Console/Commands/ScoutIndexer.php
```diff
@@ -0,0 +1,106 @@
+<?php
```
alecritson, line 1: Can I ask why this has been included, why would you use this command and not just use `php artisan scout:import` command?

packages/core/src/Models/Collection.php
```diff
@@ -104,7 +115,15 @@ public function getSearchableAttributes()
         $data = Arr::except($attributes, 'attribute_data');
 
         foreach ($this->attribute_data ?? [] as $field => $value) {
-            $data[$field] = $this->translateAttribute($field);
+            if ($value instanceof TranslatedText) {
```
alecritson, line 118: I'm not sure we want to be adding all translations into the index. For example, if the index is called `products_en` we would expect only English translations to be in there.

packages/core/src/Models/Product.php
```diff
@@ -199,7 +200,15 @@ public function getSearchableAttributes()
         $data = Arr::except($attributes, 'attribute_data');
 
         foreach ($this->attribute_data ?? [] as $field => $value) {
-            $data[$field] = $this->translateAttribute($field);
+            if ($value instanceof TranslatedText) {
```
alecritson, line 203: Same point as above regarding translations.

packages/core/src/Models/ProductOption.php
```diff
@@ -82,10 +82,10 @@ public function getSearchableAttributes()
     {
         return [
             'id'      => $this->id,
-            'name'    => $this->translate('name'),
+            'name'    => collect(json_decode($this->attributes['name']))->values()->all(),
```
alecritson, line 85: Same point as above regarding translations.

alecritson (commented):
I've had a look at this and although there is some good things in here for sure, there's a few concerns I've made in comments :)

packages/core/src/Console/Commands/ScoutIndexer.php
lucasvmds: Before continuing, let me make a correction:
The fixed models on both MeilisearchSetup and ScoutIndexer commands, should be changed to the `getcandy.indexes` config file (or other config file, but I forgot this). Now, let me explain why this add:

I was using the database engine, but I decided to switch to Meilisearch, so I had to run the `scout:import` command several times to get everything ready. And if any developer wanted to add some existing models to their project for indexing, be it an addon or another (that's why I mentioned the creation of the configuration above), they would need to manually, or in some other way, create the indexes and do the import. Imagine having to do that for ten or twenty models.
In short, adding the missing configuration as well as this command would give developers more freedom and agility to integrate their own models into the store's indexing system.

packages/core/src/Models/Collection.php
lucasvmds: First about the index name:
Some models had the index name as the table name, others with the locale suffix, so I decided to keep all the names with the suffix, but I understand what you said, maybe it's better to change everything to the table name, it's simpler and you do not need to re-index if you change the `app.locale` configuration (of course this won't be done often).

About translations, I think this is very important, maybe not so much for the hub, but for the storefront, especially for multi-language storefronts. If the indexes don't have all the translation options, a multi-language store that uses the indexes to search for products or others will not find results for a language other than the `app.locale` unless it has an index for each language (which I find it much more complicated to implement and would have to change a lot of code).

alecritson: We've had an internal chat about the approach and the one concern we have is document size. Although it would be possible to mark some attributes as not searchable which could potentially solve this issue.

I think it would be better to go for the single index approach so `{prefix}_products` but instead of having the attribute as an array, have each handle suffixed with the locale i.e. `name_en` `name_fr` etc. Just so we keep the values flat and you could easily still target specific languages if you wanted to in search.

What do you think?

packages/core/src/Console/Commands/ScoutIndexer.php
```diff
@@ -0,0 +1,106 @@
...
+     * @param array $models Models array for indexing
+     * @return void
+     */
+    private function indexer (array $models) :void {
+        foreach($models as $model) {
+            $this->newLine();
+
+            // Delete model records from the index
+            $this->warn('Deleting [' . $model . '] records from the index.');
+            $this->call('scout:flush', ['model' => $model]);
```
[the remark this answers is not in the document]
lucasvmds: I still don't have much experience with indexing mechanisms, so I'm sorry if what I'm going to say is something basic: if when importing existing records the indexes are updated, certainly deleting would not be a solution. In that case, it would be better to change the `--flush` option to say that records should be flushed before import and add a `--flush-only` option to only delete records.

alecritson:
Thanks @lucasvmds for all your work on this. The last two (small) points I have are

`searchableAs` Here we are adding an underscore `_` between the config and name i.e.

```php
return config('scout.prefix').'_collections';
```

Whereas Scout's default behaviour doesn't have this, can we remove the `_` just to be consistent with how Scout works by default, I can imagine people ending up having double underscore by accident.

Last point, `config/getcandy/indexer.php` should this be `config/getcandy/search.php`? This config could grow potentially and I feel it would be better suited.

Apart from that solid stuff, really appreciate the time you've taken on this :)

lucasvmds:
I hope everything is ok now.

alecritson:
Hi @lucasvmds Just had a run through and it seems to work pretty well! I noticed that `search` was missing from the config array in the service provider but I've added it since it was a tiny tweak.

I'm pretty happy with this @glennjacobs if you are, be good to get it in!

Thanks again @lucasvmds

packages/core/src/Console/Commands/MeilisearchSetup.php
```diff
@@ -29,39 +25,33 @@ class MeilisearchSetup extends Command
...
     {
+        // Return the models we want to search on.
+        $searchables = config('getcandy.search.models', []);
+
         $this->engine = $engine->createMeilisearchDriver();
 
         // Make sure we have the relevant indexes ready to go.
-        foreach ($this->searchables as $searchable) {
+        foreach ($searchables as $searchable) {
+        // foreach ($this->searchables as $searchable) {
```
glennjacobs, line 42: Need to remove commented out code.

docs/src/getcandy/search.md
```diff
@@ -23,9 +23,33 @@ If you are using meilisearch, there is some additional set up needed. We have a
 php artisan getcandy:meilisearch:setup
 ```
 
+The above command will create the indexes for the models listed in `getcandy/indexer.php` config file. If you want to use other models or your own models in the search engine, you can add the reference for them on the config file.
```
glennjacobs, line 26: "in the config file." would be better.

glennjacobs (commented):
Couple of comments to tidy things up.

alecritson:
@glennjacobs You happy for this to go in?

alecritson:
@glennjacobs Looks like this is set to auto-merge so just needs approval.

glennjacobs (approved):

PR #211, opened 2022-03-08
Feature: hub slots

ryanmitchell:
Here is an initial proof of concept/idea for a way that 'slots' might work on the product page in the hub. 

Everything is handled through a trait to make it easier to add to other areas of the hub (products is my only use case for now, but its probably worth adding everywhere "just in case").

From the developer point of view they would register the slot against the controller they want to use it on as follows:

`\GetCandy\Hub\Http\Livewire\Components\Products\AbstractProduct::registerSlot(\App\Slots\MySlot::class)`

or if you only wanted it on the ProductCreate view:

`\GetCandy\Hub\Http\Livewire\Components\Products\ProductCreate::registerSlot(\App\Slots\MySlot::class)`

The slot itself should inherit from `GetCandy\Hub\Slots\AbstractSlot`, which provides the methods expected by the trait. Maybe there should be a console command and stub to create the slot?

The implementation is agnostic about how the slot content is created in the render method (could be livewire, could be plain blade), to give the developer flexibility. They simply handle the appropriate request data in the slot handleSave method. You may want to tighten this up as it could be too much power to a developer.

Interested to hear your thoughts on this... I haven't added docs, tests etc at this stage as the approach is likely to change.


alecritson:
Thanks @ryanmitchell Just trying to finish up some bits on the order screen updates to get them out the way then I'll have a look at this properly :)

alecritson (commented):
I think it's a good start. When registering slots instead of having to know the whole component class name should we offer simpler syntax, much like we have for registering permissions, cart modifiers etc.

Just thinking off the top of my head...

```php
HubSlot::register('product.create', CustomSlot::class);
```

Each component that supports slots could then have a method to return their slot handle i.e.

```php
public function getSlotHandle()
{
    return 'product.create';
}
```

Then I think we take out passing of `$model` in the constructor. So we could support dependency injection perhaps?

So add a method on the slot to set it.

```php
public function setModel(Model $model);
```

Then to build it up we could do something similar to what you have except:

```php
$this->slots = HubSlot::for('product.create')->get(
  $this->getSlotModel()
);
```

Then behind the scenes the responsibility is shifted back to the class to build out the slot classes and assign the models etc... something like

```php
public function get(Model $model)
{
    // $this->component would be set from calling for(...)
    return collect($this->slots[$this->component])->map(function ($class) use ($model) {
        return app($class)->setModel($model);
    });
}
```

It would also mean then we could be quite explicit where slots can be placed. i.e. if there was two places on product create we wanted to add slots? `product.create.sidebar' `product.create.bottom` or something, they aren't bound to the class consuming them.

Means then we could also have slots which can appear in multiple places around the hub but only need to be registered once? 🤔 

Just my "out loud" thoughts :)

alecritson:
Nice! I'm going to pull this down this morning and see if I can make some slots (I appreciate it's just a concept atm 🤓 )

alecritson:
@ryanmitchell 

Had a little play around with this and it worked pretty well for displaying.

One thing I noticed was the `setModel` method should return itself.

```php
public function setModel($model)
{
    $this->model = $model;

    return $this;
}
```

Now for the tricky part. We reference `$this->slotsByLocation` and `$this->slots`. Interestingly the properties aren't defined on the trait or otherwise. But the problem is...if they are `public` properties, Livewire's hydration will fail when you attempt to save the product.

image

When I've come up against this it's due to a class not implementing `jsonSerialize`, since when Livewire hydrates it converts it to an array to do diffing or w.e. it does...

If you switch the property to `protected` so that it isn't included in the hydration, then changes don't stick and `$this->slots` will be empty when you save the product.. 🤔 

Not entirely sure what the solution is at this point.

ryanmitchell:
> One thing I noticed was the `setModel` method should return itself.
> 
> ```
> public function setModel($model)
> {
>     $this->model = $model;
> 
>     return $this;
> }
> ```

Good spot I'll fix that.

> 
> Now for the tricky part. We reference `$this->slotsByLocation` and `$this->slots`. Interestingly the properties aren't defined on the trait or otherwise. But the problem is...if they are `public` properties, Livewire's hydration will fail when you attempt to save the product.

Yeah they should be public on the trait so that livewire sees them. Did you try that to see if that made a difference?

If youre happy (in principle) with the approach, then I can make some demo slots in livewire and straight blade to work through the issues you've seen and anything else that may come up. Just didnt want to go too far without you buying into the idea.

alecritson:
@ryanmitchell Yeah I think it looks okay from the initial run through. I managed to get slots displaying fine and the view approach seems good as you can still add your own Livewire components I'd assume if you wanted.

@glennjacobs Have you looked much and would you be happy with this approach?

glennjacobs:
> @glennjacobs Have you looked much and would you be happy with this approach?

I've had a quick look and have been listening in on the comments. Sounds like a reasonable approach.

ryanmitchell:
@alecritson I've done some further work on this and now have saving/updating working, however its meant some changes:

1. I've assumed that slots are a livewire component. I think this is reasonable given the hub uses livewire. Due to this I've made the slot methods etc a trait for the component to use. A basic slot example is:

```
<?php

namespace App\Http\Livewire;

use GetCandy\Hub\Slots\Traits\HubSlot;
use Livewire\Component;

class CustomSlot extends Component
{
    use HubSlot;

    public $text = '';

    public function mount()
    {
        $this->test = 'blah de blah';
    }

    public function __construct()
    {
        $this->slotHandle = 'custom-slot';
        $this->slotPosition = 'top';
        $this->slotTitle = 'Custom Slot';
    }

    public function updated()
    {
        $this->raiseSlotErrors(['error' => 1]);
        $this->saveSlotData(['test' => $this->test]);
    }

    public function handleSave($model, $data)
    {
        var_dump($data);
    }

    public function render()
    {
        return view('livewire.custom-slot');
    }
}
```

2. We are doing a bit of a pub/sub type arrangement between the slot components and the page component. Any data the slot wants to 'save' is passed back to a slotstore on the page component, which is then passed back on save. This was a work around when creating new models, as otherwise slot data would not be saved. The slot can also highlight that it has errors so the UI is updated (maybe this should prevent saving?)

At this point I would remove the variables on the slot trait and just get authors to override and return values from getSlotPosition()/Handle/Title. What do you think?

Interested to hear your thoughts now that this is a bit more developed.

packages/admin/src/Http/Livewire/Traits/HasSlots.php
```diff
@@ -3,36 +3,48 @@
...
      */
-    public function getSlotsByLocation($location)
+    public function getSlotsByPosition($position)
     {
         if (! isset($this->slotsForOutput)) {
 
             $this->slotsForOutput = $this->getSlots()
             ->map(function ($slot) {
+
+                $slotComponentName = (string)Str::of(get_class($slot))->afterLast('\\')->snake()->replace('_', '-');
```
alecritson: There should be a method we can call to get this:

```php
$slot->getName();
```

I think we should use this instead since components could be registered in a number of ways and not always relating to the class name.

packages/admin/resources/views/partials/products/editing/sections.blade.php
```diff
@@ -221,13 +225,17 @@ class="@if(!empty($item['has_errors'])) text-red-600 @else text-gray-900 @endif
         </a>
         @endforeach
 
-        @foreach($this->getSlotsByLocation('bottom') as $slot)
+        @foreach($this->getSlotsByPosition('bottom') as $slot)
+            @php
```
alecritson, line 229: Is there any way we can avoid setting PHP variables in the template?

alecritson (commented):
> At this point I would remove the variables on the slot trait and just get authors to override and return values from getSlotPosition()/Handle/Title. What do you think?

Yeah I think that makes sense, we could then offer up an `AbstractSlot` with those kind of methods abstracted so they have to be implemented.

> The slot can also highlight that it has errors so the UI is updated (maybe this should prevent saving?)

Tricky one, since it could potentially become a blocker if it's not implemented correctly. Part of me thinks that slots should handle it gracefully since the data wouldn't be required by the hub/system to function correctly 🤔

packages/admin/src/Http/Livewire/Traits/HasSlots.php
ryanmitchell: Do you mean we should make a method, or that it already exists in Livewire?

packages/admin/resources/views/partials/products/editing/sections.blade.php
ryanmitchell: Could abstract it out to a function in the HasSlots trait if you'd prefer.

packages/admin/src/Http/Livewire/Traits/HasSlots.php
alecritson: This method should already exist in Livewire :)

ryanmitchell:
> > The slot can also highlight that it has errors so the UI is updated (maybe this should prevent saving?)
> 
> Tricky one, since it could potentially become a blocker if it's not implemented correctly. Part of me thinks that slots should handle it gracefully since the data wouldn't be required by the hub/system to function correctly 🤔

I think for now anyway as a first go at slots we can have it not blocking... maybe its something to revisit later down the line.

All other changes made. I'm going to work on a more complicated real world slot this week to make sure everything holds up.

ryanmitchell:
@alecritson this is ready for you to take another look at. I can supply some sample slots off-issue if you want them for easier testing.

alecritson:
Thanks @ryanmitchell It would be good to get an example of once you've been using! Then I could potentially refer to it, will have a look at this as soon as I get a chance :)

alecritson:
@ryanmitchell @glennjacobs 

I've added some initial docs for Slots, anything I've missed or needs changing let me know https://getcandy-2-docs-dpnr6byuy-neondigital.vercel.app/extending/admin-hub.html#slots

PR #212, opened 2022-03-09
Feature - Improved order listing

alecritson:
This PR looks to set the foundation on how we could make listing pages more extensible going forward. 

Tables that deal with large datasets within the hub should meet the following criteria:

- Ability to filter down results based on attributes.
- Be able to search across any attributes indexed in search.
- Developers should be able to add their own columns.
- Developers should have the ability to add custom filters to tables.
- Ability to apply filters and then save those as a "saved search" which appear in tabs at the top.

Each table potentially requires custom search logic, Scout is a little simple in what it offers here and given we want to support different drivers, the aim is to remove any search driver-specific logic into their own classes/methods.

---

## `GetCandy\Hub\Search\OrderSearch`

This is the search class that the listing component calls, it's quite simple and just has specific methods for each driver that fall back to just the default scout method.

The `search` method would return a `SearchResults` Data Transfer Object which is then used by the component to display the results.

## Data Transfer Objects (DTO)

There are a few DTOs that are expected, so no matter what driver is used, there's a consistent response.

### `GetCandy\Hub\DataTransferObjects\Search\SearchResults`

This is a basic class that just accepts a paginated response and a `Facets` DTO.

### `GetCandy\Hub\DataTransferObjects\Search\Facets`

These are the faceted attributes that come back from the search provider. The idea behind this DTO is we don't really know what format they will come back in, as long as we use the DTO the listing page should be able to handle it.

### `GetCandy\Hub\DataTransferObjects\Search\Facet`

This is a single faceted attribute with a collection of facet values i.e. "Brand" with "Nike", "Adidas" etc.

### `GetCandy\Hub\DataTransferObjects\Search\FacetField`

A single representation for a faceted attribute, this DTO accepts the value i.e. "Nike" and count "2" (number of results for this facet)

To put it all together it would look something like this:

```php
$orders = Order::search()->paginate();

$facets = new Facets;

$facetItemA = new FacetItem('Nike', 50);
$facetItemB = new FacetItem('Adidas', 25);

$facet = new Facet('Brand', collect([$facetItemA, $facetItemB]);

$facets->items->push($facet);

$results = new SearchResult(
    $orders,
    $facets
);
```

---

## `GetCandy\Hub\Tables\Orders`

Each listing would have its own underlying table class, this is responsible for determining things like additional columns and filterable attributes (so we can show them in the UI).

> When adding filterable columns they need to be enabled in the search index as filterable via the model's `addFilterableAttributes` method when using the setup command.

The class itself is basic, but it extends the `AbstractTable` class so any further listing pages we improve can use it.

Each table class has its own facade to make it easy to extend via a singleton.

```
GetCandy\Hub\Facades\OrdersTable
```

### Adding columns

```php
addColumn(string $header, bool $sortable = false, Closure $callback = null): TableColumn;

OrdersTable::addColumn('Delivery', 'delivery', function (Order $order) {
    return 'Delivery Info';
});
```

### Adding Filters

You can add columns to the table like so:

```php
addFilter(string $header, string $column, Closure $formatter = null): TableFilter;

OrdersTable::addFilter('Brand', 'brand', function ($value) {
    // This is an optional closure, if you need to format the value returned from the search index.
    return Str::uppercase($value);
});
```

### Exporting Orders

A basic exporter is provided by default, however, you can use your own as you see fit:

```
OrdersTable::exportUsing(CustomExporter::class);
```

These are quite basic at the moment and should just be a simple class with an `export` method that accepts an array of `$orderIds` but should return a streamable response:

```php
<?php

namespace GetCandy\Hub\Exporters;

use GetCandy\Models\Order;
use Illuminate\Support\Facades\Storage;

class OrderExporter
{
    public function export($orderIds)
    {
        // ...
        return Storage::download('order_export.csv');
    }
}
```

---

## Saved searches

You can apply filters or a search term and then save those results as a `SavedSearch`. This is a new table in GetCandy and is specific to the component it was saved on. This means you can potentially add saved searches to any component within the hub, allowing for a level of flexibility.

There is a new trait `GetCandy\Hub\Http\Livewire\Traits\WithSavedSearches` which you can add to a component to start allowing this functionality.

---

## Storefront Changes

There are some small tweaks needed to the configuration for installs that already have things set up.

`config/getcandy/orders.php`

### Old

```php
<?php

return [
    // ...
    'statuses' => [
        'awaiting-payment' => 'Awaiting Payment',
        'dispatched'       => 'Dispatched',
    ]
]
?>
```

### New

```php
<?php

return [
    // ...
    'statuses'     => [
        'awaiting-payment' => [
            'label' => 'Awaiting Payment',
        ],
        'paid' => [
            'label' => 'Paid',
            'color' => '#179669',
        ],
        'dispatched' => [
            'label' => 'Dispatched',
        ],
    ],
]
```
---

I think that's it for the moment, this is open for discussion, improvement. It would be good to get this nailed down so we can then start adding it to other areas of the hub, like products.

Once the approach is signed off, I'll add the doc updates / changelog and we can get this in. Then follow up with the order view PR.


glennjacobs:
- When saving a search, can the input field have focus set?
- Bit too easy to accidentally deleted a saved search, can we add an "Are you sure?"?
- Can the Filter button change state somehow to suggest when it's open/closed?
- If an Order Reference has a hyphen "-" in it, the reference is treated as separate words, meaning you can't find the order you're looking for.
- Saved searches don't display in mobile view?
- How would someone go about finding an order for someone that is a guest and needs to be identified by something else, e.g. email, postcode, etc.? I feel like we need to be able to search on a lot more of the order details.
- Dispatched status colour doesn't show well

image

- Trying to set an order status to "awaiting payment" gives this error

image

alecritson:
- When saving a search, can the input field have focus set? ✅ 
- Bit too easy to accidentally deleted a saved search, can we add an "Are you sure?"? ✅ 
- Can the Filter button change state somehow to suggest when it's open/closed? ✅ 
- If an Order Reference has a hyphen "-" in it, the reference is treated as separate words, meaning you can't find the order you're looking for.

Still looking at the tokenisation on this one.

- Saved searches don't display in mobile view? ✅ 
- 
- How would someone go about finding an order for someone that is a guest and needs to be identified by something else, e.g. email, postcode, etc.? I feel like we need to be able to search on a lot more of the order details.

Make some changes to the order document on the index, we now add a lot more info. Just need to reimport.

- Dispatched status colour doesn't show well ✅ 

- Trying to set an order status to "awaiting payment" gives this error ✅

glennjacobs:
Filter button not aligned correctly

image

glennjacobs:
The table isn't responsive on mobile. Needs a wrapping scrollable div?

glennjacobs:
Dates now look sweet 👍

glennjacobs (approved):

PR #213, opened 2022-03-09
Feature - Improved order view screen

alecritson:
This PR looks to overhaul the current order view screen. In its current state it's quite basic and we always knew it would need improvement, hopefully, this is it!

---

With this update we expand on the information that can be seen on an order, below is a screenshot of what's been added.

image

## Added Functionality

- Ability to update an order's status
- Initial groundwork for refunding an order (will be coming in a separate PR with the payments manager)
- Ability to edit the shipping address and if the postcode is different to billing, will display both.
- Additional information in the meta field on an order will show.
- Order lines can be expanded to show the totals (see below)

image

- A timeline has been added so you can see the journey the order goes through. This will be expanded on over time.
- You can add comments to an Order and it will show up in the timeline

image

- Transactions are listed out with colour coding and icons to show the type of card used and whether it was a refund or failed etc.
- You can select multiple lines, this will calculate a refund amount for the selected lines if you then choose to refund them. This will also be useful for add ons that extend functionality later down the line.

## Adding custom order actions

You can add items to the "More Actions" dropdown by targeting a new menu slot:

```php
use GetCandy\Hub\Facades\Menu;

$slot = Menu::slot('order_actions');

$slot->addItem(function ($item) {
    $item->name('Custom Action')
        ->handle('custom.action')
        ->component('livewire.component');
});
```

The reference in the `component` method should be a reference to your own custom Livewire component. This will receive an `$order` property which you can then do whatever you need to.


glennjacobs:
- When editing an address, the "State / Province" field should be a dropdown if the country has states in the database.
- Should be more mandatory fields than just postcode and country. In fact, postcode shouldn't always be mandatory as some countries do not have postcodes.
- I can't edit the shipping and billing addresses separately. Even if they're set to be the same in the checkout, the order shouldn't remember that. They should be treated as two separate addresses.
- Billing address details are showing shipping details.
- Dispatched status is unreadable
image
- "More Actions" doesn't do anything
- Timeline doesn't quite work 100%

Changing an order's status puts two entries into the timeline, one blank. The statuses also are the handles, not the proper names and sometimes the separator is missing. See below.

image

- Should we be displaying a user's email address? Is that not a privacy issue?
- Do we want to allow commenters a few minutes to edit their comment before it's committed? Just thinking customer service people making mistakes and causing a bit of a mess in orders.
- Refund button shows even when there's no payments
- Can't add new lines on comments, which makes them a little restrictive.
- Timestamps on comments are wrong, I added a comment at 10:13 and it shows as 10:03... in fact, all comments have the same time.

glennjacobs:
We have a client requirement to show stock levels on the order lines, do these screens give us the facility to add custom information in?

glennjacobs:
Got a double state field for some reason.

image

glennjacobs:
Billing address still shows as "Same as shipping address", even after editing. And the success message shows

image

glennjacobs:
Is the Gravatar image working?

glennjacobs:
Capture payment button shows, but there's nothing to capture.

glennjacobs:
image

I feel like the additional order lines can be easily missed. Maybe it needs something in line with the order lines at the bottom to say "And 12 more lines" or something? I guess I'm concerned someone will ship an order not realising there were more lines than shown...

glennjacobs:
image

Should it be "Captured" rather than "Capture"?

glennjacobs:
image

If the system created an activity log entry, do we need to show the `(Guest)` bit?

glennjacobs:
I imagine this should say "View Customer" as it takes you to the customer page.

image

alecritson:
@glennjacobs Think I've hit all the points above 🤔

glennjacobs:
All looks good, apart from this...

image

Seems to be missing gravatar images for the middle updates

glennjacobs:
Can we tidy up the fonts here a little?

image

glennjacobs (approved):

Issue #215, opened 2022-03-10
Thousand Point


loranger:
I'd like to add the euro currency which uses tiny space as a thousand separator, but admin does not allow me to do so:

- GetCandy version: 2.0-beta10.1
- Laravel Version: 9.4.1
- PHP Version: 8.1.3
- Database Driver & Version: Mariadb 10.7.3

### Expected Behaviour:

Add the new currency

### Actual Behaviour:

Throws an error

<img width="835" alt="Capture d’écran 2022-03-10 à 14 48 13" src="https://user-images.githubusercontent.com/6014/157674974-e3cc12fe-419d-4518-a75a-ceb81409cd9d.png">

### Steps To Reproduce:

Go to hub, and add a currency using a space as a thousand separator

glennjacobs:
I guess because Laravel trims the inputs by default, it ends up being a null value and the validation triggers.

I can see Euros need this though, so we'll need to somehow disable this behaviour on this field.

The middleware that does it looks like this

```php
class TrimStrings extends Middleware
{
    /**
     * The names of the attributes that should not be trimmed.
     *
     * @var array<int, string>
     */
    protected $except = [
        'current_password',
        'password',
        'password_confirmation',
    ];
}
```

Maybe we can dynamically add to it somehow??

lucasvmds:
I think it's better to add a "+" character in the field and replace it with a space in the `formatted()` method.

glennjacobs:
We'd have to explain that in the UI somehow if we did.

lucasvmds:
I'll take a look.

loranger:
Thank you guys for your support.

I was thinking: Candy import and uses country list, maybe it could also populate a complete (or partial) list of currencies from a package such as [money](https://github.com/moneyphp/money)…
Note that's only an idea, I don't know what I'm talking about 😁

glennjacobs:
> Thank you guys for your support.
> 
> I was thinking: Candy import and uses country list, maybe it could also populate a complete (or partial) list of currencies from a package such as [money](https://github.com/moneyphp/money)… Note that's only an idea, I don't know what I'm talking about 😁

I am considering it. We originally saw `cknow/laravel-money` which wouldn't have worked for us. But the underlying `moneyphp/money` could work.

Something I'd been mulling over for a while. But if we're going to do it, now is the time whilst we're still in beta.

glennjacobs:
Working on an idea to change how price formatting is done. Please see #222

[...]

PR #220, opened 2022-03-10
Hotfix/use space in thousand point for currency

lucasvmds:
#215 

This PR adjusts the thousand-point field of the currency form to recognize plus signs as spaces.
When displaying the formatted price, the plus sign will be changed to a space.


glennjacobs:
Thanks for the contribution, but we're going to change to using the PHP NumberFormatter class instead, therefore making this PR redundant.

PR #222, opened 2022-03-11
Price formatting

glennjacobs:
Allows prices to be formatted in locale-specific ways, using different number formatters.

Still todo...

- [x] Remove formatting information from Currency model
- [x] Documentation updates
- [x] Additional automated tests


alecritson:
Should we also consider whether to show the price with or without tax on this PR?

packages/core/tests/Unit/DataTypes/PriceTest.php
```diff
@@ -36,28 +36,43 @@ public function can_initiate_the_datatype()
...
-        $this->assertEquals('£0.155', $dataType->formatted);
+        $this->assertEquals('£0.16', $dataType->formatted);
+    }
+
+    /** @test */
+    public function can_format_numbers()
+    {
+        $currency = Currency::factory()->create([
+            'code'           => 'EUR',
+            'decimal_places' => 2,
```
alecritson, line 61: Is this now redundant?

glennjacobs: For what reason?

alecritson: In one of the tests it sets `decimal_places` to 3, but comes back with a number at 2 decimal places.

glennjacobs: I have no idea what you mean haha!

But no, this isn't redundant. It's proving the number formatted works with different locales and formatter styles.

alecritson: I think my confusion was we have something like this:

```php
$currencyUSD = Currency::factory()->create([
    'decimal_places' => 3,
    'code' => 'USD',
]);
```

Then this assertion

```php
$this->assertEquals('$13.00', $price->price->formatted('en-us'));
```

Which is 2 decimal places, so is it just used for validation and returning the decimal?

glennjacobs: That is the expected output, if you format a number using a standard NumberFormatter, then it will be 2 decimal places. But you've commented on a completely different file, hence my confusion.

alecritson: I understand that, my question was really what is `decimal_places` doing?

glennjacobs: Are you trying to say, should the decimal places change the formatter?

glennjacobs: Pushed an update to now use decimal places in the formatter.

alecritson (approved):

PR #224, opened 2022-03-11
Hotfix/fix missing urls on record creation

alecritson:
When we create a product or collection, we're initially missing a URL for the record. This can lead to breakages on storefronts that are expecting a URL to be present.

I've added some changes via an observer so if we create a collection/product and we don't have any URL we create one based on the translated name and using the default language.

I went for the observer route instead of something like validation in the hub as creation could potentially come from anywhere, such as an import routine, so this way it's unlikely to get missed.

### DIFF
--- a/packages/core/src/GetCandyServiceProvider.php
+++ b/packages/core/src/GetCandyServiceProvider.php
@@ -36,6 +36,7 @@ use GetCandy\Models\Collection;
 use GetCandy\Models\Currency;
 use GetCandy\Models\Language;
 use GetCandy\Models\OrderLine;
+use GetCandy\Models\Product;
 use GetCandy\Models\Url;
 use GetCandy\Observers\AddressObserver;
 use GetCandy\Observers\CartLineObserver;
@@ -44,6 +45,7 @@ use GetCandy\Observers\CollectionObserver;
 use GetCandy\Observers\CurrencyObserver;
 use GetCandy\Observers\LanguageObserver;
 use GetCandy\Observers\OrderLineObserver;
+use GetCandy\Observers\ProductObserver;
 use GetCandy\Observers\UrlObserver;
 use Illuminate\Auth\Events\Login;
 use Illuminate\Auth\Events\Logout;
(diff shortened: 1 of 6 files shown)


glennjacobs:
1. Is this intuitive in the UI? Would someone expect this to happen when creating a product in the hub?
For Collections, this isn't too much of a problem. For Products, they can see there are no URLs, but then one magically appears?

2. We place quite a lot of reliance on the "name" attribute in the system and there is nothing stopping a developer from removing it. I know we prevent it being changed in the hub by setting it as a "system" attribute. Are we worried about this?

3. Is it possible some might not want this behaviour? If so, should we consider a config setting for this?

alecritson:
> 1. Is this intuitive in the UI? Would someone expect this to happen when creating a product in the hub?
>    For Collections, this isn't too much of a problem. For Products, they can see there are no URLs, but then one magically appears?

I mean, I guess it could be confusing, personally if I saw a URL appear in the UI after creating a product I wouldn't really be too surprised.

> 2. We place quite a lot of reliance on the "name" attribute in the system and there is nothing stopping a developer from removing it. I know we prevent it being changed in the hub by setting it as a "system" attribute. Are we worried about this?

Tricky one, if we don't make at least some assumptions it would almost be like the Hub couldn't use attributes outside of saving/updating data 🤔 

> 3. Is it possible some might not want this behaviour? If so, should we consider a config setting for this?

Maybe we have a couple of fields:

```php

return [
    'auto_generate_urls' => true,
   'generate_urls_using' => 'some_other_attribute',
];

```

Could even have a class reference in the config perhaps,

```
return [
    'url_generator' => null, // Doesn't auto generate URLs
]
```

```php
return [
    'url_generator' => GetCandyUrlGenerator::class,
];
```

```php
public function handle(Model $model)
{
    $model->urls()->create([...]);
}
```

Or something, least people can do whatever they want then.

glennjacobs:
@alecritson I like the URL generator idea. Hopefully pretty simple.

PR #227, opened 2022-03-14
Feat - Add URL generation

alecritson:
This carries on from #224

The changes allow developers to add their own URL generator class to automatically generate URLs when creating records for models that use the `HasUrls` trait.

Currently, it just uses the `name` attribute but can be expanded if needed.

If we're happy with the approach I can update the docs/changelog and set to ready for review etc.


alecritson:
Some adjustments to this PR:

### Validation

You can set URLs to be `required` in the config. This will have an effect on validation during create/update depending on whether you have a URL generator.

### Collections

When creating a collection, if you have URLs set to be required and don't have a generator, you'll now see a `slug` field which is then used to create the default URL.

### Clean up

When you delete a model that uses the `HasUrls` trait, it will remove any existing URLs if it hasn't been soft deleted.

glennjacobs:
I feel most installs will want this turned on, so should we default to on?

Also, we'll want to update the demo store to use this once released.

glennjacobs:
Need to add it to in unpublished upgrade guide.

alecritson:
> I feel most installs will want this turned on, so should we default to on?
> 
> Also, we'll want to update the demo store to use this once released.

I did think about that. I guess we could turn it on by default, would as you say, just need to add anything to the upgrade guide that might cause an impact.

glennjacobs:
> > I feel most installs will want this turned on, so should we default to on?
> > Also, we'll want to update the demo store to use this once released.
> 
> I did think about that. I guess we could turn it on by default, would as you say, just need to add anything to the upgrade guide that might cause an impact.

I don't think anything happened on this?

I guess we could just turn it on for the demo store and be done with it.

PR #233, opened 2022-03-21
Adding a "sorting" column to Products to enable simpler sorting

glennjacobs:
This PR looks to add a JSON `sorting` column to products to allow them to be ordered easily and quickly. E.g. when displaying products in a collection.

For performance, a virtual index may want to be added, but this is something we'll have to leave up to the developer as the JSON fields are dynamic.

Example usage would be...

```php
    $collection = \GetCandy\Models\Collection::find(1);

    $products = $collection
        ->products()
        ->useSorting('price_default', 'desc')
        ->get();
```

## TODO

- [ ] Make `useSorting` compatible with PostgreSQL
- [ ] Documentation updates
- [ ] Automated tests


glennjacobs:
A consideration is this assumes an iso currency code. Currently the admin hub does not enforce this.

glennjacobs:
Another thought is should we give customer groups "handles"?

PR #236, opened 2022-03-23
Feature - Improved customer screens

alecritson:
Initial attempt at revamping the customer screens.

## Added Information

- Avg. Spend
- Spending total
- Number of orders
- Graph to show spending across the year compared with the previous year
- Table to show the customer's most popular products
- Listing of orders and which user made the order
- List of users associated with the customer
- List of addresses for the customer.

There are some bits, such as address editing which would need to wait for the order screens as there is address editing on there and so would be good to reuse those components.

Wasn't sure if we should be associating users in the hub?

Also, currently when an order is created the customer isn't associated with it, since a user can be associated with many customers, how do we want to handle this?


glennjacobs:
Is there a seeder or something I can use to test with? I don't have a suitable way to add a bunch of customers at the mo...

glennjacobs:
I feel we need to keep things consistent, and so I suggest we introduce a "view" link on customer orders, rather than clicking the ID which we've not done anywhere else.

image

Generally, though, I find the small "view" buttons fiddly. I'd like to review these at some point and make them buttons that can be extended with drop-down options.

glennjacobs:
Responsiveness could do with a tweak. The right side column should drop down before this I'd suggest.

image

glennjacobs:
Once dropped down, it should be full width really.

image

glennjacobs:
Reset password errors

image

glennjacobs:
Can we seed some addresses so I can test?

alecritson:
Points above should be actioned. For seeding I've added some to the branch on the demo store, seeder should be changed to handle existing data, might want to just double check.

glennjacobs (approved):

PR #240, opened 2022-03-25
Get pricing directly from models with the hasPrices trait

glennjacobs:
Adding a convenient method to the `hasPrices` trait, so developers can access the pricing manager by simply doing `->pricing()` on a model.

- [x] Documentation updates
- [x] Automated tests


packages/core/src/Base/Traits/HasPrices.php
```diff
@@ -26,4 +30,33 @@ public function basePrices()
...
+        $pricing = Pricing::qty($qty);
+
+        // Set currency, even if NULL
+        $pricing->currency($currency);
+
+        // Set user, even if NULL
+        $pricing->user($user);
+
+        // Set customerGroups, even if NULL
+        if ($customerGroups instanceof Collection) {
```
alecritson, line 54: Should we be worried if an array is passed?

glennjacobs: The PricingManager expects a Collection. So that would need updating also...

PR #243, opened 2022-03-29
Fix - Line item quantity type

markmead:
This PR adds a migration to change the `CartLines` table `quantity` column to use the type `unsignedInteger`.

Doing this prevents the following error when adding a LOT of items to the cart:

image 1


glennjacobs:
Do we think the qty field really needs up to `4,294,967,295`?

alecritson:
Maybe we go on up to a `MEDIUMINT` ?

glennjacobs:
It's still over 16million... seems a bit excessive. What do we think the maximum supported qty should be?

glennjacobs:
A SMALLINT would be `65535`, I guess you could argue that might not be enough?

glennjacobs:
> I checked with an existing database and there a number of orders which had a quantity higher than `65535` for an order line.

Well, I think an UNSIGNED MEDIUMINT will be more than enough, so let's do that.

markmead:
Sadly, `doctrine/dbal` doesn't support `unsignedMediumInteger`. Unsure what to do next but the two solutions I see are:

- Use `unsingedInteger`
- Modify existing migration to use `unsignedMediumInteger` and add information to the upgrade docs

glennjacobs:
You can't do...

```php
$table->mediumInteger('qty')->unsigned();
```

??

glennjacobs:
> @glennjacobs You can, but that would only work for new installs, which is why the upgrade docs were mentioned. So you'd add it to the base migration then would need something in the upgrade docs to tell people to manually change the type.

Bit manky hey, don't want that. I'm amazed this is still an issue in Laravel.

markmead:
Definitely not ideal. 

Should we leave this one then? It's quite project-specific, I can't think of many stores that need `65535`+ line item quantity 😅

glennjacobs:
> Definitely not ideal.
> 
> Should we leave this one then? It's quite project-specific, I can't think of many stores that need `65535`+ line item quantity 😅

Unfortunately, our current client project probably will need it! 

Let's stick with the unsigned integer.

We will also need to update the `order_lines` table so they match and add a test for a really big qty value on both `CartLine` and `OrderLine`.

markmead:
Updated the `OrderLines` table. 

As for testing, I'm not quite sure what's happening. I switched to `main` to check that a large quantity would fail the test but it wouldn't.

```php
public function can_make_a_cart_line()
{
  $cart = Cart::factory()->create([
    'user_id' => User::factory(),
  ]);

  $data = [
    'cart_id'          => $cart->id,
    'quantity'         => 99999999999999999999999,
    'purchasable_type' => ProductVariant::class,
    'purchasable_id'   => ProductVariant::factory()->create()->id,
  ];

  CartLine::create($data);

  $this->assertDatabaseHas((new CartLine())->getTable(), $data);
}
```

This would pass. Tried a `DB::table('getcandy_cart_lines')->insert([...])` as well and that would pass.

glennjacobs:
> Updated the `OrderLines` table.
> 
> As for testing, I'm not quite sure what's happening. I switched to `main` to check that a large quantity would fail the test but it wouldn't.
> 
> ```
> public function can_make_a_cart_line()
> {
>   $cart = Cart::factory()->create([
>     'user_id' => User::factory(),
>   ]);
> 
>   $data = [
>     'cart_id'          => $cart->id,
>     'quantity'         => 99999999999999999999999,
>     'purchasable_type' => ProductVariant::class,
>     'purchasable_id'   => ProductVariant::factory()->create()->id,
>   ];
> 
>   CartLine::create($data);
> 
>   $this->assertDatabaseHas((new CartLine())->getTable(), $data);
> }
> ```
> 
> This would pass. Tried a `DB::table('getcandy_cart_lines')->insert([...])` as well and that would pass.

Probably because tests use Sqlite?

glennjacobs:
Doesn't seem like we can add any helpful tests here, so suggest it's fine as-is.

packages/core/CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Changed
 
 - When generating media conversions, the original file format is now kept. By [@lucasvmds](https://github.com/lucasvmds)
+- Quantity column on `CartLines` table is now of type `unsignedMediumInteger`.
```
glennjacobs, line 27: This needs updating

glennjacobs: Wrong field type, wrong table name (`cart_lines`) and doesn't include `order_lines`

packages/core/database/migrations/2022_03_30_100000_update_quantity_on_order_lines_table.php
```diff
@@ -0,0 +1,22 @@
+<?php
```
glennjacobs, line 1: This file should be `2022_03_30_100001_...`

PR #245, opened 2022-03-29
Release/order screens

glennjacobs:
We have some commits directly on this branch, so created this draft PR as a reminder.


packages/core/database/migrations/2022_03_17_100000_add_fields_to_transactions_table.php
```diff
@@ -0,0 +1,36 @@
...
+
+    public function down()
+    {
+        Schema::table($this->prefix.'transactions', function ($table) {
+            $table->dropForeign(['parent_transaction_id']);
+            $table->dropColumn('parent_transaction_id');
+            $table->dropColumn('type');
+        });
+
+        Schema::table($this->prefix.'transactions', function ($table) {
```
glennjacobs, line 32: Why do we specify the same table twice?

alecritson:
The commits are cherry-picked from the various branches that make it up. I will double check when everything is good. The idea is going to be to just delete this branch when ready.

packages/core/database/migrations/2022_03_17_100000_add_fields_to_transactions_table.php
alecritson: Some engines don't allow you to have multiple changes in one transaction so they have to be separated out.

PR #246, opened 2022-03-29
Draft idea for indexing collections and prices

glennjacobs:
This is the initial start of adding indexing for Collections and Prices for Products. Needs some discussion.

- [ ] Documentation updates
- [ ] Automated tests

### DIFF
--- a/packages/core/src/Models/Product.php
+++ b/packages/core/src/Models/Product.php
@@ -43,6 +43,7 @@ class Product extends BaseModel implements SpatieHasMedia
     protected $filterable = [
         '__soft_deleted',
         'skus',
+        'collection_ids',
         'brand',
         'status',
     ];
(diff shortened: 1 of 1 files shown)


alecritson:
Should we index the channel availability? Just thinking if you filtered them out after retrieving them from the search index, the pagination is potentially going to be all skewed.

We should probably change instances of `variants()` or anything we get a relationship from to the property reference, otherwise, it's not going to be compatible with eager loading 🤔 

We should also sort out eager loading on index, something like:

```php
   /**
     * Modify the query used to retrieve models when making all of the models searchable.
     *
     * @param  \Illuminate\Database\Eloquent\Builder  $query
     * @return \Illuminate\Database\Eloquent\Builder
     */
    protected function makeAllSearchableUsing($query)
    {
        return $query->with([
            'thumbnail',
            'variants',
            'variants.values',
            'variants.values.option',
            'variants.basePrices.currency',
            'collections',
        ]);
    }
```

Should this be configurable or extendable aswell?

glennjacobs:
Closing because it is getting too opinionated. Devs can add their own indexing logic already, so this is better provided via tutorials.

Issue #251, opened 2022-04-01
HasTranslations::translate() method ignores $locale parameter 


armezit:
- GetCandy version: 2.0-beta10
- Laravel Version: 9.3.1
- PHP Version: 8.1.2
- Database Driver & Version:

### Expected Behaviour:

Attribute should be translated to the given locale or fallback to Laravel default locale.

### Actual Behaviour:

`translate()` method ignores locale, and returns the first option value saved in database.

### Steps To Reproduce:

In a multi-language store, define a product variant option.
The `name` attribute stores in the `product_option_values` table like this: `{"en":"10 USD","fa":"10 دلاری"}`
Now, if you call the `ProductVariant::getOption()` method, it returns the **first** value (even if the Laravel default locale is `fa`), which is `10 USD` in the above example.

The problem is here: https://github.com/getcandy/getcandy/blob/5020fdb7e296cae954cff4b31eda23b33fbd41cd/packages/core/src/Base/Traits/HasTranslations.php#L29 
`$values` type is `stdClass`, but used as array.

alecritson:
This should be resolved in v2.0-beta11

PR #253, opened 2022-04-01
Fixes issue #251: HasTranslations::translate() method ignores $locale parameter

armezit:
Fixes issue #251: HasTranslations::translate() method ignores $locale parameter


alecritson:
Hi @armezit Thanks for this, we'll take a look. It seems this trait has some tests missing so we can either add them to the main branch and you rebase and update your PR or feel free to add them to your PR. Just let me know what you want to do.

armezit:
Thanks @alecritson I think it's better that you add tests and I rebase main branch. Because I'm not much familiar with test cases of the project yet.

alecritson:
Hey, @armezit I've added some tests to this PR to keep things in one place, turns out there wasn't much needed. I've also added the changelog entry but in future can you make sure this is done with the PR :)

alecritson (approved):

PR #255, opened 2022-04-02
Feat - Order position of sidebar menu links

markmead:
Give developers the option to re-order the sidebar menu items.

Developers can clone down the configuration files from GetCandy and update the `sidebar.php` to include the new items.

```php
// config/getcandy-hub/sidebar.php

<?php

return [
    'order' => [
        'hub.index',
        'hub.products',
        'hub.product-type',
        'hub.collection',
        'hub.orders',
        'hub.customers',
        'hub.tickets', // New item
    ]
];
```

GetCandy Hub

Be great to get your thoughts on this @AaronNeonDigital as it was your idea.


glennjacobs:
I'm not keen on doing this via config files. We should allow addons to position their links using Collection methods.

markmead:
What are you thinking? Something like:

```php
$slot->addItem(function ($item) {
    $item->name(
        __('menu.sidebar.tickets')
    )->handle('hub.tickets')
    ->route('hub.tickets.index')
    ->icon('ticket')
    ->position(4);
});
```

This is the approach I had but it didn't play nice with the default GetCandy menu, as you could end up with multiple items with `position(4)` and then the UI doesn't reflect what you'd expect.

alecritson:
> What are you thinking? Something like:
> 
> ```
> $slot->addItem(function ($item) {
>     $item->name(
>         __('menu.sidebar.tickets')
>     )->handle('hub.tickets')
>     ->route('hub.tickets.index')
>     ->icon('ticket')
>     ->position(4);
> });
> ```
> 
> This is the approach I had but it didn't play nice with the default GetCandy menu, as you could end up with multiple items with `position(4)` and then the UI doesn't reflect what you'd expect.

It is a tricky once since add on devs would potentially all just set their position to `1` 🤔

AaronNeonDigital:
`->position($group, $order)`
Hub will have group 1, this will be order will be order within group.
group 1 then can only be used for admin hub, and can be prioritised to the top of the getcandy menu.

markmead:
Pushed an update that allows addon creators to include `->order()` without it affecting the GetCandy menu.

glennjacobs:
> Pushed an update that allows addon creators to include `->order()` without it affecting the GetCandy menu.

Reasonable idea. Although say you were making a Product related addon, it might make sense for it to be under the Products menu link.

We kinda want to allow an addon to add itself to the menu without any work. But we also want to allow a developer to arrange the menu if they're not happy with the order.

markmead:
It's for that reason I used `config/sidebar` to start with. It seems this is caught between a rock and a hard place.

Using something like `order` will cause conflicts with the GetCandy menu and the `config/sidebar` approach isn't ideal.

Potentially two solutions:
1. Keep using `order` but have something that runs when a menu item changes to keep the values unique.

Not sure how we'd approach that.

2. Add some UI in GetCandy to manage the order, this could be an addon to keep it out of core and keep it optional.

PR #260, opened 2022-04-07
Feat - Add multiple lines to the cart

markmead:
This PR adds a function called `addLines` that allows for bulk adding to the cart.

It doesn't add anything too exciting but it's functionality I had to create that I thought could be good in the core.

### DIFF
--- a/docs/src/getcandy/carts.md
+++ b/docs/src/getcandy/carts.md
@@ -221,6 +221,21 @@ CartSession::manager();
 CartSession::add($purchasable, $quantity);
 ```
 
+### Add multiple lines
+
+```php
+CartSession::addLines([
+    [
+        'id' => 1,
+        'quantity' => 25,
+        'meta' => ['foo' => 'bar'],
+    ],
+    // ...
+]);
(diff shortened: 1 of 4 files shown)


markmead:
Sure! Here's what I was doing:

```php
foreach ($this->purchasableCart->lines as $line) {
    CartSession::add(
        $line->purchasable,
        $line->quantity
    );
}
```

But with the update I can do:

```php
CartSession::addLines($this->purchasableCart->lines);
```

It's not much so I understand if it's a no to moving this into the core.

alecritson:
Yeah, I can see a use for it, like if you had a page where you could add multiple quantities of each variant in one go you don't have to loop through doing multiple queries.

glennjacobs:
Things that concern me are...

1) Passing a Collection through gives no structure to what is expected.
2) Only allows Collections, what if we want to pass something else that can be iterated, such as an array?

glennjacobs:
@markmead can you add docs for this?

glennjacobs (approved):

PR #262, opened 2022-04-08
Hotfix/only use getcandy users

glennjacobs:
Prevents edge-case scenarios whereby a non-getcandy user can be presented to the Cart or PricingManager logic.

### DIFF
--- a/packages/core/src/helpers.php
+++ b/packages/core/src/helpers.php
@@ -6,6 +6,7 @@ if (! function_exists('is_getcandy_user')) {
     function is_getcandy_user($user)
     {
         $traits = $traits = class_uses_recursive($user);
+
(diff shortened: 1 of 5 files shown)


packages/core/src/helpers.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+use GetCandy\Base\Traits\GetCandyUser;
+
+if (! function_exists('is_getcandy_user')) {
+    function is_getcandy_user($user)
+    {
+        $traits = $traits = class_uses_recursive($user);
```
alecritson, line 8: Double `$traits` and is setting the variable required?

alecritson (approved):

Issue #266, opened 2022-04-17
Autogenerate urls not set to null by default


AaronNeonDigital:
- GetCandy version: Beta12
- Laravel Version: 9
- PHP Version: 8.0

Generator variable in urls is meant to have a default of null, as per to the grtCandy2 docs. In this case it is not set to null by default.

### Expected Behaviour:
config => getcandy => urls.php
`'generator' => null`
### Actual Behaviour:
`'generator' => GetCandy\Generators\UrlGenerator::class,`

### Steps To Reproduce:
N/A

PR #267, opened 2022-04-17
Hotfix/fixed null on config urls

AaronNeonDigital:
-fixed default url generation set to null as docs states

#266


docs/src/getcandy/products.md
```diff
@@ -85,7 +85,7 @@ If you decide to delete an attribute, this will cause the association to be drop
 :::
 
 ### Retrieving the product type relationship
-If you have a product, you can fetch it's product like like so:
+If you have a product, you can fetch its product type like like so:
```
markmead, line 88: Says `like like so`

glennjacobs: @AaronNeonDigital can you fix that up?

glennjacobs:
@alecritson do we want it set as NULL by default like the docs say? Or should we change the docs?

https://docs.getcandy.io/getcandy/urls.html#automatically-generating-urls

alecritson:
@glennjacobs I'm more on the side of changing the docs, since changing code will change how it currently works so would need something in the upgrade guide. Unless we want the functionality to change.

glennjacobs:
> 

@alecritson It's not code as such, just the default config value.

glennjacobs:
@alecritson gotcha. Let's update the docs then. @AaronNeonDigital did you want to attempt this, or shall we?

glennjacobs:
@alecritson you happy with this now?

alecritson:
@glennjacobs Yeah I think it's fine!

glennjacobs (approved):

PR #269, opened 2022-04-19
Feature: Added Slots to Order Show Page in Hub

przper:
Added custom Slots in Admin Hub to OrderShow Page.

### DIFF
--- a/packages/admin/resources/views/livewire/components/orders/show.blade.php
+++ b/packages/admin/resources/views/livewire/components/orders/show.blade.php
@@ -113,6 +113,12 @@
         </header>
       @endif
 
+      @foreach($this->getSlotsByPosition('top') as $slot)
+       <div id="{{ $slot->handle }}">
+        <div>@livewire($slot->component, ['slotModel' => $order], key("top-slot-{{ $slot->handle }}"))</div>
+       </div>
+      @endforeach
+
       <section class="bg-white rounded-lg shadow">
         @include('adminhub::partials.orders.details')
       </section>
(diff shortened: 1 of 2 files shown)


packages/admin/resources/views/livewire/components/orders/show.blade.php
```diff
@@ -161,6 +167,12 @@ class="flex-shrink-0 px-4 py-2 ml-4 text-xs font-bold text-gray-700 border round
           @endforeach
         </dl>
       </section>
+
+      @foreach($this->getSlotsByPosition('bottom') as $slot)
+       <div id="{{ $slot->handle }}">
+        <div>@livewire($slot->component, ['slotModel' => $order], key("top-slot-{{ $slot->handle }}"))</div>
```
ryanmitchell, line 173: Key should be bottom-slot-xxx not top-slot-xxx

przper: Fixed in https://github.com/getcandy/getcandy/pull/269/commits/d2cee7b80a9f7e52b0e350018f7e6a53209a1cba

alecritson:
Sorry for the delay on this one @przper I think it would also be handy to have a slot for each order line and also one on the side, under the addresses and meta. What do you think?

przper:
> Sorry for the delay on this one @przper I think it would also be handy to have a slot for each order line and also one on the side, under the addresses and meta. What do you think?

- In my use case I don't need Slots for Lines. I use slot to re-dispatch Laravel Job which tries to send XML file and sends email to customer upon success, then updates Order Status. I do not modify Order data except status.
I think adding Slots to OrderLines would be an overkill. Modifying an OrderLine would likely lead to change in Order total amount or other important param. This would require Order reevaluation. Maybe in future someone will have a good reason for OrderLine slots.
- I have added slots on top of details section and below meta section. I think you've missed that :)

alecritson:
@glennjacobs This one is a bit overdue, you think we can get this one in?

glennjacobs:
> @glennjacobs This one is a bit overdue, you think we can get this one in?

Yep, don't see why not.

alecritson (approved):

PR #274, opened 2022-04-22
Hotfix - Fix collection loading bottlenecks

alecritson:
Currently, if you have a lot of collections then the collection tree has a massive hit on browser performance due to the amount of Javascript being loaded up for each node tree. 

This PR rewrites most of the logic so we only load up one tree at a time and switches to using a Livewire component to fetch the children. 

The logic for doing the updates has mostly stayed on the CollectionGroup component since we want to trigger events to update the tree which we do things like reordering.

This also includes an eager loading fix to the collection view which was slowing it down.


packages/admin/resources/views/livewire/components/collections/collection-groups/show.blade.php
```diff
@@ -146,7 +146,12 @@ class="block w-full px-4 py-2 text-sm text-left hover:bg-gray-100"
         </x-hub::modal.dialog>
       @endif
       <div class="mt-4 space-y-2">
-        <x-hub::nested-tree :tree="$this->getCollectionTree()" :owner="$group" sort-group="root" />
+        @livewire('hub.components.collections.collection-tree', [
+          'nodes' => $tree,
+          'sortGroup' => 'root',
+          'owner' => $group,
+        ], key('tree-root'))
+        {{-- <x-hub::collection-tree :tree="$tree" :owner="$group" sort-group="root" /> --}}
```
glennjacobs, line 154: Should we keep commented code?

glennjacobs (approved):

PR #275, opened 2022-04-24
sidebar urls with permission

nurmuhammet-ali:
Right now, even if the staff does not have permission, can view links in the sidebar. My solution is not the best. Open to any ideas. But please let's get it done.


alecritson:
We currently have some code in the [MenuSlot](https://github.com/getcandy/getcandy/blob/main/packages/admin/src/Menu/MenuSlot.php#L94) class, would this mean this isn't functioning properly?

nurmuhammet-ali:
Hey, what's up? I'm not saying there is a bug or error, I'm saying Users of the hub who don't have permission to view the resource (example: product types, ...), can see the link in the sidebar. Would this mean this isn't functioning properly? Well, I think yes. They don't have permission to view it so links in the sidebar should be hidden. I'm sure my solution is not the best one, so if you have a better solution, let's pull that in.

PR #276, opened 2022-04-29
Allow number fields to be nullable

ryanmitchell:
When a number field was not required it was throwing an error when blank due to the strict blank check (the value can be null or blank).


alecritson:
Thanks @ryanmitchell !

Issue #286, opened 2022-05-13
Adding attributes after products exist causes errors


ryanmitchell:
GetCandy version: 2.0-beta12
Laravel Version: 8.83.6
PHP Version: 8.0
Database Driver & Version: Mysql 8

### Expected Behaviour:
Adding attributes once products/variants should be possible without errors.

### Actual Behaviour:
Adding attributes causes errors in the hub on existing product editing (e.g. GetCandy\FieldTypes\Text::__toString(): Return value must be of type string, null returned).

### Steps To Reproduce:
Create a product
Add a new attribute on variations that is a text or toggle field type
Edit the product
See error.

It can be fixed by updating getValue() in the fieldtypes to check for null but I didn't want to PR that in case you had reasons for not doing it.

alecritson:
Thanks @ryanmitchell , I just tried this and things seemed to work as expected for me. Here's what I did.

1) Added a new text attribute on a product variant
2) Assign it via the product type editing
3) Click on product to edit
4) Click on a variant
5) See the new field

<img width="787" alt="image" src="https://user-images.githubusercontent.com/1488016/168535945-f248ea00-5827-4174-b6dd-4d7beb53211e.png">

ryanmitchell:
When you try to save the product (through code not the UI) do you get the errors? if I PR the changes to the fields required (minimal null checks) are you open to merging anyway?

alecritson:
I must admin I only tried the UI, if there's an error though, happy to see a PR :)

PR #287, opened 2022-05-16
Handle null values in fieldtypes

ryanmitchell:
Fixes: #286 

When attributes are added after a product / variant exists, they will be null and throw an error if being saved through code. So we need to handle this in the getValue() check.


glennjacobs:
@ryanmitchell any chance you could fix up the tests on this one, so we can include in the next release?

ryanmitchell:
Probably not today. I need to think about how to get around the test failures (or if indeed I should). I’ll move to draft for now.

glennjacobs:
@ryanmitchell will you be able to finish this up? Be good to get this one out the way.

ryanmitchell:
I need to get some time to do some digging to see if it’s still an issue and the best approach to fix it. 

On holidays at the moment so won’t be able to look at it for a couple of weeks.

ryanmitchell:
Ive updated it to only care about the toString() return value, which handles the livewire issues I was seeing. I may need another PR to address the original issue, but for now this is enough to get me moving again.

alecritson (approved):

PR #296, opened 2022-05-19
Feat - Allow product restoring

### LINKED ISSUE #285: Product delete cascading
GetCandy version: 2.0-beta12
Laravel Version: 8.83.6
PHP Version: 8.0
Database Driver & Version: Mysql 8

### Expected Behaviour:
Deleting a product should delete its variants, prices and urls.

### Actual Behaviour:
The product is deleted (soft-deleted) but this does not cascade down to variants, their prices or urls. This means if you try to add the same SKU or URL again you get an error, even though the variant is 'orphaned'.

### Steps To Reproduce:
Add a product, and give it 2-3 variants.
Delete the product
Add a new product with the same name and use the same skus you used previously
You'll see errors for unique skus and probably URLs to

I appreciate this isnt going to happen a whole lot in the real world, but I'm writing import scripts so coming up against it and though it worth reporting.

alecritson:
- Adds initial filtering based on Product status and deleted status
- Adds ability to restore deleted products
Fixes #285

### DIFF
--- a/packages/admin/resources/views/livewire/components/products/index.blade.php
+++ b/packages/admin/resources/views/livewire/components/products/index.blade.php
@@ -8,43 +8,6 @@
     </div>
   </div>
 
-  <div>
-    {{-- <div class="sm:hidden">
-      <label for="tabs" class="sr-only">Select a tab</label>
-      <select id="tabs" name="tabs" class="block w-full py-2 pl-3 pr-10 text-base border-gray-300 rounded-md focus:outline-none focus:ring-indigo-500 focus:border-indigo-500 sm:text-sm">
-        <option>My Account</option>
-
-        <option>Company</option>
-
-        <option selected>Team Members</option>
-
-        <option>Billing</option>
-      </select>
-    </div> --}}
-    {{-- <div class="hidden sm:block">
-      <div class="border-b border-gray-200">
-        <nav class="flex -mb-px space-x-8" aria-label="Tabs">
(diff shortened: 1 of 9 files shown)


packages/admin/resources/views/livewire/components/products/index.blade.php
```diff
@@ -183,9 +191,20 @@
...
 
         <x-hub::table.cell>
+          @if($product->deleted_at)
+            <x-hub::button
+              size="xs"
+              theme="gray"
+              type="button"
+              wire:click="restoreProduct({{ $product->id }})"
+            >
+              Restore
```
glennjacobs, line 201: Language file

packages/admin/resources/views/livewire/components/products/index.blade.php
```diff
@@ -61,41 +61,24 @@
...
-              <x-hub::input.text placeholder="To" />
-            </div>
-          </x-hub::input.group>
-
-          <x-hub::input.group label="Price Range" for="brand">
-            <div class="grid grid-cols-2 gap-2">
-              <x-hub::input.text placeholder="From" />
-              <x-hub::input.text placeholder="To" />
-            </div>
+          <x-hub::input.group label="Show Deleted" for="brand">
```
glennjacobs, line 80: Does the label need to be in a language file?

glennjacobs (approved):

PR #310, opened 2022-06-06
Hotfix - Allow pipeline modifiers to be chained

### LINKED ISSUE #304: CartLineModifiers and CartModifiers only work on the first pipeline registered
### Expected Behaviour:
That you can register multiple cart line and cart modifiers and they would all process.

### Actual Behaviour:
Only the first one you register processes.

### Steps To Reproduce:
Make 2 simple cart line modifiers:

```
class ModifierOne extends CartLineModifier
{
    public function calculating(CartLine $cartLine)
    {
        echo 1;
    }
}
```

```
class ModifierTwo extends CartLineModifier
{
    public function calculating(CartLine $cartLine)
    {
        echo 2;
    }
}
```

and register them:

        $cartLineModifiers->add(
           ModifierOne::class,
        );
                
        $cartLineModifiers->add(
           ModifierTwo::class
        );

Add something to your cart and you'll see only modifier one echoes a response.

Looking at [this article](https://jeffochoa.me/understanding-laravel-pipelines), it seems its due to you not calling the $next closure in the modifiers.

alecritson:
Fixes #304 

---

The way modifiers work makes it difficult for them to be chained. This is due to needing to call `$next` in the modifier so it's sent through the pipeline.

The issue with that is it will rely on developers who use modifiers to implement this themselves, it will likely cause issues i.e. if it's forgotten or if developers end up passing the wrong object/value back through the pipeline 

A proposed solution is to implement a new method on each abstract modifier class called via the pipeline instead, meaning GetCandy can retain control and means any current modifiers can stay the same.

For example, on the CartLineModifier

```php
    /**
     * Handle the process calculating pipeline
     *
     * @param CartLine $cartLine
     * @param \Closure $next
     *
     * @return \Closure
     */
    public function processCalculating(CartLine $cartLine, $next)
    {
        $this->calculating($cartLine);

        return $next($cartLine);
    }
```

Which is then called in the CartLineManager like so:

```php
$pipeline->send($this->cartLine)->via('processCalculating')->thenReturn();
```

### DIFF
--- a/packages/core/src/Base/CartLineModifier.php
+++ b/packages/core/src/Base/CartLineModifier.php
@@ -8,6 +8,36 @@ use GetCandy\Models\CartLine;
 abstract class CartLineModifier
 {
     /**
+     * Handle the process calculating pipeline
+     *
+     * @param CartLine $cartLine
+     * @param \Closure $next
+     *
+     * @return \Closure
+     */
+    public function processCalculating(CartLine $cartLine, $next)
+    {
+        $this->calculating($cartLine);
+
+        return $next($cartLine);
+    }
+
+    /**
+     * Handle the process calculating pipeline
+     *
+     * @param CartLine $cartLine
+     * @param \Closure $next
+     *
+     * @return \Closure
+     */
+    public function processCalculated(CartLine $cartLine, $next)
+    {
+        $this->calculated($cartLine);
+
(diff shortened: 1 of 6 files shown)


alecritson:
After chatting this over with @glennjacobs we're going to go with a different approach...

PR #311, opened 2022-06-06
Hotfix - Adjust modifiers going through pipelines

### LINKED ISSUE #304: CartLineModifiers and CartModifiers only work on the first pipeline registered

alecritson:
Fixes #304 

---

This PR looks to change the way Modifiers are defined and used throughout GetCandy. Currently, we don't enforce developers to call `$next(...)` on the modifier as they do with other pipelines such as Middleware. This change will bring this into GetCandy and should make things feel more "Laravel".

This is a breaking change so people upgrading should consult the upgrade guide to see what changes they should make to each of their custom modifiers.

### DIFF
--- a/docs/src/extending/order-modifiers.md
+++ b/docs/src/extending/order-modifiers.md
@@ -32,7 +32,6 @@ class CustomOrderModifier extends OrderModifier
     public function creating(Cart $cart, Closure $next): Cart
     {
         //...
-
         return $next($cart);
     }
 
--- a/docs/src/extending/order-modifiers.md
+++ b/docs/src/extending/order-modifiers.md
@@ -22,20 +22,24 @@ public function boot(OrderModifiers $orderModifiers)
 ```php
 namespace App\Modifiers;
 
+use Closure;
 use GetCandy\Base\OrderModifier;
 use GetCandy\Models\Cart;
 use GetCandy\Models\Order;
 
 class CustomOrderModifier extends OrderModifier
 {
-    public function creating(Cart $cart)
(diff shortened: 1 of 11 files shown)


docs/src/extending/order-modifiers.md
```diff
@@ -22,20 +22,24 @@ public function boot(OrderModifiers $orderModifiers)
...
 use GetCandy\Models\Order;
 
 class CustomOrderModifier extends OrderModifier
 {
-    public function creating(Cart $cart)
+    public function creating(Cart $cart, Closure $next): Cart
     {
-        //
+        //...
+
```
glennjacobs, line 35: This example has an extra break line for some reason?

glennjacobs (approved):

PR #315, opened 2022-06-09
Feat - Support scout database

### LINKED ISSUE #302: Hub orders index throws ErrorException Attempt to read property "name" on null
- GetCandy version: dev-main
- Laravel Version: 8.75
- PHP Version: 8.1
- Database Driver & Version: MySQL 8.0.28

### Expected Behaviour:
It should open orders list.

### Actual Behaviour:
ErrorException: Attempt to read property "name" on null.

### Steps To Reproduce:
By opening admin hub's orders list (hub/orders).

image

alecritson:
Fixes #302 

---

This PR will add support for the `database` driver in Scout. The only thing I've come across is if we use the same method as the `mysql` driver, search doesn't seem to do anything. It's almost as if we need something specific for database drivers such as:

```php
public function toDatabaseSearchableArray() { ... }
```

Do we want to do this though?


alecritson:
> > The only thing I've come across is if we use the same method as the `mysql` driver, search doesn't seem to do anything. It's almost as if we need something specific for database drivers such as:
> > ```
> > public function toDatabaseSearchableArray() { ... }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Do we want to do this though?
> 
> I could probably do with understanding this a bit more.

It was just basically when I tried out the scout driver and typed in a search term, the results didn't change 🤔

glennjacobs:
This will never work, so closing. We have a new approach with a custom Scout driver for GetCandy.

PR #320, opened 2022-06-17
Add variant slot on product view

### LINKED ISSUE #318: Showing variant slots when there is only 1 variant
This may end up being more a question rather than an issue, but interested to hear your thoughts.

I have a slot for variants (downloadable files), which works fine when there are multiple variants, but when a product only has 1 it doesnt get shown as I'm registering it against the variant show slot.

Do we need to update the product views to also show variant slots in this case? Should we give the dev the option of always linking through to the variant show view? Or whats your opinion on what should happen? I could manually add another slot that links through to the variant view as a workaround, but it feels like something that should be considered a bit.

ryanmitchell:
Closes #318 

This adds a product variant slot when there is only 1 variant on a product.

### DIFF
--- a/packages/admin/resources/views/partials/products/editing/sections.blade.php
+++ b/packages/admin/resources/views/partials/products/editing/sections.blade.php
@@ -43,7 +43,7 @@
        --}}
         @foreach($this->getSlotsByPosition('top') as $slot)
        <div id="{{ $slot->handle }}">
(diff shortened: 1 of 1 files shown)


alecritson:
Thanks @ryanmitchell This looks like it should do the trick.

alecritson (approved):

PR #324, opened 2022-06-21
[0.2] Feat - Discounts

alecritson:
This is an initial PR to get discounts started in the core...


packages/core/src/DiscountTypes/ProductDiscount.php
```diff
@@ -0,0 +1,132 @@
...
+
+namespace GetCandy\DiscountTypes;
+
+use GetCandy\Base\DataTransferObjects\CartDiscount;
+use GetCandy\DataTypes\Price;
+use GetCandy\Facades\Discounts;
+use GetCandy\Models\CartLine;
+use GetCandy\Models\Discount;
+
+class ProductDiscount
```
glennjacobs, line 11: Probably needs an interface?

packages/core/src/Models/Cart.php
```diff
@@ -7,6 +7,7 @@
 use GetCandy\Base\Traits\LogsActivity;
 use GetCandy\Database\Factories\CartFactory;
 use GetCandy\DataTypes\Price;
+use GetCandy\Discounts\Models\DiscountReward;
```
glennjacobs, line 10: Think this is old code?

packages/core/src/Models/CartLine.php
```diff
@@ -42,6 +42,8 @@ class CartLine extends BaseModel
      */
     public ?Price $unitPrice = null;
 
+    public ?Discount $discount = null;
```
glennjacobs, line 45: Does this need a type? Not sure what it should be?

ryanmitchell:
I've had a quick look through this and it all looks promising so far. One use case that we would have would be setting an entire `brand` or `category` on discount... from what I'm seeing that would mean adding every purchasable in those brands, which could be unwieldy. Would it be an option to change how you work out what items the discount applies to, so that a developer can write some custom logic to determine that?

glennjacobs:
> I've had a quick look through this and it all looks promising so far. One use case that we would have would be setting an entire `brand` or `category` on discount... from what I'm seeing that would mean adding every purchasable in those brands, which could be unwieldy. Would it be an option to change how you work out what items the discount applies to, so that a developer can write some custom logic to determine that?

We did intend to allow Collections to be added also. As for Brands, I guess it could be possible, but we're just trying to get an MVP out there. You can make your own discount types if you need something specific.

packages/admin/resources/views/livewire/components/discounts/types/coupon.blade.php
```diff
@@ -0,0 +1,9 @@
+<div class="space-y-4">
+    <x-hub::input.group for="coupon" label="Coupon" instructions="Coupon codes are case insensitive">
+        <x-hub::input.text wire:model="discount.data.coupon" />
+    </x-hub::input.group>
+
+    <x-hub::input.group for="type" label="Percentage" >
+        <x-hub::input.text wire:model="discount.data.value" />
```
ryanmitchell, line 7: It would be great if this could also be a value, not just a percentage (eg £10 off your order, rather than 10%)

packages/core/src/DiscountTypes/Coupon.php
```diff
@@ -0,0 +1,113 @@
...
+    {
+        $currency = $cartLine->cart->currency;
+
+        $value = ($values[$currency->code] ?? 0) * 100;
+
+        if (! $value) {
+            return $cartLine;
+        }
+
+        $amount = (int) (round($value / $cartLine->cart->lines->count(), 2));
```
ryanmitchell, line 82: For VAT rules in the UK at least this should 'apportion' the value over the cart (work out what each line is as a percentage of the subtotal, and apply that amount of discount per line).

alecritson: @ryanmitchell Yeah this is something I came up against when I was writing it. I think it was too late on Friday for my head to work it out 😅  but I'm aware this is wrong at the moment as it would currently be impossible to do a £10 fixed amount for a cart with an odd number of lines...

ryanmitchell: Yeah its a complicated one. If me chipping in is annoying jut say and I'll hold back!

alecritson: Not at all :) happy for any help in any capacity, discounts are a bit of a beast!

adam-code-labx: Just pulled down the branch and merged with develop @alecritson this looks great so far 👍 
I assume there is no create component yet for the discounts? as the route goes to create product.

From my side looking to complete all hub features for the my client in the next 2-3 weeks.
So if you need any assistance with this discounts feature, please let me know as they use discounts for most of their products, so it's important to get something functional so their order totals match up with their current store.

markmead:
Looks good, but I found a few issues/concerns.

- When saving a discount it flashes the notification but immediately refreshes the page.
- Errors at the top of the form are not clear when trying to save a discount.

--- 



Few points here.

- Why is the "Priority" input so big?
- We need an explanation of what "Priority" is. It's not obvious, especially for non-developers.
- "Stop" is misleading. It looks like it will stop the discount.
- The word "propagating" is confusing, it needs a non-developer explanation. Something like...

> Checking this will prevent other discounts from being applied if the current discount criteria is matched.



The UI is a little wonky here. It looks like some classes are missing on the parent, possibly `p-2 space-y-2`.



- Show which discounts are active based on the date.
- Add sorting for date and status (active/inactive/scheduled).
- I don't think "Stop" is needed.



Adding a "Select All" could be handy, especially if there are lots of collections/brands.



The active button is invisible.



More of a generic issue across the hub, but the error icon overlaps the spinners in the number input.

glennjacobs:
## Feedback...

We want to split the edit screen into 3 sections. The general information, e.g. name, handle, type, etc. Then the conditions for the discount, and then finally the rewards. We should break it up into separate panes to indicate this, as it's quite a full-on form currently.

Coupons should be part of the generic conditions for a discount. A discount should either automatically apply or be applied by a matching coupon.

Discount types should be "Buy X Get Y" or "Discount".

## Buy X Get Y
This is the typical BOGOF type promotion. You can state which products you need to buy to qualify, and state which products you can get for free. We need to update the Cart to list the available free products (I did something before on a previous PR we closed). The free products should be grouped by the qualifying discount so that customers can easily apply them to their cart. If a customer already has a qualifying product in their cart, then they get it for free. If they remove it or don't apply a free product, then they simply won't get it.

## Discount
This gives a % off the cart if the conditions are met. It may only discount certain products in the cart, or all products in the cart. We currently allow for a fixed amount off, if we're doing that it will need to be multi-currency. We also need to be careful that fixed amounts don't make something too cheap, free or even a negative amount. I guess there is an onus on the store owner to ensure the logic is correct, but we certainly should never have a negative number.

## Misc. Feedback
The "min. cart amount" field is far too wide and it should also default to "0". In fact, there are a LOT of very wide inputs, which makes it hard to see what is expected. 

E.g. 
image

There are no conditions for channels and customer groups, we'll need those.

Do we need to consider reporting on how many times a discount has been used?

AlessandroAmos:
hello , I found an error in this branch ... it happens when I perform a clean install

error: 
"Foreign key constraint is incorrectly formed" 

explanation: 
the error happens because the "brand_discount" table asks for a primary key "brand_id" that does not yet exist

how I fixed: 
simply renamed the "brand_discount" migration to migrate after the "brands"

alecritson:
@glennjacobs I've pushed a bunch of updates to the UI. Hopefully you should be able to pull down without any issue, but if there are problems let me know. it's a WIP atm.

AlessandroAmos:
@alecritson  
the migration error still exists
but now because of the "discounts" table

image

to fix it I renamed the discounts migration `2022_11_18_100000_create_discounts_table.php` to `2022_06_16_100000_create_discounts_table.php`

image

AlessandroAmos:
a suggestion....

I believe that when there are several registered brands or collections, it can end up creating a rather large scroll, making it difficult for the end user

change this
image

to this: 
image

glennjacobs:
image
Migrations issue.

---

image
Can we put "Priority" under the name/handle pane?

---

image
Can we have more than 2 priorities? Even if it's just adding medium setting?

---

image
Use 2 columns for name and handle.

---

image
Nothing done to separate Collection Groups in the selection box.

---

image
Switching currency causes an error.

---

image
Do we need to say "Leave this blank for unlimited uses."?

---

Validation doesn't show what's missing/wrong. E.g. not adding any Qualifying Products.

glennjacobs:
image

Can it be

"Qualifying Products"

"Product Rewards"

?

glennjacobs:
image
Can we give the placeholder a grey background, perhaps along the lines of the image upload box? At the moment it looks too close to an input box.

glennjacobs:
image
Setting max uses and then unsetting it causes this error.

---

image
Changing between "Percentage" and "Fixed Amount" clears the "Minimum cart amount" for some odd reason.

---

image
Selecting a (minimal) product causes this error.

---

image
The placeholder "No products currently selected" doesn't have the `bg-gray-50` background color.

--- 

No validation error if no products are selected on a "Buy X Get Y" discount type.

--- 

Validation errors display in the Discount Type pane, but the menu on the right doesn't indicate this by showing the menu item in red.

--- 

image
No warning if a discount has no availability like products do.

glennjacobs:
No validation on "Percentage" or "Fixed Amount" inputs.

glennjacobs:
image
Validation issue when using fixed price. Weirdly I can't reproduce it again. 

---

Can we use the titles 

"Qualify Products" & "Product Rewards"

?

--- 

How about showing the discount type on the index datatable?

glennjacobs (approved):

Issue #325, opened 2022-06-22
CartLine will intermittently return an incorrect unit price


daikazu:
- GetCandy version: 2.0-beta13.1
- Laravel Version: 9.18.0
- PHP Version: 8.1.7
- Database Driver & Version:

### Expected Behaviour:
```php
$cartline->unitPrice->formatted(); //$9.12
```
### Actual Behaviour:
```php
$cartline->unitPrice->formatted(); //$9.11
```

### Steps To Reproduce:
I have a product variant with the price of $9.12. When added to cart it displays and calculates in the total as $9.11.

I have a forthcoming pull request to fix the issue.

alecritson:
This should be sorted in `0.1.0-rc.1`

PR #326, opened 2022-06-22
fix floating point precision error in unit price calc

daikazu:
fix for issue #325 

My apologies but I couldn't create a test for this at the moment.

### DIFF
--- a/packages/core/src/Actions/Carts/CalculateLine.php
+++ b/packages/core/src/Actions/Carts/CalculateLine.php
@@ -44,13 +44,10 @@ class CalculateLine
             );
         }
 
-        $decimalString = (string) round(
-            $price->decimal / $purchasable->getUnitQuantity(),
+        $unitPrice = (int) round(
+            (($price->decimal / $purchasable->getUnitQuantity())
+                * $cart->currency->factor),
             $cart->currency->decimal_places);
-
-        $factorString = (string) $cart->currency->factor;
-
-        $unitPrice = (int) bcmul( $decimalString, $factorString);
 
         $subTotal = $unitPrice * $cartLine->quantity;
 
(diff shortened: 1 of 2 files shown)


alecritson:
Thanks for submitting this @daikazu , although there is a link to the issue, can you explain what the change is doing in a bit of detail? Would this affect other currencies etc?

daikazu:
So I updated the pull request to a simpler solution which does not require someone to have the PHP BC Math Functions extension. Essentially I moved the rounding of the unit_price last before it is cast to an int. You should generally always round at the very end when doing any kind of monitory calculations like this. When selling in large volumes loosing a penny on the unit price really adds up over time.

I extracted and created a little [example ](https://3v4l.org/P06DF#v8.1.7) demonstrating the problem and showing the solution results. I hope that helps. I've tested it with other decimal places and it seems to work correctly.


Not necessarily within in the same scope of this fix but I also see a potential problem where you are getting a unit price based off of `$price->decimal / $purchasable->getUnitQuantity()` within the same piece of code.  If that get's rounded off and then multiplied later by the quantity, the results are going to be off also.

daikazu:
Hey @alecritson,  I added a test that addresses the initial rounding issue. I used a known failing value because creating a big loop to test through would chug and use a lot of memory because it requires adding a lot of lines to the cart. I can add the looping checks if needed though.

This still doesn't address the problem if the ProductVariant unit_quantity is more than 1. When I have a few moments I'll dig a little deeper into the code to see what all needs to be done to address this, create an issue if was does not already exist, and work up a fix for it. I'm sure just moving all the decimal rounding for currency to just before the output would solve it.

alecritson:
Thanks @daikazu that makes sense. I'll have a chat with @glennjacobs and see if we can get this in the next release. Thanks again for your work on this!

glennjacobs (approved):

PR #316, opened 2022-06-10
Fix directory path (remove slashes)

iboldurev:

PR #327, opened 2022-06-23
Fix addons manifest getting wrong path on addons

skyrowl:
#328


alecritson:
thanks for this @skyrowl , looks like it could be a dupe of #316 ?

alecritson:
Thanks @skyrowl. After having a look, not all add-ons will have their providers within a `Provider` namespace, I think I'll close this in favour of the existing PR but I appreciate you taking the time with this.

PR #330, opened 2022-06-24
Feat - Stock level on order lines

alecritson:
This PR will add the current stock level and stock level at purchase, if supported.


glennjacobs:
Missing change log entry for admin hub. Looks good otherwise I think.

alecritson:
> Missing change log entry for admin hub. Looks good otherwise I think.

Added!

glennjacobs (approved):

PR #332, opened 2022-06-24
Feat - Refactor how activity log is displayed

alecritson:
Currently, the activity log is rendered on the front end with a bunch of if statements (on orders). This isn't really flexible and doesn't allow developers to tap into it.

This PR aims to fix this...

The idea is that you can register your own renders for specific events that appear in the activity log. This means if you log something on your storefront, you can add a rendering class for that event.

## Defining the class

```php
<?php

namespace App\ActivityLog;

use GetCandy\Hub\Base\ActivityLog\AbstractRender;
use Spatie\Activitylog\Models\Activity;

class CustomRender extends AbstractRender
{
    public function getEvent(): string
    {
        return 'created';
    }

    public function render(Activity $log)
    {
        return view('activity-log.custom-render', [
            'log' => $log,
        ]);
    }
}
```

## Registering

```php
use GetCandy\Hub\Facades\ActivityLog;

// We want to add our `CustomRender` to the activity log on orders.
ActivityLog::addRender(Order::class, CustomRender::class);
```

So this means on the Order screen, we'll loop through the activity log, find any renders for each event and display them.


glennjacobs:
The idea looks neat, see no problems there. Needs some docs and maybe some tests?

glennjacobs (approved):

PR #333, opened 2022-06-24
Feat - order notifications

alecritson:
Initial work on bringing in order notifications to the hub

---

This PR will also rely on #332


glennjacobs:
Needs a change log entry and example mailable to show how content is injected.

docs/src/getcandy/orders.md
```diff
@@ -309,3 +309,27 @@ Once updated, GetCandy will keep a render of the email sent out in the activity
...
+Here's an example of what the template could look like:
+
+```html
+<h1>It's on the way!</h1>
+
+<p>Your order with reference {{ $order->reference }} has been dispatched!</p>
+
+<p>{{ $order->total->formatted() }}</p>
+
+@if($content ?? null)
```
glennjacobs, line 327: Is `??` correct here?

alecritson: Yeah cause if the email is sent via something outside of the order notification, it won't be set potentially.

docs/src/getcandy/orders.md
```diff
@@ -281,6 +281,57 @@ In terms of an order, all it's worried about is whether or not the `placed_at` c
 
 And as always, if you have any questions you can reach out on our Discord!
 
+## Order Notifications
+
+GetCandy allows you to specific what Laravel mailers/notifications should be available for sending when you update an order's status. These are configured in the `getcandy/orders` config file and are defined like so:
```
glennjacobs, line 286: "GetCandy allows you to specific" < should be "specify"?

docs/src/getcandy/orders.md
```diff
@@ -281,6 +281,57 @@ In terms of an order, all it's worried about is whether or not the `placed_at` c
...
+            App\Mail\MyMailer::class,
+            App\Mail\MyOtherMailer::class,
+        ],
+        'notifications' => [],
+    ],
+    // ...
+],
+```
+
+Now when you update an order's status in the hub, you will have these mailers available if the new status is `awaiting-payment`. You can then choose the email addresses which should be sent out and also add an additional email address if required.
```
glennjacobs, line 304: "email addresses which should be sent out" < change to "sent to"?

glennjacobs (changes requested):
Just some doc changes as highlighted, then good to go.

glennjacobs (approved):

PR #336, opened 2022-07-01
update/login-ui

markmead:
This PR updates the login and reset password UI.

Added - New translations
Added - Stamp component
Added - Loading indicator component
Updated - Login form UI
Updated - Reset password form UI
Updated - Use 
Removed - Stamp partial
Fixed - Notifications delay between removing

I've dropped the use of some form components for these updates, I didn't want to make changes that would have any side effects on the rest of the hub UI. Probably best to make PRs for those updates.


markmead:
Failing tests are odd, they're passing on local... 🧐

PR #342, opened 2022-07-08
Feat - Publishable migrations

### LINKED ISSUE #321: Publish migrations
When using multi-tenancy packages the migrations need to be available so they can be run on each tenant database. 

Currently migrations cannot be published, which makes things difficult.

alecritson:
Fixes #321


glennjacobs:
I've never tried, but not sure if you can register migrations and also make them publishable. I don't know if Laravel will complain or handle it nicely. Probably need to check. Be cool if does handle it and the app's migrations override the package's.

alecritson:
@glennjacobs Yeah it seemed to work just fine

glennjacobs (approved):

nathandaly:
Legends...

PR #343, opened 2022-07-08
Feat - Remove saved carts

alecritson:
This PR removes the Saved Cart functionality and adds upgrade docs for users to keep if they're using them.


glennjacobs:
Might want to update the tests 😆

glennjacobs:
This will have to wait a few releases, as we need to release migration publishing first. Either that, or we update the upgrade guide so they just copy/paste the migration into their app?

glennjacobs (approved):

PR #347, opened 2022-07-12
Hotfix - Fix macro registering

### LINKED ISSUE #346: Macros with the same name defined on separate models affect eachother
- GetCandy version: latest
- Laravel Version: 9.16
- PHP Version: 8

### Expected Behaviour:
Macros with the same name but on different models should not affect eachother. 

### Actual Behaviour:
Macros with the same name defined on different models overwrite the previous model. 

### Steps To Reproduce:
```
Product::macro('getPermalink', function () {
            return route('product.show', $this->defaultUrl->slug);
        });
        
        Collection::macro('getPermalink', function () {
            return route('collection.show', $this->defaultUrl->slug);
        });
        ```

### PR DESCRIPTION
**Pull request by alecritson**
Looking to diagnose and fix #346

### CODE REVIEW

**Comment by glennjacobs** [comment]
@alecritson looks good. Need a changelog entry and to set as review for review.

File: docs/src/upgrading.md [hunk 1]
```diff
@@ -32,6 +32,26 @@ php artisan getcandy:meilisearch:setup
 
 ## 2.0-beta14
 
+### Removal of Macro functionality from BaseModel - Low Impact
+
+If you have custom models that extend the GetCandy `BaseModel` you will need to implement the new `HasMacros` trait.
```
glennjacobs, line 37: they probably don't in all honesty. we should probably say if you're using the macroable functionality.

glennjacobs (approved):

PR #353, opened 2022-07-14
Feat - Tax management

alecritson:
This PR looks to add Tax management screens to the Admin Hub.


MiguelFCode (changes requested):
$storeSection->addItem is now $slot->additem .Doesn't really work or save , cant create either ( it does read correctly from the database tho)

alecritson:
> $storeSection->addItem is now $slot->additem.

`$storeSection->addItem` is used for the menu, not slots, so this appears fine.

> Doesn't really work or save , cant create either

I have been able to create and save tax zones with rates etc. Would need more information for this one.

glennjacobs (approved):

PR #358, opened 2022-07-15
Hotfix - Add a bunch of toolbar options to the RichText editor

### LINKED ISSUE #335: Can't insert videos in description
### Expected Behaviour:

That the Rich Text field would allow for videos or video urls to be added so they can be embedded in content.

### Actual Behaviour:

Its not possible to add videos.

### Steps To Reproduce:

Should hopefully be self-evident :)

alecritson:
Fixes #335 

---

I appreciate we want to make this configurable potentially, however not entirely sure about the best approach, so thought this will suffice until we can figure that out.

### DIFF
--- a/packages/admin/resources/views/field-types/text/settings.blade.php
+++ b/packages/admin/resources/views/field-types/text/settings.blade.php
@@ -6,3 +6,20 @@
 >
   <x-hub::input.toggle :disabled="!!$attribute->system" :on-value="true" :off-value="false" wire:model="attribute.configuration.richtext" id="fieldType" />
 </x-hub::input.group>
+
+@if($attribute->configuration['richtext'] ?? false)
+  <div class="space-y-4 mt-4">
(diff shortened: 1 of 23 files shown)


packages/admin/resources/views/field-types/text/settings.blade.php
```diff
@@ -6,3 +6,17 @@
 >
   <x-hub::input.toggle :disabled="!!$attribute->system" :on-value="true" :off-value="false" wire:model="attribute.configuration.richtext" id="fieldType" />
 </x-hub::input.group>
+
+@if($attribute->configuration['richtext'] ?? false)
+  <div class="space-y-4 mt-4">
+    <p class="text-sm">RichText fields use <a href="https://quilljs.com/docs/configuration/" class="text-blue-500">Quilljs</a>, you can you enter any available configuration below.</p>
```
glennjacobs, line 12: Not translated

glennjacobs (changes requested):
Translations need adding

glennjacobs (approved):

PR #405, opened 2022-07-15
Extend settings to manage product features - Part 1

adam-code-labx:
Here is my 1st PR extending settings which includes (small refactor of SettingsMenu) 

Requested by our client, previously using Prestashop they could manage a global list of attributes and features.
I have created a new menu group under settings called "Product" This new addition in this 1st PR called "Features"

Currently this is only for features. I would like to extend on this to add another new menu item for attributes which will be called "Options" to again handle the global list and be able to define each product option (group) field type for example colour picker.  This way our client can manage a global list of translatable features and values.

Additionally and will be created next week in another PR.
I plan to extend the ProductType Attributes field currently you have List type.
This is where user enters their values, which is great for some custom information on their site.

I will propose a new Dynamic List field type where you can choose a model.
For example if the model selected is ProductFeature then the user will be able to select a feature which will automatically populate the list with the feature values which can then be sorted within the ProductType page and the possibility to override via the product page.

I have created a basic set of tests for this PR.
Please let me know if any further tests are required.

Also I can put another PR together for the demo site to include seeders for this new feature if required.

### DIFF
--- a/packages/admin/resources/lang/en/components.php
+++ b/packages/admin/resources/lang/en/components.php
@@ -179,4 +179,49 @@ return [
     'activity-log.orders.capture'                   => 'Payment of :amount on card ending :last_four',
     'activity-log.orders.authorized'                   => 'Authorized of :amount on card ending :last_four',
     'activity-log.orders.refund'                   => 'refund of :amount on card ending :last_four',
+
+    /**
+     * Product Features
+     */
+    'product.features.no_feature_value_text' => 'No feature values exist, either drag existing feature value or add new ones here.',
+
+    /**
+     * Product feature create.
+     */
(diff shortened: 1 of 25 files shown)


glennjacobs:
Can you explain what this provides in terms of value and how you would use it in real world examples?

glennjacobs:
Also, I would suggest we keep each issue/pr to a single feature. You may have noticed we have overhauled the issues section now so you can propose enhancements and new features. Ideally, we'd start with issues and then link PRs to them.

glennjacobs:
Let's get an issue created for each suggestion and we can chat them over there. Cheers.

adam-code-labx:
@glennjacobs just opened an issue to continue discussion here thanks #406

PR #414, opened 2022-07-22
Dynamic attribute group models

adam-code-labx:
Currently adding a custom attribute group via the settings for example `attributes/product` allows you to add a group and then manually set the values. This is a great and gives quite a lot of flexibility to add custom fields.

This small PR extends this feature to support dynamic groups.
In the example below I am using this for Product Options & Collections

The following models are supported (Note: ProductFeature currently not enabled until #405 has been approved)
```php
public function getModelsCollectionProperty(): Collection
{
    return collect([
        '--' => 'Select a model',
        'collection' => CollectionModel::class,
        //'features' => ProductFeature::class,
        'options' => ProductOption::class,
    ]);
}
```

Note: Selecting collections will automatically populate all your collection groups.
Another use case for example could be BlogPostModel which will allow you to display a post either across a product type or a post, the idea is you will be able to use any model currently manually defined in ModelsCollectionProperty.

The next PR shortly which will be linked below will enable you to use these new dynamic groups for ProductType.
This will include an example of managing both Product Options & Product Features where different configurations can be setup for each product type.

Screenshot 2022-07-22 at 11 52 23
Screenshot 2022-07-22 at 11 45 44

### DIFF
--- a/packages/admin/resources/lang/en/components.php
+++ b/packages/admin/resources/lang/en/components.php
@@ -42,7 +42,7 @@ return [
     'attribute-group-edit.name.placeholder'     => 'e.g. Additional Details',
     'attribute-group-edit.create_btn'           => 'Create attribute group',
     'attribute-group-edit.update_btn'           => 'Update attribute group',
-    'attribute-group-edit.non_unique_handle'    => 'This name has been already been taken',
+    'attribute-group-edit.non_unique_handle'    => 'This name has already been taken',
     /**
      * Attribute show.
      */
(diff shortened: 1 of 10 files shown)


glennjacobs:
This doesn't make any sense to me, I'm sorry. 

Attribute Groups are for grouping attributes. Why would we put product options (and features) in them?

I could understand adding "product options" to "product types", so product options can be pre-filled for speed.

I think I need to understand the problem you're trying to solve here, so I can offer a better opinion on the solution.

adam-code-labx:
@glennjacobs Closing this one will update ProductType PR tomorrow to include Options and Features as hard coded values for now. I see your point and believe there are other ways to achieve the same dynamic results which we can look at later with the refactor.

PR #422, opened 2022-07-27
Feat - Add copy to clipboard button for order references

alecritson:


glennjacobs:
Ony appears to be for the order reference, I couldn't see the customer reference in there?

alecritson:
> Only appears to be for the order reference, I couldn't see the customer reference in there?

Done

glennjacobs (approved):

PR #423, opened 2022-07-27
Feat - Add date format for orders in the hub.

alecritson:
This PR looks to add a date format config setting for orders across the hub.


glennjacobs:
Should this format only be for orders? I figured it could be used any where in the hub.

alecritson:
> Should this format only be for orders? I figured it could be used any where in the hub.

Closed to start again since if that's the case will need more thought with how javascript/php format diffs will work

PR #428, opened 2022-07-27
Feat - Move order reference generator to config

alecritson:
This PR looks to move the Generator for Order references out of a callback approach and into the config files to be more in line with how this will be in future releases.

### DIFF
--- a/docs/src/upgrading.md
+++ b/docs/src/upgrading.md
@@ -128,6 +128,11 @@ Finally, you will need to define a dynamic relationship if your service provider
 });
 ```
 
+### Removal of `override` method for OrderReferenceGenerator - Medium Impact
+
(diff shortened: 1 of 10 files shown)


docs/src/upgrading.md
```diff
@@ -128,6 +128,11 @@ Finally, you will need to define a dynamic relationship if your service provider
 });
 ```
 
+### Removal of `override` method for OrderReferenceGenerator - Medium Impact
+
+If you're using the `override` method to generator your own order references, this has been removed in favour of a config based approach.
```
glennjacobs, line 133: Do we need to get devs to update their config file to match?

alecritson: We default to our one, so if they need to customise they can follow the link to add their own config, put the link in to save duplicating the content.

glennjacobs (approved):

PR #438, opened 2022-07-29
Feat - Customer impersonation

alecritson:
This PR starts the initial work for allowing developers to implement user impersonation into their storefront.


glennjacobs:
Do you not think we should have a default driver for standard Laravel auth? I guess 90% of devs will be using the standard set-up.

alecritson:
> Do you not think we should have a default driver for standard Laravel auth? I guess 90% of devs will be using the standard set-up.

It would require GetCandy providing a route, middleware and session logic for it to do the impersonation. Which seemed a bit wrong over just providing the means to allow dev's to implement it however they want.

glennjacobs (approved):

PR #443, opened 2022-07-29
[0.2] Allow extending of validation rules

wychoong:
this is a proof of concept on customising validation rules in hub. Current only applied for Product's collection.

now you can define more rules in config

```php
# in config/getcandy-hub/product.php
<?php

use GetCandy\Models\CollectionGroup;

return [
    # ... 

    'validation' => [
        'variant.sku' => 'required|unique|min:8',
        'collections.*.group_id' => [
            'required',
            'distinct',
        ],
        'collections' => ['required' => true, 'array', function ($attribute, $value, $fail) {
            $missing = [];

            $collections = collect($value)->pluck('id', 'group_id');

            $groups = CollectionGroup::whereIn('id', [1, 2])
                ->select(['id', 'name'])
                ->get()
                ->pluck('name', 'id');

            $missing = $groups->diffKeys($collections)->all();

            if (filled($missing)) $fail("Require Collection Group from: " . implode(', ', $missing));
        }],
    ]
];
```

**result**
image

image

**issue**
the main problem with the current implementation is unable pass in Model instance to those additional rules. it can be done but I don't think is good approach.


wychoong:
@glennjacobs need your input on this

this is referring to this https://discord.com/channels/497326121441034240/916819091841486869/1001891823146520646

glennjacobs:
@alecritson thoughts on this?

alecritson:
@glennjacobs I'm not against this but don't feel like the config is the correct place. Due to some validation using closures (like in the example above) this means that Laravel wouldn't be able to serialise the config for optimisation.

Would it not be better to provide a method on the component that dev's can tap into at the component level? Something like:

```php
ProductEdit::extendValidation([
    'name' => 'required|min:50',
]);
```

wychoong:
> @glennjacobs I'm not against this but don't feel like the config is the correct place. Due to some validation using closures (like in the example above) this means that Laravel wouldn't be able to serialise the config for optimisation.
> 
> Would it not be better to provide a method on the component that dev's can tap into at the component level? Something like:
> 
> ```
> ProductEdit::extendValidation([
>     'name' => 'required|min:50',
> ]);
> ```

Fair point. Will look into this approach

wychoong:
now can extend validation by using component method

```php
ProductCreate::extendValidation(
            [
                'variant.sku' => ['required', 'min:8'],
                'collections' => ['required', 'array', function ($product) {
                    return function ($attribute, $value, $fail) use ($product) {
                        $missing = [];

                        $collections = collect($value)->pluck('id', 'group_id');

                        $groups = CollectionGroup::whereIn('id', [1, 2])
                            ->select(['id', 'name'])
                            ->get()
                            ->pluck('name', 'id');

                        $missing = $groups->diffKeys($collections)->all();

                        if (filled($missing)) $fail($product->translateAttribute('name') . " Require Collection Group from: " . implode(', ', $missing));
                    };
                }],
            ]
        );
```

$product is able to pass from component by using `app()->call(...)`
refer: `getExtendValidation` method

alecritson:
@wychoong It's looking promising, I can see it's still in draft, is there still work you want to do on it?

wychoong:
> @wychoong It's looking promising, I can see it's still in draft, is there still work you want to do on it?

hey. I think in terms of implementation it is done. I kept it as draft is to get feedbacks and since this opens up a lot more validation, we also need to add few more changes to display the error message.
in
- `getSideMenuProperty` `hasError`
- `@error` in blade

packages/admin/src/helpers.php
```diff
@@ -12,20 +12,20 @@ function max_upload_filesize()
     }
 }
 
-if (! function_exists('get_validation')) {
+if (!function_exists('get_validation')) {
```
alecritson, line 15: Is this still needed since we're not going through the config anymore?

glennjacobs: @wychoong does this still need to be sorted?

glennjacobs:
@wychoong should we have documentation for this?

wychoong:
> @wychoong should we have documentation for this?

added doc

wychoong:
> > > @wychoong should we have documentation for this?
> > 
> > 
> > added doc
> 
> Awesome. Just Alec's question and then I think this can be merged?

hrmmm... you referring to this?
image

not sure why its pending or should i click the resolve converstation?

wychoong:
figured it out that i need to submit my "review" comment 😓

glennjacobs:
@wychoong the VuePress config file has conflicts now. You have changed the indentation from 2 spaces to 4 spaces, any reason for this?

wychoong:
> @wychoong the VuePress config file has conflicts now. You have changed the indentation from 2 spaces to 4 spaces, any reason for this?

I think that was because of my VSCode does that, can you help on this as I’m away from computer till the weekend

wychoong:
@glennjacobs it's ready for review

packages/admin/src/Http/Livewire/Components/Products/AbstractProduct.php
```diff
@@ -245,9 +249,12 @@ protected function rules()
         return array_merge(
             $baseRules,
             $this->hasImagesValidationRules(),
+            // $this->withAttributesValidationRules(),
```
glennjacobs, line 252: Commented code?

glennjacobs (approved):

Issue #442, opened 2022-07-29
Side menu rendering not smooth in updated UI


glennjacobs:
@markmead this is for you.

When I click around the new admin hub UI (with the menu expanded), the side menu isn't very smooth. It seems it initially renders collapsed and then a few ms later it pops open. Can you work your magic?

markmead:
It's because we're using `defer` on the Alpine JS script which is the recommended approach.

As this video highlights, the best result is when removing `defer` but that's not really an option in my opinion. There's some improvement when Alpine JS is bundled but it's not perfect.

https://user-images.githubusercontent.com/50486078/182105870-9acefb34-28e4-48e4-8066-85c2ab5e5c2b.mov

glennjacobs:
@markmead I'm not an expert on this subject, so can't really comment. But currently, I would say it's not acceptable, so we'll need to find a way to resolve it.

@alecritson do you have any ideas?

alecritson:
@glennjacobs Not too sure what the perfect solution is really. As it stands we gotta wait for Alpine's storage to kick in and add the correct classes. Only thing I can think of atm is storing the preference in a cookie and grabbing it server side so we can set the correct class on render, before Alpine takes over.

glennjacobs:
@alecritson yeah, I think that's correct. It really needs to be server-side.

alecritson:
@glennjacobs Currently we don't have anything in the hub to set this kind of info and retrieve it, I guess we can use the `GetCandyHub` ? 🤔 

Then we'd probably need a bit of a rewrite of the menu so we can tap into Livewire (since they're just layout files atm) so not sure this would be a "quick fix" if you're wanting it for the next release.

alecritson:
This should be sorted in the latest release.

PR #445, opened 2022-08-02
Side menu fix

skyrowl:
Fix of sidebar menu stuttering on page load and tooltips not showing when hovering reduced menu

- Dropped Alpine persist plugin
- Removed unnecessary Alpine directives
- Added vanilla JS to manage side menu and dark mode through local storage
- Added CSS classes for the above changes

#442


packages/admin/resources/views/layouts/app.blade.php
```diff
@@ -22,22 +22,49 @@ class="h-full">
...
+            display: none !important;
+        }
     </style>
 
-    <script defer
-            src="https://unpkg.com/@alpinejs/persist@3.x.x/dist/cdn.min.js"></script>
     <script defer
             src="https://unpkg.com/alpinejs@3.8.1/dist/cdn.min.js"></script>
+
+    <script>
```
alecritson, line 38: Looks like quite a lot is happening here and there's some repetition in the overall logic. Could this be added to the `hub.js` file so it's compiled and can be reused?

markmead:
I don't think we should be straying too far from the current implementation, we just need a way to handle the side menu `width` before Alpine JS initializes. After that Alpine JS can manage the rest.

alecritson:
Thanks @skyrowl We're going to go with this pr #446 for the time being, if we need to look at dark mode then we can do that in a later release. Thanks for taking a look though.

packages/admin/resources/views/layouts/app.blade.php
skyrowl: Not really, it must be it must be executed before the DOM starts loading else the stutter will come back even for a second or less but still visible.
Repetition can be removed tho it is to make sure that the value is either "true" or "false" (localStorage is treated as a string even for booleans) so an error through JSON.parse() can be avoided if the value is a string other than these two values (JSON.parse can't be used with a literal string that is not a bool).

markmead: Don't need to worry about passing wrong data to `JSON.parse` as Alpine JS will always be sending either `true` or `false` and as you say, local storage treats it as a string 😄

PR #446, opened 2022-08-03
fix/side-menu

markmead:
This PR fixes the side menu rendering when the page refreshes.


packages/admin/resources/views/partials/navigation/side-menu.blade.php
```diff
@@ -3,7 +3,8 @@
          :class="{
              'w-64': showExpandedMenu,
```
alecritson, line 4: Looks like we have a custom class to handle it now, should those be reflected here?

markmead: It's still needed, there's no custom class being used.

alecritson: @markmead  What's `app-sidemenu-expanded` then?

markmead: It's added to the `<html>` so we can target the side menu in CSS.

Then once Alpine JS inits, we remove it and allow Alpine JS to manage the size of the side menu.

alecritson:
@markmead Looks reasonable I think and much simpler @glennjacobs Just need your nod on this

glennjacobs (approved):

PR #451, opened 2022-08-03
fix/sidemenu-tooltips-css

skyrowl:
- Removed the use of Alpine on tooltips for tailwind


markmead:
Haha, we did the exact same thing! #450 Let's use yours! 🥳

Can we remove the `transition opacity-0 group-hover:opacity-100` on the tooltips? It doesn't look right.

skyrowl:
> Haha, we did the exact same thing! #450 Let's use yours! 🥳
> 
> Can we remove the `transition opacity-0 group-hover:opacity-100` on the tooltips? It doesn't look right.

Glad that we thought of the same solution !

Sure we can it was just for a smooth apparition it's not useful at all.

packages/admin/resources/views/livewire/sidebar.blade.php
```diff
@@ -26,10 +24,7 @@ class="text-sm font-medium">
                             {{ $item->name }}
                         </span>
 
-                        <span x-cloak
-                              x-transition
-                              x-show="showTooltip"
-                              class="absolute z-10 p-2 ml-4 text-xs text-center text-white bg-gray-900 rounded dark:bg-gray-800 w-28 left-full">
+                        <span class="invisible group-hover:visible absolute z-10 p-2 ml-4 text-xs text-center text-white bg-gray-900 rounded dark:bg-gray-800 w-28 left-full">
```
markmead, line 27: Can we move `group-hover:visible` to the end of the class names?

packages/admin/resources/views/livewire/sidebar.blade.php
```diff
@@ -58,10 +51,7 @@ class="text-sm font-medium">
                     {{ __('adminhub::global.settings') }}
                 </span>
 
-                <span x-cloak
-                      x-transition
-                      x-show="showTooltip"
-                      class="absolute z-10 p-2 ml-4 text-xs text-center text-white bg-gray-900 rounded dark:bg-gray-800 w-28 left-full">
+                <span class="invisible group-hover:visible absolute z-10 p-2 ml-4 text-xs text-center text-white bg-gray-900 rounded dark:bg-gray-800 w-28 left-full">
```
markmead, line 54: And the same here, please.

markmead (commented):
Looks good to me! 🚀

Just two tiny changes that will save my OCD from keeping me up at night 😅

markmead (approved):
Happy with this! Any thoughts @alecritson?

alecritson:
@markmead nothing from me. @glennjacobs If you're happy, seems like a small tweak.

glennjacobs (approved):

Issue #408, opened 2022-07-15
Suggestion to refactor livewire pages and components


adam-code-labx:
Currently for any new livewire components these need to be registered in `AdminHubServiceProvider`
There is a better way. Although livewire does not currently support registering outside the app namespace.

Dan has come up with a great solution here to autoload components which can be used as a trait 
https://gist.github.com/danharrin/0e92e248c94e195a3f444d8745820c7f

He also uses for filament directly in the service provider
https://github.com/filamentphp/filament/blob/2.x/packages/admin/src/FilamentServiceProvider.php

Once the components can be auto registered in this way.
I would suggest possible cleanup for the page and components structure in the hub as there does seem to be a lot of duplication at least in the component names. in fact do we even need Livewire\Pages  
Can we not call the component directly from the route then apply the layout?

Let me know if you would like me to work on a PR for this one.
Thanks.

glennjacobs:
The autoloading was something we wanted to look at, as the list of registered components will only get bigger and bigger.

I would welcome a draft PR to suggest an approach.

PR #456, opened 2022-08-06
Livewire component autoload

adam-code-labx:
Referenced to #408 

@glennjacobs @alecritson  
You can now autoload components without having to register via AdminHubServiceProvide.
This is the 1st step before my refactoring suggestions for the livewire page and components.

Not all the components follow the same naming convention so I have added a check to override the alias if overrideComponentAlias exists in a livewire component class.


adam-code-labx:
All tests are passing locally the test runner seems to be the issue here.

There were 325 errors:

1) GetCandy\Hub\Tests\Feature\Http\Livewire\Pages\Settings\Products\CollectionGroupIndexTest::cant_view_page_as_guest
ErrorException: Undefined array key -1

alecritson:
@adam-code-labx Looks like doctrine/dbal did an update and it broke something. I've locked it to the last working version for now, could you update your branch and hopefully it'll solve it.

adam-code-labx:
Thanks @alecritson merging develop branch solved the issue. Let me know what you think about this approach to auto loading.

I started from Dan’s trait but made a lot of changes in order to get this working across different namespaces.

alecritson:
@adam-code-labx I think it's a good start and handy if we're taking inspiration from something that already has it in place. Would save a lot of aggro and maintenance. We already reference components in the hub code quite a lot, does this preserve that?

adam-code-labx:
@alecritson yes you will see currently I have kept the global and auth components.
Those could also be auto-loaded by setting overrideComponentAlias name is the component for backwards compatibility.

```php
// @todo We should be able to autoload these components, otherwise update the alias for these ones.
$this->registerGlobalComponents();
$this->registerAuthenticationComponents();
```

Once you get chance please check the autoload branch your side.
During my component refactor #467  I also registered the page components, which is not necessary here as we are currently using hub.components for the interaction and updates. But you can see how this can be applied to any component directory or namespace as long as the component alias prefix are consistent.

```php
$this->registerLivewireComponentDirectory(
     directory: __DIR__.'/Http/Livewire/Pages',
     namespace: 'GetCandy\\Hub\\Http\\Livewire\\Pages', aliasPrefix: 'hub.pages',
);
```

adam-code-labx:
@alecritson Have you had chance to check this on your side yet I will mark it now for review.
Would be good to get this merged over to develop next week.

Anything else you need on this one please let me know.
Thanks.

packages/admin/src/AdminHubServiceProvider.php
```diff
@@ -194,14 +196,21 @@ protected function registerMenuBuilder()
...
         $this->registerGlobalComponents();
         $this->registerAuthenticationComponents();
-        $this->registerProductComponents();
-        $this->registerCollectionComponents();
-        $this->registerReportingComponents();
-        $this->registerSettingsComponents();
-        $this->registerOrderComponents();
-        $this->registerCustomerComponents();
+
+        // $this->registerProductComponents();
```
glennjacobs, line 208: Commented code, this would need to be removed before this could be merged.

packages/admin/src/AdminHubServiceProvider.php
```diff
@@ -388,10 +397,24 @@ protected function registerPermissionManifest()
...
+    /**
+     * This list is dynamically updated when new components are added in order to work with IDE completion.
+     *
+     * @note Please do not manually register components here
+     *
+     * @return void
+     */
+    private function _cachedLivewireComponents(): void
+    {
+        // @todo Supports Laravel Idea (livewire component ide support - needs to be moved to a separate service provider class
```
glennjacobs, line 417: Again, more code commented out. I don't feel like this is ready, so will change the status of the PR to draft for the moment.

adam-code-labx:
@glennjacobs 
might be worth holding back doing any more with the autoload as Livewire v3 will be released soon not sure if auto loading outside the app namespace is solved I guess we will have to wait and see 😅

glennjacobs:
Closing this, as it's not longer relevant with our plans to move to Filament v3.

PR #459, opened 2022-08-08
Make sure the correct attributes show when changing product type

ryanmitchell:
When changing the product type on a product that only had 1 variant the attribute fields were not updating correctly as `$this-> variantAttributes` was only set on mount().

This PR updates the logic to ensure the right attributes are parsed.


alecritson (approved):

alecritson:
@ryanmitchell This looks good thanks for sorting, could you just put a quick change-log entry under the admin package? Feel free to credit yourself.

ryanmitchell:
Yep - thats done

alecritson (approved):
Perfect! Thanks for sorting.

PR #461, opened 2022-08-08
Feature - Data tables

alecritson:
This PR looks to introduce Filamentphp's tables package for listing pages where it makes sense.


.nova/Configuration.json
```diff
@@ -0,0 +1,3 @@
+{
+  "index.enabled" : true
```
glennjacobs, line 2: Do we want Nova files in the repo?

alecritson:
@adam-code-labx @glennjacobs I think I've got the main ones we would want covered, happy for this to be given a test run to see if it holds up or if anything is missed.

packages/admin/src/Http/Livewire/Components/Tables/ProductsTable.php
```diff
@@ -0,0 +1,112 @@
...
+        ];
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    protected function getBaseTableBulkActions(): array
+    {
+        return [
+            BulkAction::make('delete')
```
adam-code-labx, line 81: @alecritson We need to ask for confirmation when deleting
`->requiresConfirmation()`

`DeleteBulkAction` looks like there is a filament issue with this one.
"Call to undefined function Filament\locale_has_pluralization()"

alecritson: Good spot, will make sure this gets added!

packages/admin/src/Http/Livewire/Components/Tables/ProductsTable.php
```diff
@@ -0,0 +1,112 @@
...
+        ];
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    protected function getBaseTableActions(): array
+    {
+        return [
+            ActionGroup::make([
```
adam-code-labx, line 68: If we can delete in bulk we should also have an action to delete a single record 😉
Perhaps you can add these as default to the GetCandyTable class?

alecritson: Thinking about delete methods on products, I'm in two minds now since we have the Danger zone on the product where you're meant to be able to confirm you want it deleted by typing in text. Not sure if the data tables supports this?

adam-code-labx: Yeah you can certainly customise the confirmation component to support the danger zone.
In terms of batch delete perhaps not a good idea for products or any entity which implements danger zone

alecritson: Yeah after reflection, I removed the bulk delete action. I think if someone really wanted it they could make a conscious choice to add it.

wychoong:
will this be merged soon? kinda need the Brand PR to be ready

glennjacobs:
Having tested the performance and flexibility of Filament tables, we've decided that they are not suitable for this project.

This isn't a dig at Filament, we ❤️ Filament, but this project demands a lot more than a generic datatable component can provide. 

However, do not fear! We have an awesome solution in the pipeline, more to be announced soon.

adam-code-labx:
@glennjacobs 
Do you mind sharing what performance issues you came across?

glennjacobs:
> @glennjacobs Do you mind sharing what performance issues you came across?

Discuss it on Discord if that's OK?

PR #466, opened 2022-08-09
Feat - Add named parameters to `savePricing` method on trait.

alecritson:
When using the `HasPrices` trait on Livewire components, you may not always want to use the pricing properties when saving pricing. This PR adds named parameters so developers can chose which pricing they want to save.


glennjacobs:
@alecritson do we need a test for this?

alecritson:
> @alecritson do we need a test for this?

Added a tiny test

glennjacobs (approved):

PR #421, opened 2022-07-27
Continue sidebar menu grouping feature

adam-code-labx:
@glennjacobs I can see this one was started so I have continued this grouping feature.
If we can't make the next beta release then please include in the new develop branch to avoid switching branches 😂

PR #467, opened 2022-08-09
Component suggested refactor

adam-code-labx:
This is the 2nd PR (WIP) in relation to #408 
This includes merges from branches (#456 #421 #461)

I don't think we need the duplication of the show edit index components.
Also moved ProductType to the pages root as I believe it doesn't really suit any purpose in Products component namespace

Pages and components are now auto-loaded so I feel this is a better direction to remove those duplicate components.

Also I have introduced abstract resource classes and PageComponent.
One of the reasons is to be able to define the navigations labels icon and group instead of having too hard code them.
If this sounds like a good idea I will continue on this and dynamically generate the menus.

@alecritson @glennjacobs These are just suggestions at this stage and I have not had chance to create any tests around this. I have a separate branch for hub modifications on my side so I will be going in this direction for my clients project.

Look forward to hearing your feedback.


adam-code-labx:
@alecritson @glennjacobs 
I will wait for the data tables and other PR's to get merged to develop then open another PR for the page refactor later.
Closing this one for now 👀

PR #475, opened 2022-08-11
`Blink` default queries to avoid multiple database calls

ryanmitchell:
This PR adds `spatie/laravel-blink` as a dependency to the core project and uses it to store any defaults (eg currency, channels) for the lifecycle of the request.

On our test homepage this has removed 80 duplicate queries - which as you can imagine is a lot :)

There are definitely other areas this could be applied to, such as the CartSession manager methods, but this is a good start for now.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         "livewire/livewire": "^2.0",
         "php": "^8.0",
         "spatie/laravel-activitylog": "^4.4",
-        "spatie/laravel-blink": "^1.6.*",
+        "spatie/laravel-blink": "^1.6",
         "spatie/laravel-medialibrary": "^9.0.0|^10.0.0",
         "yab/laravel-scout-mysql-driver": "^5.0"
     },
--- a/composer.json
+++ b/composer.json
@@ -19,6 +19,7 @@
         "livewire/livewire": "^2.0",
         "php": "^8.0",
         "spatie/laravel-activitylog": "^4.4",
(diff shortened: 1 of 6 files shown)


glennjacobs:
I think this is definitely needed. @alecritson are we happy this is safe enough to do when it comes to updates, etc?

alecritson:
Yeah I think it's a good idea to have.

@ryanmitchell I think you need to run `vendor/bin/monorepo-builder merge` at the root level so the tests get access to the new packages. Pretty sure that's why the tests are having a hard time.

ryanmitchell:
@alecritson thanks. I think the issue is that I need to register the service provider:
<https://github.com/spatie/laravel-blink#installation>

Any thoughts on how to do that?

alecritson:
You should be able to do that here:

https://github.com/getcandy/getcandy/blob/develop/packages/core/tests/TestCase.php#L25
https://github.com/getcandy/getcandy/blob/develop/packages/admin/tests/TestCase.php#L30

glennjacobs:
> That’ll work for the tests, but does it not need registered to be picked up in real world usage?

I don't think so, the package registers itself with Laravel automatically.

ryanmitchell:
@glennjacobs got you, it pulls it from composer.json - should have noticed that. 

That should be it now.

alecritson (commented):
I think this looks reasonable, what do you think @glennjacobs ?

alecritson (approved):

PR #476, opened 2022-08-12
Hotfix - Fix activity log form on variants

alecritson:
There is currently an issue on product variant editing which results in the comment form on the activity log not working as intended, this PR fixes that.

Also comments weren't showing for Product variants, this fixes that too.


glennjacobs:
@alecritson can you fix the conflicts? :-)

glennjacobs (approved):

PR #477, opened 2022-08-12
Enhancement - Fix variant image duplication

alecritson:
Currently when adding images to a variant it's quite restrictive and doesn't reuse existing media records very well.

At the moment, when you add an image to the variant (even from choosing an existing product image) it will duplicate it, this isn't great as some variants will want to use the same image i.e. a generic green t-shirt across all sizes, causing the amount of media that gets added to the database and ultimately in storage to grow quickly.

This PR aims to solve that with these notable changes:

- Added a new link table `media_product_variant` with a `position` pivot column
- When adding new images to a variant directly, behind the scenes we will save these images against the product and populate the link table.
- When choosing an image that exists on the product, instead of duplicating it we simply add a row to the link table.

This then allows variants to have multiple images whilst still retaining them on the product, for use by other variants and also keeps the amount of files and media records being generated to a minimum.

- Removed the `media` relation from the `ProductVariant` in favour of an `images` `BelongsToMany` relationship.

The method `getThumbnail()` remains intact so has backwards compatibility in that respect.


markmead (commented):
Had a play and it looks good to me!

Only thing I spotted, not even sure if it's an issue, but the `media_product_variant` table has empty values...

Screenshot 2022-09-02 at 09 44 35

glennjacobs (approved):

PR #493, opened 2022-08-22
fix error: str_contains(): Passing null to parameter #1

wychoong:
fix error `str_contains(): Passing null to parameter #1` in laravel.log

### DIFF
--- a/packages/admin/src/Listeners/SetStaffAuthMiddlewareListener.php
+++ b/packages/admin/src/Listeners/SetStaffAuthMiddlewareListener.php
@@ -15,11 +15,13 @@ class SetStaffAuthMiddlewareListener
     public function handle(RouteMatched $event)
     {
         // Is this a livewire route and are we looking for a hub component.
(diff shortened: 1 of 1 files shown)


alecritson (approved):

adam-code-labx:
@wychoong @alecritson 
There seems to be an issue with this last update.

From my side pulling the latest changes on develop branch I now get a 403 when trying to edit a product or collection.
Agreed you are initially going to get null for `$event->route->name` as the route name param is not defined until after the initial livewire requests.

The name is a magic method which will get the parameter name from the route.

```php
/**
 * Dynamically access route parameters.
 *
 * @param  string  $key
 * @return mixed
 */
public function __get($key)
{
    return $this->parameter($key);
}
```

`$event->route->getName()` returns livewire.message where as `$event->route->name` returns the route name parameter

```
+parameters: array:1 [▼
   "name" => "hub.components.products.show"
]
```

I would put a quick PR together to solve this issue.

PR #498, opened 2022-08-25
Fix for SetStaffAuthMiddlewareListener

adam-code-labx:
@wychoong @alecritson 
Refactored and updated the conditions reference to the latest fix causes 403 issue #493

### DIFF
--- a/packages/admin/src/Http/Middleware/Authenticate.php
+++ b/packages/admin/src/Http/Middleware/Authenticate.php
@@ -30,12 +30,8 @@ class Authenticate extends Middleware
      */
     protected function authenticate($request, array $guards)
     {
-        $check = $this->auth->guard('staff')->check();
-
-        if ($check) {
-            return $this->auth->shouldUse('staff');
-        }
-
-        $this->unauthenticated($request, $guards);
+        $this->auth->guard('staff')->check()
+            ? $this->auth->shouldUse('staff')
+            : $this->unauthenticated($request, $guards);
     }
 }
--- a/packages/admin/src/Listeners/SetStaffAuthMiddlewareListener.php
+++ b/packages/admin/src/Listeners/SetStaffAuthMiddlewareListener.php
@@ -23,7 +23,7 @@ class SetStaffAuthMiddlewareListener
     {
         $this->event = $event;
 
-        if (!$this->routeAlreadyHasStaffMiddleware() && $this->shouldApplyStaffMiddleware()) {
+        if (! $this->routeAlreadyHasStaffMiddleware() && $this->shouldApplyStaffMiddleware()) {
             $event->route->middleware(
                 array_merge($event->route->middleware(), ['auth:staff'])
             );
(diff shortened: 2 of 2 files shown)


alecritson:
Thanks @adam-code-labx This seems to come up every now and then 🤦  does this need to worry about the password reset / login routes? I can see the check was removed somewhat.

adam-code-labx:
> Thanks @adam-code-labx This seems to come up every now and then 🤦 does this need to worry about the password reset / login routes? I can see the check was removed somewhat.

Yeah in the condition previously you were saying always ignore password and login. 
So not required as we are defining what is passable now as an array for the middleware. Any reason you did not go with creating a middleware class instead of this listener?

alecritson:
> > Thanks @adam-code-labx This seems to come up every now and then 🤦 does this need to worry about the password reset / login routes? I can see the check was removed somewhat.
> 
> Yeah in the condition previously you were saying always ignore password and login. So not required as we are defining what is passable now as an array for the middleware. Any reason you did not go with creating a middleware class instead of this listener?

Cool, hopefully this will put an end to it then! Yeah the reasoning was Livewire has it's own set of middleware that we can't extend as a package without adding to the main Laravel app, which in turn alters the developers Laravel app and if they're using Livewire we don't want it to trigger if they're not in the hub :)

adam-code-labx:
Might be good to add a test for this before merging. I won’t have time until tonight and you will have a better idea with what checks to add to this test 😜

adam-code-labx:
@alecritson 
“ if they're using Livewire we don't want it to trigger if they're not in the hub :)”

In this case the condition would be wrong. As the middleware would trigger for all livewire. events. If you only want this in the hub we will need to do some additional  checks on the livewire message event. I have about an hour at midday so will work on this and get the tests done.

adam-code-labx:
@alecritson @glennjacobs @wychoong 
I have improved the middleware check. 
Tested in all cases so not sure we really need a unit test for this one as that may take a little time to put together with faking the livewire events etc. this should really be merged today as it currently causes 403 issues with the latest develop branch.

Here is what the middleware is now doing:
1. We avoid anything which already has the staff middleware applied
2. Check what routes we can apply middleware to then ⬇
  a. Specifically checking this is livewire event route
  b. Only ever apply to hub.pages & hub.components
  c: Make sure we disallow certain auth routes as these are also hub.components
  
So should be good to go now 👍

alecritson:
Thanks @adam-code-labx agree we need this in as it's currently not working. Looks sound enough.

alecritson (approved):

wychoong:
hey, first of all I’m sorry of the oversight that fixing this for my frontend and didn’t aware the `getName` and `name` is totally different, damn magic 😜

also, isn’t making the checking to `hub.pages` `hub.components` is way too narrow than `hub` and yet the hub routes none is pages/components, unless I’m reading it wrong

adam-code-labx:
@wychoong It's easy to miss those 🪄 methods especially with livewire.
The check is based on the livewire components which have been registered in AdminHubServiceProvider

The problem with just defining only hub and using contains, means this middleware could get applied outside the hub if someone was to create a livewire component with the name Hub in for example GitHubRepos could be livewire component of someone wishes to create this, so in that case the middleware would be applied. 
I feel it was better in this situation to be more explicit.

Also hub.pages & hub.components are the name parameter for the livewire components.
So I feel this covers all the components that we use in the hub right now.

PR #499, opened 2022-08-25
Laravel Pint (code style)

adam-code-labx:
StyleCI is ok for the GitHub actions however until you merge the changes this is not triggered.
I have also set this up from my side so when you push to branch it auto fixes the styles with an additional commit.

tbh this is a little annoying especially when you forget to pull changes and the fact it co auth PR's lol. 
I think most devs would prefer to run the simple command to check and fix code before they submit to the repo.

The reason behind this PR is to help us all maintain Laravel coding standards when contributing to GC.
I have added a set of rules to ignore as it flagged up way too many files especially with binary_operator_spaces.
With the Laravel preset and ignore rules there were only 7 style issues found in 643 files. I imagine StyleCI solved most of the issues.

I have also applied the pint fixes in the last commit.
@glennjacobs @alecritson if your good with this suggestion I will write the dev docs although we could just say pint is now included and more info can be found here https://laravel.com/docs/9.x/pint

Screenshot 2022-08-25 at 17 37 51


alecritson:
If we're going down this route, should we look to get an action set up to do this automatically? Otherwise we'd be expecting developers to do this manually?

adam-code-labx:
It’s almost like the tests I love to see all green before making a commit so we should mention in the docs to run manually but not make this a requirement.

We should also set this up as an action perhaps on merge https://heydarmen.com/blog/integrating-laravel-pint-with-github-actions.html

We could also adjust the rules to be a little more stricter then remove php language from the StyleCI as pint will take care of PHP

glennjacobs:
My initial concern with this was that StyleCI and Pint could be slightly different and constantly fight each other.

However, the idea of having Pint handle PHP and StyleCI other stuff (like HTML, JS, etc) is a reasonable suggestion.

glennjacobs:
@adam-code-labx if you could do the following, I think we could merge this in.

- Update the styleci config to not cover PHP
- Add a GitHub Action to run Pint and auto-commit any fixes
- Add to the contributing guide in the docs about using Pint

glennjacobs:
@alecritson actually, do we want it to auto-fix? 🤔

adam-code-labx:
> @alecritson actually, do we want it to auto-fix? 🤔

Auto fix on every commit I would say no, there is a --test flag so certainly can do the additional checks. 
On merge would be best for auto fixing. Once we remove php from styleci then we need to discuss the rules to apply to pint.

Depending on how strict we want to go with the code style?
You will see if you only had the laravel preset without those extra rules then it would flag from 653 files, 185 style issues
So really depends on how strict you want this to be 😉

adam-code-labx:
@glennjacobs Pint is all setup for PR checks. Looks like styleci is still running the php checks so I assume this will need to be merged over to develop before that change takes effect. Not quite sure how to reorder the Git Actions ideally pint should run just before the StyleCI tests. Also what would you like this to renamed to? Currently "Laravel Pint Code Style Tests / Test"

I would suggest setting up another action to auto fix when merging to develop.

glennjacobs:
Hi @adam-code-labx after a lengthy chat with @alecritson we've decided there's little benefit in keeping StyleCI, so we'll look to just have Pint.

glennjacobs:
I would just call it "Laravel Pint"?

adam-code-labx:
Sure sounds good 👍 

The default preset without the extra rules flags up 185 file with binary_operator_spaces issues.
Mainly the spacing between array key values, so really depends on if you would like to keep this for readability or go with standards to clean this up.

@glennjacobs For the next step perhaps you guys should discuss internally how strict the pint rules should be.
https://laravel.com/docs/9.x/pint#rules

PR #501, opened 2022-08-27
fix toggle UX

wychoong:
- set `:on` for `<toggle>` so the initial state is correct and the toggle will move left/right when clicked
&nbsp; or remove `class="{{ $on ? 'bg-green-500' : 'bg-gray-200' }} ` to let alpine handle the class
- bind `offValue` to `:class` to get better UX when clicked instead wait for livewire response to update toggle state


markmead:
Hi, @wychoong I can't replicate an issue with the toggle component, do you have an example of what this PR fixes?

packages/admin/resources/views/components/input/toggle.blade.php
```diff
@@ -25,6 +25,7 @@ class="{{ $on ? 'bg-green-500' : 'bg-gray-200' }} relative inline-flex flex-shri
     x-on:click="toggle"
     :class="{
         'bg-green-500': checked == onValue,
+        'bg-gray-200': checked == offValue,
```
markmead, line 28: Do we need this? We could just add `bg-gray-200` to the class list and that would be the default.

packages/admin/resources/views/components/input/toggle.blade.php
```diff
@@ -33,10 +34,11 @@ class="{{ $on ? 'bg-green-500' : 'bg-gray-200' }} relative inline-flex flex-shri
 
     <span
       aria-hidden="true"
+      class="{{ $on ? 'translate-x-5' : 'translate-x-0' }} pointer-events-none inline-block h-5 w-5 rounded-full bg-white shadow ring-0 transition ease-in-out duration-200"
       :class="{
         'translate-x-5': checked == onValue,
+        'translate-x-0': checked == offValue,
```
markmead, line 40: Same here, we could just add `translate-x-0` to the class list.

packages/admin/resources/views/components/input/toggle.blade.php
```diff
@@ -33,10 +34,11 @@ class="{{ $on ? 'bg-green-500' : 'bg-gray-200' }} relative inline-flex flex-shri
 
     <span
       aria-hidden="true"
+      class="{{ $on ? 'translate-x-5' : 'translate-x-0' }} pointer-events-none inline-block h-5 w-5 rounded-full bg-white shadow ring-0 transition ease-in-out duration-200"
```
markmead, line 37: I think we could/should remove the `{{ $on > '...' : '...' }}` from this component, it's not needed if we set a default class.

markmead (commented):
Looks good! Added some food for thought.

packages/admin/resources/views/components/input/toggle.blade.php
markmead: Worth noting there could be issues with CSS importance, if there is we could do something like this:

```php
@class([
  'pointer-events-none inline-block h-5 w-5 rounded-full bg-white shadow ring-0 transition ease-in-out duration-200',
  'translate-x-5' => $on,
  'translate-x-0' => !$on,
])
```

wychoong: > I think we could/should remove the `{{ $on > '...' : '...' }}` from this component, it's not needed if we set a default class.

Initially I wanted to remove this `$on`** but its being used in multiple places, so I didn't go with that route, and submitted this quick fix as product page was the one I was working on at that time.

If we are set with letting alpine to handle the toggle state, which IMO is cleaner and also better. These are the changes I can think of now:
- toggles will show as `off` on load and set to the right state after alpine is ready
- immediate state/color change when toggled instead of livewire response, better UX

\** need to take extra care when changing this as toggles can support custom value instead of just booleans

markmead: In that case, it's probably worth addressing the feedback as that will align with mine and your thoughts on using Alpine JS for this.

glennjacobs:
@markmead is this one good to go now?

packages/admin/resources/views/components/input/toggle.blade.php
wychoong: hey @markmead help needed. removing `$on` seems working fine. however I'm facing issue with product image set primary toggle. when set one image as primary, the other images should set to false, from the html i can see the `x-data checked` is updated with the correct value but the classes are not.

this happens when `model` is not passed and only product images only

wychoong: https://user-images.githubusercontent.com/67364036/188496348-697585dd-5540-4b90-b8a8-d301a02d09a7.mov

markmead:
Hey, @wychoong I made that change, not sure if you want to pull it down and have a play.

@glennjacobs It looks good to me.

packages/admin/resources/views/components/input/toggle.blade.php
markmead: Looks like an issue that already exists, @alecritson do you have any guidance on this one? I think a Livewire component would solve the issue, but do we want to add one for this?

alecritson: I saw this was an issue but should have been resolved in #497 Let me know if this still isn't the case if this PR branch has been updated to the latest develop...

markmead: Would that fix have fixed the issue here? The fix was for URLs. I assume the same logic could be applied.

alecritson: The screen recording specifically shows the URL manager and I remember hitting the issue shown when working on that PR, that's why I think it's related. If it's happening elsewhere then potentially it's something else.

alecritson: Ignore that, I thought this was showing the URL manager 🤦 need a coffee

alecritson: However, it's possible that the image manager isn't unsetting the primary value when checked, so I'm not sure it's related to this PR?

markmead: OK in that case is it worth getting this PR merged and creating an issue/PR for what's been spoken about?

wychoong: passing `wire:model` to primary toggle solved that. perhaps its not something fixable
```js
x-data="{
    checked: false, // <- when `wire:model` is not passed the value is `true/false` instead of binding to state,
}"
```
on backed `image.*.primary` is updated, and `checked` is updated with correct value in the html, however the js is not triggered to toggle the classes

markmead (approved):

wychoong:
> Looks like there's some conflicts that need addressing before this can go in.

should be resolved now

wychoong:
> @wychoong apologies, this has been hanging around for a while.
> 
> Mark is no longer in the Lunar team, so Alec and I are trying to pick up on these UI PRs.
> 
> Could you remind us what problem this solves?

Toggle is not showing the correct state

alecritson:
> > @wychoong apologies, this has been hanging around for a while.
> > Mark is no longer in the Lunar team, so Alec and I are trying to pick up on these UI PRs.
> > Could you remind us what problem this solves?
> 
> Toggle is not showing the correct state

Is there a page you've specifically seen this happen?

wychoong:
> > > @wychoong apologies, this has been hanging around for a while.
> > > Mark is no longer in the Lunar team, so Alec and I are trying to pick up on these UI PRs.
> > > Could you remind us what problem this solves?
> > 
> > 
> > Toggle is not showing the correct state
> 
> Is there a page you've specifically seen this happen?

Product page url and channel section. Most of the page with toggle likely have the issue

alecritson:
@wychoong This has been sitting around for a while, is this something that is still needed, can it be revived or should this be closed?

wychoong:
> @wychoong This has been sitting around for a while, is this something that is still needed, can it be revived or should this be closed?

I would say its still needed. but i think best is to relook into it again, so will close it for now and open new PR later

PR #506, opened 2022-09-02
fix/log-ui

markmead:
Small PR to fix up some issues I saw in the activity log UI.

I noticed the variant form activity log `required` was causing issues, might be fixed with #476


markmead:
Nope, as you've addressed it in #476

alecritson:
> Nope, as you've addressed it in #476 I can fix it in this one if you want tho.

Cool, yeah since it's on another PR, need point duplicating

glennjacobs (approved):

PR #507, opened 2022-09-02
update/version-in-ui

markmead:
Shows the current GetCandy version in the UI.

If it's a semver version it defaults to `2.0-beta`, do you think this should be set in the config or nah?


packages/admin/src/Views/Components/GetCandy/Version.php
```diff
@@ -0,0 +1,37 @@
...
+use Illuminate\Support\Str;
+use Illuminate\View\Component;
+
+class Version extends Component
+{
+    public $installedVersion;
+
+    public function __construct()
+    {
+        $installedVersion = InstalledVersions::getPrettyVersion('getcandy/getcandy');
```
alecritson, line 15: 🤯

alecritson:
Just pulled down the branch and the menu toggle is now appearing at the top 
image

markmead:
Odd, have you published the CSS?
Screenshot 2022-09-05 at 10 13 39

alecritson:
@markmead Just republished the assets and still the same, does the css need to be recompiled on the branch?

markmead:
Yeah, that's what I did @alecritson

PR #508, opened 2022-09-03
Fix/misc

wychoong:
- fix hub.js typo
- fix button css classes
&nbsp;&nbsp;when include own css file, the classes importance causes issue when `$theme == 'gray'` because of `border-transparent border-gray-300` appeared at the same time
- fix activity log feed styling
&nbsp;&nbsp;- move the 'Add Comment' button position slightly and added border and shadow to make it standout better
**Before**
image
**After**
image
&nbsp;&nbsp;- fixed comment input field left margin
**Before**
image
**After**
image


adam-code-labx:
@wychoong couple more suggestions to add to your misc fixes

If will be easier to target custom styles if we have an identifier I suggest adding data-section to the body.
```html
<body
    data-section="{{ collect(request()->segments())->implode('-') }}"
    class="h-full overflow-hidden antialiased bg-gray-50 dark:bg-gray-900"
    :class="{ 'dark': darkMode }"
    x-data="{
      showExpandedMenu: $persist(false),
      showMobileMenu: false,
      darkMode: $persist(false),
    }">
```

Also the hub sections are currently using boxed width max-w-screen-2xl. 
This should be optional imo. full-width would be better by default.

```html
<div class="px-4 py-8 mx-auto {{ config('getcandy-hub.system.layout_width', 'max-w-screen-2xl') }} sm:px-6 lg:px-8">
     @yield('main', $slot)
</div>
```

Then add this config to getcandy-hub/system.php
```php
<?php

return [
    'layout_width' => 'w-full',
    'path' => 'hub',
];
```

alecritson:
Thanks @wychoong we'll try have a look at this asap, one thing I saw was the change to the activity log form on variants, there is a known issue with this and is fixed in PR #476 so I'm not sure there should be changes in here for it.

wychoong:
> Thanks @wychoong we'll try have a look at this asap, one thing I saw was the change to the activity log form on variants, there is a known issue with this and is fixed in PR #476 so I'm not sure there should be changes in here for it.

My changes for the mentioned items are in separate files, i think it would be better if you can copy the changes to the mentioned PR and close this. As my changes are pure TW classes, no functional changes

markmead:
Hey @wychoong, the activity log feed changes aren't required, it's the HTML that's broken and has been fixed on #476, there's also a complementary PR #506 which cleans up the UI.

wychoong:
> Hey @wychoong, the activity log feed changes aren't required, it's the HTML that's broken and has been fixed on #476, there's also a complementary PR #506 which cleans up the UI.

sure, I will remove variant page related changes, but I don't think the `Add Comment` button have any changes in both commits? you can refer the screenshots above to get what I meant

glennjacobs (approved):

PR #509, opened 2022-09-05
Add ability to remove a modifier

ryanmitchell:
This is a bit niche but we add a modifier in our SP and I want to not run that modifier for an integration with a specific 3rd party service, so adding the ability to remove a modifier would be really helpful.

### DIFF
--- a/packages/core/src/Base/CartLineModifiers.php
+++ b/packages/core/src/Base/CartLineModifiers.php
@@ -22,7 +22,7 @@ class CartLineModifiers
     {
         $this->modifiers->push($modifier);
     }
-    
+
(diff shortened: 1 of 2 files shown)


glennjacobs:
@ryanmitchell we're using Laravel Pint now. You should be able to run it to fix up the styling issues.

ryanmitchell:
Done. Maybe auto-run the cleanup in the action instead?

glennjacobs:
> Done. Maybe auto-run the cleanup in the action instead?

It was discussed. Might still do it, see how things go.

glennjacobs (approved):

PR #512, opened 2022-09-06
Show digital product lines as well as physical ones

ryanmitchell:
Not sure if this is handled by changes in other PRs, but I noticed that digital products were not being shown in the order view.

### DIFF
--- a/packages/admin/resources/views/livewire/components/orders/show.blade.php
+++ b/packages/admin/resources/views/livewire/components/orders/show.blade.php
@@ -48,7 +48,7 @@
                     </ul>
                 </div>
 
-                @if ($this->physicalLines->count() > $maxLines)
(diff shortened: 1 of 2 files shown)


ryanmitchell:
If you mark an item as not requiring shipping, then it gets marked as digital when you purchase and then doesn't show up. This change, or something like it is necessary. What specifically do you think needs documented?

glennjacobs:
> If you mark an item as not requiring shipping, then it gets marked as digital when you purchase and then doesn't show up. This change, or something like it is necessary. What specifically do you think needs documented?

ha, you know more than me :-) Just checked and you're correct.

If that's the case, then we probably don't need anything added to the docs.

ryanmitchell:
Real world usage :)

glennjacobs (approved):

PR #513, opened 2022-09-07
dark-mode/customers

markmead:
Initial pass at dark mode for the customer screens.

I'm leaving buttons, tables and pagination as these are components that will be tackled on a different PR.


packages/admin/resources/views/livewire/components/customers/index.blade.php
```diff
@@ -1,21 +1,15 @@
-<div class="flex-col space-y-4">
-    <div class="flex items-center justify-between">
-        <strong class="text-xl font-bold md:text-2xl">
-            {{ __('adminhub::catalogue.customers.index.title') }}
-        </strong>
-    </div>
+<div class="space-y-4">
+    <h1 class="text-xl font-bold text-gray-900 md:text-2xl dark:text-white">
```
wychoong, line 2: can we only merge dark mode classes only if dark mode is enabled?

```php
@class([
       'text-xl font-bold text-gray-900 md:text-2xl',
       'dark:text-white' => config('getcandy-hub.system.dark_mode'),
])
```

markmead: At the moment dark mode is based on system preferences but there will be a toggle to enable/disable.

wychoong: i meant disable dark mode entirely.
because if you included the `dark:', the css comes with GC will have those classes and works fine, but for those additional slot/pages we added, it will stay white unless we add dark mode support.

markmead: Yeah, I think an option to disable it as a developer makes sense! Great suggestion! 🙌

PR #505, opened 2022-09-01
[0.2] Image editor

wychoong:
Integrate Filerobot Image Editor [(FIE)](https://github.com/scaleflex/filerobot-image-editor) to product image editor

- using cdn instead of npm as its giving a lot warning message during npm install
- should consider allow config file to configure editor tools or whole config?
- Some css styling for readability and more fitting to GC color theme, still need polishing but I’ll leave it up to core team

PR #515, opened 2022-09-07
Fix - Issue where uploading variant media overrides product media.

alecritson:
Fixes an issue brought up on #505 by @wychoong

### DIFF
--- a/packages/admin/src/Http/Livewire/Components/Products/Variants/VariantShow.php
+++ b/packages/admin/src/Http/Livewire/Components/Products/Variants/VariantShow.php
@@ -240,7 +240,7 @@ class VariantShow extends Component
                 if ($newImage) {
                     $media->setCustomProperty('caption', $image['caption']);
                     $media->setCustomProperty('primary', false);
-                    $media->setCustomProperty('position', $image['position']);
+                    $media->setCustomProperty('position', $owner->media()->count() + 1);
                     $media->save();
                 }
 
--- a/packages/admin/src/Http/Livewire/Components/Products/Variants/VariantShow.php
+++ b/packages/admin/src/Http/Livewire/Components/Products/Variants/VariantShow.php
@@ -212,6 +212,8 @@ class VariantShow extends Component
             $imagesToSync = [];
 
             foreach ($this->images as $key => $image) {
+                $newImage = false;
+
                 if (empty($image['id'])) {
                     $file = TemporaryUploadedFile::createFromLivewire(
                         $image['filename']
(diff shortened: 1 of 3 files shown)


packages/admin/src/Http/Livewire/Components/Products/Variants/VariantShow.php
```diff
@@ -229,14 +231,18 @@ public function updateImages()
...
                 $media = app(config('media-library.media_model'))::find($image['id']);
 
-                $media->setCustomProperty('caption', $image['caption']);
-                $media->setCustomProperty('primary', false);
-                $media->setCustomProperty('position', $image['position']);
-                $media->save();
+                if ($newImage) {
+                    $media->setCustomProperty('caption', $image['caption']);
+                    $media->setCustomProperty('primary', false);
+                    $media->setCustomProperty('position', $image['position']);
```
wychoong, line 243: I think only can set property for caption, position will clash with main product too. Maybe store in the media_product_variant table?

alecritson: Good point, the problem is we are storing the actual media against the product so it will need a position, I've done a small tweak to use the number of media models against a product + 1 so it should give a proper position 🤔

wychoong: What if the reorder image position in variant? Probably count() + position?

However it still somehow feels wrong to me. Seems will still be affecting when image is linked to more than 1 variant 🤔

alecritson: @wychoong I've adjusted it and added position on the pivot for variant images, so they are independent from the product image position. We then only apply the position to the product media if it's a new one since it's going on the product, what do you think?

wychoong: Sounds good to me. It’s the changes is not committed right? Just in case I missing anything 😅

alecritson: Good point, I didn't push it up on friday 🤦 I've pushed it up now though 🤞

wychoong: LGTM! Just to remove line 242 and 243 😃

alecritson: Will ping @glennjacobs for the final review!

glennjacobs (approved):

Issue #519, opened 2022-09-09
Tax_Breakdown of Order multiplication by number of items


AphgaDesign:
- GetCandy version: 2.0@beta
- Laravel Version: 9.2
- PHP Version: 8.1
- Database Driver & Version: MySQL, 8.0.29

### Expected Behaviour: 
The Order's TaxBreakdown percentage at 5% for TPS (exemple)

### Actual Behaviour: 
![GetCandy-Bug-Tax_Breakdown](https://user-images.githubusercontent.com/17551643/189261535-fb64df87-e655-4110-a342-a595572d70fd.JPG)
When you create an Order from Cart, the Order's TaxBreakdown percentage are multiplicated by number of items in order (TPS of 5% are at 15% for an Order with 2 items and 1 shipping).

### Steps To Reproduce: 
1. Configure TPS at 5%
2. Configure TVQ at 9.975%
3. Add 2 items with 1 shipping option in cart
4. Create Order from Cart
5. Look tax_breakdown on Order's table in database.

glennjacobs:
Can you please show your tax configuration? Depending on how you've set it up, this would be the correct behaviour.

We don't have enough information here to reproduce.

AphgaDesign:
Ok, i show you the configuration in hub parameters and an exemple in a product
![GetCandy-Bug-Tax_Breakdown-1](https://user-images.githubusercontent.com/17551643/189345431-ca80b168-ff07-4e54-90f1-38bb73c47bd6.JPG)
![GetCandy-Bug-Tax_Breakdown-2](https://user-images.githubusercontent.com/17551643/189345433-914ecea2-293f-4333-9020-a39065cae4f3.JPG)

glennjacobs:
You have configured two tax rates, so it's doing what is expected.

How do you need it to work?

AphgaDesign:
In Quebec, we have TPS (5 %) and TVQ(9,975 %). In DB, the tax_breakdown for a product have right % for TPS and TVQ. But the tax_breakdown for an order have the wrong % because it was multiply by the number of lines in order

AphgaDesign:
And an other information, in cart, the % is working fine but not in order

glennjacobs:
Right I see, so the 5% for example should always be 5%, a percentage shouldn't increase (e.g. to 15%).

Apart from the percentages, are the other figures correct?

glennjacobs:
FYI, I have been able to reproduce this.

[...]

PR #522, opened 2022-09-09
Change sum to min for getting tax percentage

glennjacobs:
Fix for #519 

We shouldn't be summing the percentage on order tax summaries, so changed to `min()` as the percentages should always be the same regardless.


glennjacobs:
@alecritson tests failed on this, but nothing to do with this code. Seems like there may be an issue with the existing tests and setting language code. Guessing it randomly sets a code and every once in a while a conflict occurs.

alecritson:
I think there is a failure that happens once in a blue moon, will look to see what's going on when I can.

Does this PR still need tests?

glennjacobs:
> Does this PR still need tests?

It's why I have it in draft. I looked at making a test, but there was nothing quick I could borrow to make one.

PR #532, opened 2022-09-15
Add migrate command

alecritson:
This is a single use command to upgrade your GetCandy database to the Lunar database schema.

```sh
php artisan lunar:migrate:getcandy
```

## Prerequisites 

- Renamed `config/getcandy/database` to `config/lunar/database`
- Updated the table prefix to `lunar_`

> If you have used your own prefix, then you do not need to do this.

## Steps

- If you've used the `getcandy_` prefix, it will add the `lunar_` tables and copy the data across. It won't delete the original `getcandy_` tables, so you'll need to do this manually at the end.
- Update any Polymorphic tables Lunar knows about replacing `GetCandy` with `Lunar` in the namespace. If you have custom polymorphic tables you'll need to do this yourself.
- Updates any `attribute_data` columns that Lunar knows about, again, if you have custom tables for this it's a manual task.

---

The intention is for this command to be available during one minor release cycle i.e. `0.1.0` and removed in `0.2.0`


packages/core/src/Console/Commands/MigrateGetCandy.php
```diff
@@ -0,0 +1,181 @@
...
+
+    /**
+     * Execute the console command.
+     *
+     * @return void
+     */
+    public function handle()
+    {
+        $tables = collect(
+            DB::select('SHOW TABLES')
```
glennjacobs, line 35: This isn't PostgreSQL compatible.

glennjacobs: Maybe try this instead?

```php
$tables = DB::connection()->getDoctrineSchemaManager()->listTableNames();
```

glennjacobs (approved):

PR #536, opened 2022-09-16
Bug/allow null indexing in scout db

glennjacobs:
It was identified that the Scout DB engine wasn't able to handle null values. This PR introduces a test and fix for this.

### DIFF
--- a/utils/scout-database-engine/src/DatabaseEngine.php
+++ b/utils/scout-database-engine/src/DatabaseEngine.php
@@ -55,7 +55,7 @@ class DatabaseEngine extends Engine
                     'key' => $model->getScoutKey(),
                     'index' => $model->searchableAs(),
                     'field' => $field,
-                    'content' => $data,
+                    'content' => $data ?? '',
                 ];
             });
 
--- a/utils/scout-database-engine/tests/database/migrations/2022_08_31_000000_create_posts_table.php
+++ b/utils/scout-database-engine/tests/database/migrations/2022_08_31_000000_create_posts_table.php
@@ -15,8 +15,8 @@ return new class extends Migration
     {
         Schema::create('posts', function (Blueprint $table) {
             $table->id();
-            $table->string('title');
-            $table->text('body');
+            $table->string('title')->nullable();
+            $table->text('body')->nullable();
             $table->timestamps();
         });
     }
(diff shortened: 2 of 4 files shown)


utils/scout-database-engine/tests/database/migrations/2022_08_31_000000_create_posts_table.php
```diff
@@ -15,8 +15,8 @@ public function up()
     {
         Schema::create('posts', function (Blueprint $table) {
             $table->id();
-            $table->string('title');
-            $table->text('body');
+            $table->string('title')->nullable();
```
poppabear8883, line 18: Why does these need to be nullable ?

glennjacobs: So that we can test null values, otherwise we can't save null values and therefore run a test to check them.

poppabear8883: Sure but doesn't the fix set empty string if null otherwise it sets the content of `$data` ? so effectivelt `content` will never be `null`

glennjacobs: I'm not following, sorry. But the file you're commenting on is a test, it's not fixing anything.

utils/scout-database-engine/src/DatabaseEngine.php
```diff
@@ -55,7 +55,7 @@ public function update($models)
                     'key' => $model->getScoutKey(),
                     'index' => $model->searchableAs(),
                     'field' => $field,
-                    'content' => $data,
+                    'content' => $data ?? '',
```
alecritson, line 58: Is there a reason we would want to allow nothing to be added? Why wouldn't we just filter out empty values in the first place?

glennjacobs: I thought about that and it was primarily for debugging and reporting. If there's simply no entry, it would be a lot harder to determine if something wasn't indexing properly and also avoid confusion with those less knowledgeable on how it works. 

Is that worth making redundant database entries? I'm not sure.

PR #546, opened 2022-09-21
Hotfix - Tweak nested collection styling

alecritson:
Tweaks the padding on nested collections so the right is aligned correctly.


markmead:
Already been done on #544 and fixes the issue where the `pb` grows.

PR #562, opened 2022-09-23
Add ability to remove a shipping modifier

ryanmitchell:
Similar to the change previously merged to allow cart line and cart modifiers to be removed, this allows shipping modifiers to be removed.

Again the same use case - we have an external service creating orders so we dont want processing to be run on those.


ryanmitchell:
The test failures dont seem related to the code change at all... is something else going on?

alecritson:
There was a breaking change to a minor version on L8 🤦 Fixed on L8 but the same change doesn't exist on L9 (yet) so figuring that out...

glennjacobs (approved):

PR #568, opened 2022-09-23
Hotfix - Fix up state upgrading

alecritson:
There are some issues currently when running state up dates that have resulting in some schema updates being added to the state management classes that run after migrations have happened.

Schema updates should only happen in migrations otherwise they can be missed and as a result won't run. An example of this was for the brand update, on a fresh install the `brand` column on products wasn't being removed.

The PR looks to make some tweaks to the way the state is managed. 

### Additional `prepare` method required on state classes

This method will be triggered before any outstanding migrations are run, this will give developers the chance to grab any data and store it against the class before the `run` method is triggered.

For example:

```php
<?php

namespace Lunar\Database\State;

use Lunar\Models\Product;

class DoSomethingToTheDatabaseData
{
	/**
	 * The legacy brands to import.
	 *
	 * @var array
	 */
	protected $legacyBrands = [];

	public function prepare()
	{
		$this->data = Product::get();
	}

	public function run()
	{
		foreach ($this->data as $row) {
			// Do something...
		}
	}
}
```

### DIFF
--- a/packages/core/database/state/EnsureBrandsAreUpgraded.php
+++ b/packages/core/database/state/EnsureBrandsAreUpgraded.php
@@ -8,30 +8,53 @@ use Lunar\Models\Product;
 
 class EnsureBrandsAreUpgraded
 {
+    /**
+     * The legacy brands to import.
+     *
+     * @var array
+     */
+    protected $legacyBrands = [];
+
+    public function prepare()
+    {
+        $prefix = config('lunar.database.table_prefix');
+
+        $hasBrandsTable = Schema::hasTable("{$prefix}brands");
+        $hasProductsTable = Schema::hasTable("{$prefix}products");
+
+        if ($hasBrandsTable || ! $hasProductsTable) {
+            return;
+        }
+
+        $legacyBrands = Product::query()->pluck('brand', 'id')->filter();
+
+        if ($legacyBrands->isEmpty()) {
+            return;
+        }
+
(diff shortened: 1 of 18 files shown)


packages/core/database/state/EnsureBrandsAreUpgraded.php
```diff
@@ -2,37 +2,73 @@
...
+        if (! $this->canRun() || ! $this->shouldRun()) {
+            return;
+        }
+
+        $brands = null;
+
+        try {
+            $brands = Storage::get('tmp/state/legacy_brands.json');
+        } catch (FileNotFoundException $e) {
+            // $brands = null;
```
glennjacobs, line 54: Comment

packages/core/database/state/EnsureBrandsAreUpgraded.php
```diff
@@ -2,37 +2,73 @@
...
+                    'name' => $brandName,
+                ]);
+
+                Product::whereIn('id', $productIds)->update([
+                    'brand_id' => $brand->id,
+                ]);
+            }
         }
+
+        Storage::disk('local')->delete('legacy_brands.json');
```
glennjacobs, line 71: Will this error if there's no file?

alecritson: Pretty sure it just falls through. No reason to error if the file doesn't exist

glennjacobs (approved):

PR #579, opened 2022-09-27
Hotfix - Add OG image to docs

markmead:
This needs some spice.
Screenshot 2022-09-27 at 11 49 16
Adds the OG image from the main website.


glennjacobs:
Image file should be in "docs/src/.vuepress/public/images"

markmead:
I'm not sure why that matters, if anything the approach taken is preferred as we want the OG image to be consistent. Ideally, we'd use the CDN but that's not setup, right?

glennjacobs:
I mention it because that's where VuePress says to put it. 🤷‍♂️

And for consistency, we don't have any other assets directly in the src folder.

glennjacobs:
Oh, I see you've referenced the website's logo. Sorry, doing this on the mobile app. 

Unless on a CDN I would keep them separated for the moment. Just because we could update the website and break the docs.

markmead:
No worries and understood! All updated 🚀

glennjacobs (approved):

PR #583, opened 2022-09-27
Hotfix - Docs spelling

markmead:
Some minor fixes.

### DIFF
--- a/docs/src/lunar/images.md
+++ b/docs/src/lunar/images.md
@@ -59,7 +59,7 @@ class StorefrontConversions
 {
     public function apply($model)
     {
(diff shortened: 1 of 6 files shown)


docs/src/lunar/images.md
```diff
@@ -59,7 +59,7 @@ class StorefrontConversions
 {
     public function apply($model)
     {
-        // .. Register spatie media conversions here...
```
markmead, line 62: Seemed a bit weird having "spatie" here, it should be "Spatie" in my opinion, but that looked weirder.

alecritson (approved):

PR #593, opened 2022-09-29
Hotfix - Consistent publish names

markmead:
I noticed the publish names are using different separators (`.`, `-`, `:`) which looks messy and causes the order to be off.

**Before**
Screenshot 2022-09-29 at 08 47 37

**After**
Screenshot 2022-09-29 at 08 49 23


alecritson:
Can we add a changelog entry for this :) Likely we'd also need a bit on the upgrade guide in case anyone is manually doing this.

markmead:
Added some docs.

alecritson (approved):

glennjacobs (approved):

PR #602, opened 2022-09-30
Fixing docs

glennjacobs:
Brings the images docs back into the lunar folder and corrects the currencies docs.


docs/src/lunar/images.md
```diff
@@ -24,17 +24,17 @@ If you've used the medialibrary package before you will feel right at home.
 ```php
 $product = \Lunar\Models\Product::find(123);
 
-$product->addMedia($request->file('image'))->toMediaCollection('images');
```
alecritson, line 27: Is there a reason we're undoing these docs? The images reference was the latest iteration I think?

glennjacobs: Oh, it's actually the wrong way round. So we just need to delete the getcandy entry then by the looks of things.

alecritson:
@glennjacobs Could this go straight on to main branch?

alecritson (approved):

PR #613, opened 2022-10-05
Fix passing options to Quill editor

ryanmitchell:
This previously wasnt working as the json string was being html encoded, but the quotes also need replaced to be alpine-friendly.

### DIFF
--- a/packages/admin/resources/views/components/input/richtext.blade.php
+++ b/packages/admin/resources/views/components/input/richtext.blade.php
@@ -3,7 +3,7 @@
     x-data="{
         value: @entangle($attributes->wire('model')),
         init() {
-          {{ $instanceId }} = new Quill($refs.editor, {{ json_encode($options) }})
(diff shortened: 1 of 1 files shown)


packages/admin/resources/views/components/input/richtext.blade.php
```diff
@@ -3,7 +3,7 @@
     x-data="{
         value: @entangle($attributes->wire('model')),
         init() {
-          {{ $instanceId }} = new Quill($refs.editor, {{ json_encode($options) }})
+          {{ $instanceId }} = new Quill($refs.editor, {!! str_replace('"', "'", json_encode($options)) !!})
```
glennjacobs, line 6: I think str_replace is risky. As we're passing data into an HTML attribute, it should be HTML encoded I would have assumed.

glennjacobs:
@ryanmitchell why do you think HTML encoding is wrong? That's the standard approach for HTML attributes.

Can you post an example of what doesn't work? All seems fine our end.

ryanmitchell:
Yeah you're right, ignore me, the issue was coming from elsewhere.

PR #614, opened 2022-10-05
Feat - Enable positioning of columns on tables

alecritson:
It would be quite useful to allow developers to specify where additional columns should appear on tables. Currently adding columns just puts them at the end, this could be problematic if developers want to add important information to a table, it may not be visible without scrolling.

This PR adds an `after` method to columns which reference the field they should render after.


alecritson:
> @alecritson if we're implementing `after`, should we also support `before`?

I did think about this but I thought it would be useful for Lunar to retain at least some control in what the first column is.

glennjacobs (approved):

PR #630, opened 2022-10-08
Feature/extend Lunar Migrate Command

adam-code-labx:

- Documentation updates
- Automated tests
- Changelog entries (core and hub)

@alecritson Your current command works very well when there are no modifications made to getcandy tables.
However when developers have customisations across multiple tables the command fails when trying to add the additional columns and data over to lunar tables.

This PR solves this issue by allowing the developer to prepare their upgrade migrations which will then be processed once the lunar tables have been created but before copying over the data.

I think we should remove the old getcandy details once we have confirmed each table has the same number of records.
By default I have set cleanup to true, however you can opt out by setting `--cleanup=false`

Docs have been updated to reflect all these changes.

### DIFF
--- a/docs/src/upgrading.md
+++ b/docs/src/upgrading.md
@@ -139,6 +139,11 @@ We have created a command for this purpose to try make the switch as easy as pos
 php artisan lunar:migrate:getcandy
 ```
 
+#### Instructions for Custom Application (Optional)
+We realise that developers may have added their own fields to a number of the `getcandy_` tables.
+In order for your upgrade to include the custom fields. Please move any migration files which previously extended GetCandy over to `database/migrations/upgrade`
+Next step you will need to change the migration namespace from `GetCandy\Base\Migration` to `Lunar\Base\Migration`
+
 #### What this command will do
 
 - Remove any previous GetCandy migrations from the `migrations` table.
@@ -146,6 +151,9 @@ php artisan lunar:migrate:getcandy
 - Copy across the data from the old `getcandy_` tables into the new `lunar_` tables.
 - Update any polymorphic `GetCandy` classes to the `Lunar` namespace.
 - Update field types in `attribute_data` to the `Lunar` namespace.
+- Remove previous GetCandy migrations located in `database/migrations/upgrade` from the migrations table.
+- Run the upgrade migrations again with the lunar_ prefix, updating the new tables.
+- By default, this command will remove the `getcandy_` tables once checking the data has been transferred.
 
 
 #### What this command will not do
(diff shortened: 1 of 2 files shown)


packages/core/src/Console/Commands/MigrateGetCandy.php
```diff
@@ -15,7 +15,8 @@ class MigrateGetCandy extends Command
      *
      * @var string
      */
-    protected $signature = 'lunar:migrate:getcandy';
+    protected $signature = 'lunar:migrate:getcandy
+        {--cleanup=true : Removes the getcandy tables after successful migration}';
```
alecritson, line 19: Part of me thinks cleanup should default to `false`, should be an opt in behaviour since it involves data removal.

docs/src/upgrading.md
```diff
@@ -139,13 +139,21 @@ We have created a command for this purpose to try make the switch as easy as pos
 php artisan lunar:migrate:getcandy
 ```
 
+#### Instructions for Custom Application (Optional)
+We realise that developers may have added their own fields to a number of the `getcandy_` tables.
+In order for your upgrade to include the custom fields. Please move any migration files which previously extended GetCandy over to `database/migrations/upgrade`
```
alecritson, line 144: I'm not sure this is the best approach, wouldn't it be better to have an array you can add migrations that need to be removed so they can be re-run?

packages/core/src/Console/Commands/MigrateGetCandy.php
adam-code-labx: Well perhaps just add a warning in the docs to backup 1st 😁
Personally true is better which currently checks the count match in each table we could go further to check the data integrity?

docs/src/upgrading.md
adam-code-labx: Yeah @alecritson I see your point and perhaps need to extend the docs further in this case.
In terms of an array. This seems like a lot of extra work as for example I have 15 migrations in the upgrade table and would need to copy each filename into an array.

The reason I have done this on my side was to upgrade one time so scanning the directory made sense.
However as migrations are run in order of the timestamp file names. It might be better to also have the hub migrations to run from `packages/core/database/migrations/upgrade` directory which will contain a copy of only the create_ migrations and combining the add_ update_ this was all the timestamps will be 2021 in the upgrade directory and run before the customisation migrations in the app.

It will then be worth mentioning in the docs to add the app upgrade path to AppServiceProvider only after Lunar has been successfully upgraded so that `migrate:fresh` will continue to run?

glennjacobs:
@alecritson @adam-code-labx is this still needed?

glennjacobs:
This isn't something we feel we now need.

PR #631, opened 2022-10-09
Allow to override Lunar table component namespace

adam-code-labx:
Now I am sure this PR will most likely get rejected 👀
However I feel as the aim for Lunar is to be extendable as possible, we should have this option for developers to switch out the table components namespace via the system config. Any table components which have not been overridden will simply fallback to the default Lunar implementation.

With my own use-case (filament tables package)
Before the Lunar update we had already implemented filament tables across the hub and did a lot of work to provide custom filters and actions for our client. Also recently taken advantage of the new table grid listing feature and soon to work on the new batch editing.

Now although I can appreciate the amount of work involved especially from your @alecritson creating the new tables components from scratch. Personally I will be continuing with Filament tables when using the hub as I feel Lunar's tables will struggle to keep up with the same features especially with the v3 coming up soon. 

Of course what I am suggesting here can simply be achieved without this PR if a developer was to override the AdminHubServiceProvider. The idea is to allow the developer to choose how they want to implement the tables features. 
My suggestion is not only based on filament, it would work with any Livewire table library implementation.

### DIFF
--- a/packages/admin/config/system.php
+++ b/packages/admin/config/system.php
@@ -1,5 +1,8 @@
 <?php
 
 return [
+    'components_namespace' => [
+        'tables' => 'Lunar\\Hub\\Http\\Livewire\\Components',
+    ],
     'path' => 'hub',
 ];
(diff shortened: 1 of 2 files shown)


glennjacobs:
After our team meeting, we have decided that this isn't the direction we want to take. If developers want a different screen for something, we want to encourage them to create their own custom add-on page and link to that instead. 

To achieve this we will be updating the hub menu builder so that existing menu items can be removed, e.g. if you wanted to provide your own Orders screen.

PR #639, opened 2022-10-14
Add caching and additional discount information to Cart

glennjacobs:
Introduces caching of cart properties. This means we don't have to recalculate the cart totals when accessing via a different cart instance.

- [ ] Documentation updates
- [ ] Automated tests


packages/core/src/Models/Cart.php
```diff
@@ -89,6 +144,18 @@ protected static function newFactory(): CartFactory
         'meta' => 'object',
     ];
 
+    /**
+     * The "booted" method of the model.
+     *
+     * @return void
+     */
+    protected static function booted()
```
alecritson, line 152: Can this not go on the trait? So you don't have to do it per model

```
public static function bootedCachesProperties();
```

glennjacobs: I think so. And it's also a reminder I've not even put it on the other models. 

I'll get that done and then start to make some tests for it.

alecritson:
@glennjacobs Is this finished with? Be good to get it merged into `main` if ready.

alecritson:
Closing as #676 has these changes in.

Issue #644, opened 2022-10-21
Table bulk action component is hard coded


maurice-ellis:
- Lunar version: #.#.#
- Laravel Version: 9.35.1
- PHP Version: 8.0.24
- Database Driver & Version: Maria

### Expected Behaviour:
When adding a bulk action to the table builder, I'm expecting it to show the component I choose

### Actual Behaviour:
When adding a bulk action, no matter what action, it opens the update status.

In `vendor/lunarphp/lunar/utils/livewire-tables/resources/views/index.blade.php`, livewire, the component, is never passed as a parameter

```
<div :hidden="!selected.length" class="py-4">
      <p class="lt-text-sm lt-font-medium lt-text-gray-900">
          Bulk Actions
      </p>

      <div class="lt-flex lt-flex-wrap lt-gap-4 lt-mt-2">
          @foreach ($this->bulkActions as $action)
              @livewire($action->getName(), [
                  'label' => $action->label,
              ])
          @endforeach
      </div>
  </div>
```

In `vendor/lunarphp/lunar/utils/livewire-tables/resources/views/actions/bulk.blade.php`, 
```
<div x-data="{ showModal: false }">
    <button type="button"
            wire:click="$set('showModal', true)"
            class="lt-px-3 lt-py-2 lt-text-sm lt-font-medium lt-text-gray-600 lt-transition lt-bg-white lt-border lt-border-gray-200 lt-rounded-md hover:lt-bg-gray-50 hover:lt-text-gray-700 hover:lt-shadow-sm focus:lt-outline-none focus:lt-ring focus:lt-ring-blue-100 focus:lt-border-blue-300">
        <span class="lt-capitalize">
            {{ $label }}
        </span>
    </button>

    <x-l-tables::support.modal wire:model="showModal">
        <div>
            @livewire('hub.components.tables.actions.update-status', [
                'ids' => $selectedIds,
            ])
        </div>
    </x-l-tables::support.modal>
</div>
```

The livewire property is not set, just defaults to `hub.components.tables.actions.update-status`

### Steps To Reproduce:

```
$this->tableBuilder->addBulkAction(
    BulkAction::make('update_availability')
        ->label(__("hub/tables.actions.availability"))
        ->livewire('hub.components.printing.tables.actions.update-availability')
);
```

maurice-ellis:
Fixed through a pull request #646

alecritson:
This should be resolved in `0.1.2`

PR #646, opened 2022-10-21
Hotfix - Unable to use table bulk action component

maurice-ellis:
Documents mentioned that we can choose a component for the table bulk action, however, that does not work because the component is hard-coded to use the update status action. This request fixes that and properly allows for the component mentioned when adding the bulk action to a table to be used.


maurice-ellis:
Fixed #644

alecritson:
Thanks for sorting this @maurice-ellis I had a quick look and it looks good 👍  Just need @glennjacobs to have a look then we can get this merged in

glennjacobs (approved):

PR #648, opened 2022-10-21
Improve query result from product table

Aslam97:
Would you @glennjacobs considering improving queries for admin-hubs in the future? this is just example of how i minimize the total queries executed into 4 queries from 70 queries.

> note this is a test with 13 products which produce 70 queries

before

Screenshot from 2022-11-07 14-02-25

after

Screenshot from 2022-11-07 14-02-46


alecritson:
Thanks @Aslam97 

My only concern with eager loading in the variants is if a product has a lot of different variations it's potentially going to be quite slow to eager load them all into memory, just to get a stock figure and SKU. This is the reason why it was done this way initially. Interested to hear @glennjacobs thoughts.

Aslam97:
@alecritson the only thing I can think of is to use a standard native query that can be executable on mysql or postgresql. i know most people prefer not to use native query. this is just an example but this only handle `stock` and `sku` not `variant->thumbnail`. 

```php
$query = Product::query()
            ->selectRaw("lunar_products.*,
            (
                CASE
                    WHEN (
                        SELECT
                            COUNT(*)
                        FROM
                            lunar_product_variants
                        WHERE
                            lunar_product_variants.product_id = lunar_products.id
                    ) > 1 THEN 'Multiple'
                    ELSE (
                        SELECT
                            sku
                        FROM
                            lunar_product_variants
                        WHERE
                            lunar_product_variants.product_id = lunar_products.id
                        LIMIT
                            1
                    )
                END
            ) AS sku")
            ->orderBy($this->sortField, $this->sortDir)
            ->withTrashed()
            ->withSum('variants', 'stock')
            ->with(['thumbnail', 'productType', 'brand']);
```

glennjacobs:
@Aslam97 @alecritson I think a raw select in this instance is probably correct. It's the most efficient way to get the data out of the database.

We just need to ensure it works in both MySQL and PostgreSQL and that we handle the database table prefixes.

I guess we could also use this approach for the stock summary?

Aslam97:
@glennjacobs I will update this PR. let's see if I can handle thumbnails too with a raw query.

Aslam97:
@glennjacobs I have update the pr. but to im not able to get the media from the first variant from each product and the code is a bit messy.

glennjacobs:
- [ ] Need to check this definitely works for both MySQL and PostgreSQL
- [ ] Check that devs can still add to the select OK, e.g. `addSelect()`

PR #652, opened 2022-10-27
Fix minutes in orders table date formats

ryanmitchell:
Orders table date format is wrong - it was showing month instead of minutes


markmead (approved):
Oops, that explains why all the orders were at `XX:10` 😅 

Working all fine now.

ryanmitchell:
Yep! I was looking at our order list wondering if it was something to do with a queue...

PR #671, opened 2022-10-31
Report back to user if email is invalid

### LINKED ISSUE #660: Display error message if user reset password with an email that is not registered in the system.
**Is your feature request related to a problem? Please describe.**
If the user resets the password and then enters the wrong email when clicking the "send reset email" button, there is no notification telling the user that the email is not registered in the system. its doing nothing.

**Describe the solution you'd like**
option 1. add unique email validation for staff something like `exists:lunar_staff,email`
option 2. use custom validation error message something like this

```php
    public function sendResetEmail()
    {
        $this->validate();

        $staff = Staff::whereEmail($this->email)->first();

        if (is_null($staff)) {
            $this->addError('email', 'The email address you entered is not registered with us.');
        }

        // ...
    }
```

Aslam97:
fixed #660

### DIFF
--- a/packages/admin/src/Http/Livewire/Components/Authentication/PasswordReset.php
+++ b/packages/admin/src/Http/Livewire/Components/Authentication/PasswordReset.php
@@ -56,7 +56,7 @@ class PasswordReset extends Component
      * {@inheritDoc}
      */
     protected $rules = [
-        'email' => 'required|email',
+        'email' => 'required|email|exists:'.Staff::class.',email',
         'password' => 'nullable|confirmed',
         'password_confirmation' => 'nullable',
     ];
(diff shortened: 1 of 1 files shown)


markmead:
Hi @Aslam97 I don't think we should be telling users if an email exists in the system or not. Something we could do is show the notification regardless of the email's existence.

glennjacobs:
We definitely should NOT be indicating if an email exists or not, that would be a security issue.

Aslam97:
@markmead @glennjacobs What are the best words in your opinion to tell that the email you entered is wrong? at least there is a notification to the user. so I can update the PR

markmead:
@Aslam97 I would just remove this https://github.com/lunarphp/lunar/blob/develop/packages/admin/src/Http/Livewire/Components/Authentication/PasswordReset.php#L107-L109 outside of the `if` statement.

Aslam97:
@markmead i've update the PR

markmead (approved):
All looks! Thanks for updating @Aslam97

PR #680, opened 2022-11-03
[0.2] Feature - Nested menu

markmead:
Adds nested menus through sections with links to those sections.


alecritson:
@markmead Looks like this might need some documentation updates https://docs.lunarphp.io/extending/admin-hub.html#adding-to-menus

docs/src/extending/admin-hub.md
```diff
@@ -20,9 +22,10 @@ php artisan vendor:publish --tag=lunar.hub.views
 
 Lunar uses dynamic menus in the UI which you can extend to add further links.
 
-::: tip
-Currently, only the side menu and settings menu are available to extend. But we will be adding further menus into the core editing screens soon.
-:::
+::: tip Currently, only the side menu and settings menu are available to extend.
```
alecritson, line 25: Looks like the formatting has gone wrong a bit here.

docs/src/extending/admin-hub.md
```diff
@@ -32,15 +37,51 @@ use Lunar\Hub\Facades\Menu;
...
+$products->addItem(function ($item) {
+    $item
+        ->name(__('menu.sidebar.tickets.expiring'))
+        ->handle('hub.tickets.expiring')
+        ->route('hub.tickets.expiring')
+        ->icon('times');
+});
+```
+
+Lunar comes with a collection of icons you can use in the Resources folder. If
```
alecritson, line 83: Feel like this should have it's own heading

docs/src/extending/admin-hub.md
```diff
@@ -32,15 +37,51 @@ use Lunar\Hub\Facades\Menu;
...
+
+$products->addItem(function ($item) {
+    $item
+        ->name(__('menu.sidebar.tickets.expiring'))
+        ->handle('hub.tickets.expiring')
+        ->route('hub.tickets.expiring')
+        ->icon('times');
+});
+```
+
```
alecritson, line 82: Can we get a list of available sections dev's can tap into.

markmead: Literally a list like

```
By default Lunar has two sections in the menu:

- Products
- Orders
```

alecritson: What I mean is, how would someone add a menu item under products or orders?

packages/admin/src/Menu/SidebarMenu.php
```diff
@@ -47,60 +47,46 @@ protected function addSections()
...
 
-        // $catalogueManager = $slot->section('catalogue-manager')->name(
-        //     __('adminhub::menu.sidebar.catalogue-manager')
-        // );
-
-        $slot->addItem(function ($item) {
-            $item->name(
-                __('adminhub::menu.sidebar.products')
-            )->handle('hub.products')
+        $products = $slot->section('hub.products')
```
alecritson, line 50: Currently the `MenuSlot` class tries to find the section using `Str::slug` which, given `hub.products` won't match.

I guess we should either change the handles to support this or require the handle to be exact.

markmead: Should probably just check the `$handle`, right? That change works.

alecritson: Yeah I think being explicit with the handle should be fine.

glennjacobs:
- [ ] Add facility to remove menu items programatically

glennjacobs:
image

Not sure why we have renamed "Orders" to "Purchases"? It should be "Orders".

markmead:
It's the update functionality kicking in as I was testing it @glennjacobs

glennjacobs:
@markmead this menu appears quite different to me from the original concept. There is no collapsed menu which shows just the icons and the mobile menu doesn't have all the entries. I've got the latest version of the branch and have published the assets, am I missing something?

glennjacobs:
@markmead OK, I think I know where I was going wrong. It no longer auto-collapses, you have to press the arrow button near the bottom.

Would be nice to keep the animations from the original demo.

glennjacobs:
Tweaked the menu to work as below

image

glennjacobs:
Having discussed with @alecritson we feel the below would be the ultimate.

So "sections" are just that, but we introduce child menu items.

image

markmead:
@glennjacobs It was never auto-collapsing on the mockup I sent, as can be seen here - https://ui-menu.vercel.app/

glennjacobs:
> @glennjacobs It was never auto-collapsing on the mockup I sent, as can be seen here - https://ui-menu.vercel.app/

Yeah, I realise that now. But I would like to retain that functionality if possible.

markmead:
Just to confirm, what is this "auto-collapsing" functionality that needs to be retained?

glennjacobs:
It was when your screen isn't very wide, it defaults to the collapsed menu, rather than the expanded out one. Otherwise, the forms etc will be very squished, unless we go to a mobile burger menu style?

markmead:
That's not something the menu has ever done, it's always switched to a burger menu on smaller screens.

glennjacobs:
> That's not something the menu has ever done, it's always switched to a burger menu on smaller screens.

No worries. Let's get it coded so we can review how it feels to use. The latest design has had very positive feedback on Twitter.

glennjacobs:
Can we get this closer to the design?

image
Left, original design
Right, what I'm seeing at the moment

markmead:
The original design hasn't taken into account the space needed for the hover effect on child items, is that something we want to keep, or should we simplify the hover effect to just change the text color?

It also has accessibility issues on the child items with the font size, color, and weight combination. Hence the increase in size and color change.

glennjacobs:
> The original design hasn't taken into account the space needed for the hover effect on child items, is that something we want to keep, or should we simplify the hover effect to just change the text color?
> 
> It also has accessibility issues on the child items with the font size, color, and weight combination. Hence the increase in size and color change.

I think just change text colour is fine on the child items.

The font size is exactly the same, the colour may need to come up, but I'd like to keep the differentiation between the child and parent items.

markmead:
> The font size is exactly the same, the colour may need to come up, but I'd like to keep the differentiation between the child and parent items.

The font size isn't the issue, as mentioned it's a combination of size, color, and weight. I'll have a play and see what works.

---

Sorry about editing your comment, I thought I clicked "Quote Reply" 😅

PR #685, opened 2022-11-04
Patch 2

Aslam97:


alecritson:
> @glennjacobs @alecritson I am about to remove `$this->currency->format` and `$symbol` in price component. since `format` is no longer have a value. what do you think?

I think that makes sense since I believe it's just blank. Happy to see what you come up with.

alecritson:
@Aslam97 We're going to be moving away from the `develop` branch going forward. When you get a chance can you target `main` (I don't want to potentially mess anything up, but it should be fine)

Aslam97:
@alecritson ok I'll update towmorrow

markmead:
I've looked through this and it seems all OK but there is quite a lot to take in 😅

There are some small tweaks I'd make to the `views/partials/pricing.blade.php` but that is more personal preference than anything and is not something that should stop this from being approved from my perspective.

packages/admin/resources/views/partials/pricing.blade.php
```diff
@@ -136,54 +135,70 @@
...
+                            required
+                        >
+                            <x-hub::input.text 
+                                id="tier_field_{{ $index }}"
+                                wire:model='tieredPrices.{{ $index }}.tier' 
+                                type="number"
+                                min="2" 
+                                steps="1" 
+                                required
+                                onkeydown="return event.keyCode !== 190"
```
markmead, line 175: Can we use Alpine JS for this?

Aslam97: updated

markmead (approved):
Looks good to me! Be good to get this merged! 🚢

PR #692, opened 2022-11-07
[0.1] Hotfix - Fix price saving

### LINKED ISSUE #662: Pricing changes when saved in hub
- Lunar version: 0.1.0-rc.5
- Laravel Version: 8.83.25
- PHP Version: 8.1.8

### Expected Behaviour:

When I enter a price of 18.99 into the price fields in the hub, they get saved as 18.98...

### Actual Behaviour:

The price I enter is saved.

### Steps To Reproduce:

I've narrowed the issue down to[ this line](https://github.com/lunarphp/lunar/blob/9194dd7f8c53c4e8e5a10f3fb409498430524bae/packages/admin/src/Actions/Pricing/UpdatePrices.php#L58)

The price starts at 18.99, by the currency factor becomes 1899.0 and then by the int casting becomes 1898... so very strange. I've tried changing it to use intval instead of casting as an int, but it does the same.

Maybe you just need to treat it as a string and strip what comes after the decimal away?

alecritson:
Closes #662

### DIFF
--- a/packages/admin/src/Actions/Pricing/UpdatePrices.php
+++ b/packages/admin/src/Actions/Pricing/UpdatePrices.php
@@ -54,8 +54,8 @@ class UpdatePrices
         $currency = Currency::find($currencyId);
 
         $priceModel->fill([
-            'price' => (int) ($price * $currency->factor),
-            'compare_price' => $comparePrice ? (int) ($comparePrice * $currency->factor) : null,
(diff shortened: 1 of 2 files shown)


ryanmitchell:
Amazing, thank you.

markmead (approved):
Tested and works well.

PR #696, opened 2022-11-08
[0.1] Hotfix - Handle missing status from table column

alecritson:
Currently, if the status is not passed to the Status column component it will throw an error if it can't find it in the config.

Closes #697


markmead:
What's the issue this fixes? I can't replicate.

alecritson:
> What's the issue this fixes? I can't replicate.

If you have an order with a status that isn't in `config/lunar/orders` an error gets thrown

markmead (approved):

PR #697, opened 2022-11-10
Add paid status to orders config

kylekanderson:
The demo-store package sets the status for orders to 'paid' once payment has been captured. When this happens, the Orders page in the hub breaks with a vanilla lunar install, as the 'paid' status does not exist in the default 'statuses' config entry in the orders.php config file. Adding this default entry here will avoid this from happening for new users trying out the demo-store.

### DIFF
--- a/packages/core/config/orders.php
+++ b/packages/core/config/orders.php
@@ -42,5 +42,11 @@ return [
             'mailers' => [],
             'notifications' => [],
         ],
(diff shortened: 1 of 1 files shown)


alecritson:
Thanks for reporting. There is a PR which aims to handle missing statuses, I think if anything the demo store should be updated to reflect the config shipping with Lunar.

PR #701, opened 2022-11-10
Hotfix - Properly handle adding media when disk is not local

maurice-ellis:
In order to obtain the media from a remote location,addMediaFromDisk has to be used.addMedia only works when the file is stored on the server. You can set a default media disk for Spatie Media Library, so I used that config and then determined if it was local or not, so the image can be found and added properly.

### DIFF
--- a/packages/admin/src/Http/Livewire/Traits/HasImages.php
+++ b/packages/admin/src/Http/Livewire/Traits/HasImages.php
@@ -218,10 +218,10 @@ trait HasImages
                         $image['filename']
                     );
                     
-                    $media_libary_disk = config("media-library.disk_name");
+                    $media_libary_disk = config('media-library.disk_name');
                     $media_libary_driver = Storage::disk($media_libary_disk)->getConfig()['driver'];
 
-                    if($media_libary_driver == "local") {
+                    if ($media_libary_driver == 'local') {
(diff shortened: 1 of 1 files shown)


maurice-ellis:
@alecritson I ran pint and not sure what the issue is and why it's failing 😔

alecritson:
Strange one! Will take a look if needed. One thing I did notice was the variable names, could you please update to camel case 🙂

maurice-ellis:
> Strange one! Will take a look if needed. One thing I did notice was the variable names, could you please update to camel case 🙂

Done

alecritson:
Thanks @maurice-ellis I had a little look at this and on Laravel 8 I got this error:
image

Also, instead of trying to figure out the disk being used, can we not just check for the file existence instead? Something like:

```php
if (File::exists($file->getRealPath())) {
    $media = $owner->addMedia($file->getRealPath())
        ->toMediaCollection('images');
} else {
    $media = $owner->addMediaFromDisk($file->getRealPath())
        ->toMediaCollection('images');
}
```

maurice-ellis:
@alecritson It seems this issue happens when the file is not local and then requires `addMediaFromDisk`.  I believe the condition would be true because even on s3 let's say, the file would exist.

glennjacobs (approved):

PR #715, opened 2022-11-15
[0.1] Hotfix - Hide error icon on interactive

markmead:
This PR fixes the error icon overlapping the `type="number"` input spinners.

Another approach would be to remove the spinners but that doesn't feel right in an admin system.

### DIFF
--- a/packages/admin/resources/views/components/input/group.blade.php
+++ b/packages/admin/resources/views/components/input/group.blade.php
@@ -3,7 +3,9 @@
            class="flex items-center text-sm font-medium text-gray-700">
         <span class="block">{{ $labelPrefix ?? null }}</span>
 
-        <span class="block">{{ $label }}>
+        <span class="block">
+            {{ $label }}
+
             @if ($required)
                 <sup class="text-xs text-red-600">&#42;</sup>
             @endif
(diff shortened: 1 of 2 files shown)


markmead:
I would argue do we even need the icon to be placed inside the input... Could it not be inline with the error text? 🤔

packages/admin/resources/views/components/input/group.blade.php
```diff
@@ -1,36 +1,46 @@
...
-          @foreach($error as $text)
-            <p class="text-sm text-red-600">{{ $text }}</p>
-          @endforeach
-        @else
-          <p class="text-sm text-red-600">{{ $error }}</p>
+    <label for="{{ $for }}"
+           class="flex items-center text-sm font-medium text-gray-700">
+        <span class="block">{{ $labelPrefix ?? null }}</span>
+
+        <span class="block">{{ $label }}>
```
alecritson, line 6: There's an rogue `>` here.

alecritson:
> I would argue do we even need the icon to be placed inside the input... Could it not be inline with the error text? 🤔

I think this is a good shout, currently the icon gets in the way I think more than it benefits the UI.

alecritson (approved):

PR #716, opened 2022-11-15
[0.1] Hotfix - Consistent image size

markmead:
This PR makes all thumbnail images use the same `w-16 h-16 rounded` classes to stop inconsistency in the UI.

And an extra plus, make some of the images visible without the need for a high-powered microscope.

### Changes

- https://github.com/lunarphp/lunar/pull/716/files#diff-9e006142ad9d36742166c89225ea67f5d0cbdd3704d5ea808348d5bd8b7633e6L171-L177
- https://github.com/lunarphp/lunar/pull/716/files#diff-9ccc191d8553898dcebacd51320c8ff5b469f816f2b91fb91ca357fb647fabfaL124-L128
- https://github.com/lunarphp/lunar/pull/716/files#diff-90868f9ca7b448f05044c92333f76eb4663c374d51b507f8d40d41ae47ce9238R103-R108
- https://github.com/lunarphp/lunar/pull/716/files#diff-ddfd1d7e3ec6ace82adf37d004a48e6df0a52b57aa448db5ccc1c624acabfff3R12-R18
- https://github.com/lunarphp/lunar/pull/716/files#diff-ba958b1919ec55f726cb902737d2c6a5bd1c6b65a9a37b2784f42d6e8494a52aL5-L12


alecritson:
If we're doing this, would it be better to just make a blade image component?

```html
<x-hub::image src="..." />
```

alecritson:
@markmead Looks like there is a bunch of prettier changes in this as well, makes it hard to see just the image changes. Can we either do a separate PR with prettier changes but then sharing the set up in the repo so all dev's can benefit, otherwise it feels like this will keep coming up.

alecritson (approved):

Issue #728, opened 2022-11-17
Hub UI Tweaks


glennjacobs:
## Drag icon on images

<img width="313" alt="image" src="https://user-images.githubusercontent.com/647407/202430787-e0d802d3-b505-4748-b433-e50945328f4a.png">

Could we change this to a drag icon, so it doesn't look like a menu button?

---

## Clear search

<img width="1318" alt="image" src="https://user-images.githubusercontent.com/647407/202430982-24cdcd51-828a-4181-a404-4ffe6a7ffea7.png">

Could we have a cross button to clear the search keywords?

---

## Translate button half hidden

/hub/settings/attributes/product

<img width="276" alt="image" src="https://user-images.githubusercontent.com/647407/202431221-64e9f145-1acd-42b7-9dfa-aaba81a5b310.png">

---

## Make "Select all" more button-like

<img width="666" alt="image" src="https://user-images.githubusercontent.com/647407/202433039-478c7c5c-37d2-4881-87d8-23453b1b447c.png">

Can we make it feel more clickable? I looks like a header at the moment.

---

### Can't click on recent orders

Order number seems like it's linked, but doesn't click through it appears. This is on the dashboard.

<img width="181" alt="image" src="https://user-images.githubusercontent.com/647407/202435544-36350564-7d18-4e93-ba30-bc4e32651fa3.png">

Aslam97:
can this issue handle large nominal amount as well? maybe by reducing the font size or shortening the nominal only if it exceeds 1 million. for example: 27.210.001.891,05 becomes 27.210B. then if the user hovers on it the details will appear. 

![Screenshot from 2022-11-21 09-39-36](https://user-images.githubusercontent.com/25027592/202952938-c1a7ea31-88fd-4f7d-aba5-fd455da1f365.png)

AlessandroAmos:
Shouldn't dashboard cards have shadow like the customer page?

![image](https://user-images.githubusercontent.com/118480471/203072297-1f3a97a0-b164-408f-b7c9-7ee8761846ff.png)
![image](https://user-images.githubusercontent.com/118480471/203072901-1ec45d3c-da19-4bc0-908c-62039366d7eb.png)

markmead:
> can this issue handle large nominal amount as well? maybe by reducing the font size or shortening the nominal only if it exceeds 1 million. for example: 27.210.001.891,05 becomes 27.210B. then if the user hovers on it the details will appear.
> 
[...]

PR #732, opened 2022-11-17
[0.1] Hotfix - UI tweaks

markmead:
Fixes for #728.

I've changed how the locale button looks as:

1. It is unclear what it is
1. It doesn't work in modals
1. It doesn't work in slide-overs


markmead:
These are all done @glennjacobs. I had to make some tweaks to the locale toggle so it's worth taking a look.

glennjacobs (approved):

PR #736, opened 2022-11-18
[0.3] Feat - Sticky table cells

### LINKED ISSUE #731: Sticky columns in datatables
The action menu buttons are all the way on the right of the datatables and they often become hidden due to the table content. This isn't very helpful.

image

Ideally, the content should scroll horizontally, but the checkboxes on the left and the action menus on the right should be "sticky".

markmead:
Adds stickiness to the checkboxes and action cells in the table.


glennjacobs:
The Action menu doesn't seem to drop-down in the correct position.

image

It's also very difficult to see the button. We might want to introduce some colour here or something.

---

Action buttons are sticky, as are the checkboxes, but the header isn't on the right.
e.g. can still see the "Date" header.

image

alecritson:
@markmead I noticed if you have a few menu items, one you hit the hidden overflow you can't select the other options 
image

markmead:
Good spot. I've made a small tweak to swap `top-0` to `bottom-0` once the row is past the halfway point of the loop.

I tried a more complex solution that takes in the position of the menu when it opens in relation to the `tbody` but that wasn't perfect as you had to wait for the menu to appear before setting classes.

markmead:
What do you think of these changes? @glennjacobs It would be great to get them in the next release.

glennjacobs:
I can barely see the menu items above the datatable, it's far too subtle.

image

Could we try a drop shadow? Then if that doesn't work, I'll have to get into Figma.

markmead:
Shadow and a darker border have been added as the shadow wasn't enough.

PR #737, opened 2022-11-18
[0.1] Hotfix - Product status filter

markmead:
I've hardcoded the statuses array as it seems that it seems only `published` and `draft` is available.


glennjacobs:
@markmead there's also "deleted" which technically isn't a status, but it shows as one on the datatable.

markmead:
Do we want `deleted` as part of the dropdown or as a separate checkbox?

glennjacobs:
> Do we want `deleted` as part of the dropdown or as a separate checkbox?

I would suggest separate.

alecritson (approved):

PR #745, opened 2022-11-21
Use static::class when checking model search engine mapping

### LINKED ISSUE #744: It's not possible to reference some of Lunar extended models (like ProductVariant) in search engine mapping
- Lunar version: 0.1.4
- Laravel Version: 9.x
- PHP Version: 8.x
- Database Driver & Version:

### Expected Behaviour:

Any Lunar model should be open to be defined as searchable.

### Actual Behaviour:

Currently it's not possible to modify some of extended Lunar models to be searchable. (ProductVariant for example)

And even if it's possible, it requires a lot of hacking.

### Steps To Reproduce:

For example, suppose we want to define an extended `ProductVariant` as a searchable model.

1. First we extend the base Lunar model and add `Searchable` trait:

```php
namespace App\Models;

use Lunar\Base\Traits\Searchable;
use Lunar\Models\ProductVariant as BaseProductVariant;

class ProductVariant extends BaseProductVariant
{
    use Searchable;
}
```

2. Then, we register it in service provider ([as stated in the documentation](https://docs.lunarphp.io/extending/models.html#registration)):

```php
public function boot()
{
    $models = collect([
        \Lunar\Models\ProductVariant::class => \App\Models\ProductVariant::class,
    ]);
    ModelManifest::register($models);
}
```

3. Now we want to register our model in Lunar `search.php` config and define it's engine mapping.

But the problem is right here.

**If we use our new class ` \App\Models\ProductVariant` in `engine_map`:**

```php
'engine_map' => [
    \App\Models\ProductVariant::class => 'meilisearch',
],
```

Lunar `Searchable` trait [would ignore our engine mapping](https://github.com/lunarphp/lunar/blob/a243d8f1e35b6cee395ff62c7d39dd390ab675e6/packages/core/src/Base/Traits/Searchable.php#L154-L158).

**If we use the Lunar base model ` \Lunar\Models\ProductVariant` in `engine_map`:**

```php
'engine_map' => [
    \Lunar\Models\ProductVariant::class => 'meilisearch',
],
```

then we can't use search-specific events `searchSetup` and `indexing` in our model observer. simply because the Lunar base model has not `Searchable` trait.

### Possible Solution:

Maybe we just need to modify the Searchable trait to use `static::class` instead of `self::class` in [these lines](https://github.com/lunarphp/lunar/blob/a243d8f1e35b6cee395ff62c7d39dd390ab675e6/packages/core/src/Base/Traits/Searchable.php#L154-L158)?

armezit:
Fixes #744

### DIFF
--- a/packages/core/src/Base/Traits/Searchable.php
+++ b/packages/core/src/Base/Traits/Searchable.php
@@ -151,9 +151,9 @@ trait Searchable
     {
         $engines = config('lunar.search.engine_map', []);
 
-        if (isset($engines[self::class])) {
+        if (isset($engines[static::class])) {
             return app(EngineManager::class)->engine(
-                $engines[self::class]
+                $engines[static::class]
             );
         }
 


armezit:
The proposed changes are breaking. so i closed the pull request by now. needs more investigating.

PR #749, opened 2022-11-24
fix: saving product url with core model instead of extended model

wychoong:
refer to https://discord.com/channels/497326121441034240/712995078884294656/1034588048131891262


wychoong:
there are other places used `get_class` but I'm not sure should it be updated as well.
maybe @adam-code-labx can advice https://github.com/lunarphp/lunar/search?p=4&q=get_class%28&type=code

armezit:
This PR fixes #767

alecritson:
Thanks @wychoong do we also need to consider other area's where we use polymorphic relations 🤔

armezit:
@wychoong @alecritson The `UpdatePrices` action must be changed too:
https://github.com/lunarphp/lunar/blob/a27ce4b061b949773fa374dbeedcf0d18f9a4fcd/packages/admin/src/Actions/Pricing/UpdatePrices.php#L63

wychoong:
> Thanks @wychoong do we also need to consider other area's where we use polymorphic relations 🤔

ideally yes, but I unable to test all

wychoong:
replace with #807

PR #769, opened 2022-12-09
[0.2] Feat - Ability to mark orders as new customer

alecritson:
It would be handy to be able to know if an order that comes in is for a new customer, i.e. it's the first time they've ordered. This is good for reporting amongst other things. The issue is the query could be quite taxing on large datasets.

This PR adds a `new_customer` column to the orders table, which, when an order is created from a Cart will run a job to determine whether it's a new customer and update the column.


glennjacobs:
- Should we have a routine to update old orders?
- Should the reports now utilise this new field?
- Update docs to inform devs?
- Show on the order edit screen?

glennjacobs (approved):

PR #771, opened 2022-12-16
Feat - Automatic Price Conversion

armezit:
##### Also discussed in https://github.com/lunarphp/lunar/discussions/763

In addition to the price conversion, there were other requirements for implementing the feature.
Most important part was support for background jobs, both in core and admin hub.

The PR contains 3 main parts:
- Background jobs support in lunar
- Progress bar column for livewire tables
- Automatic price conversion

### Background jobs

- Background jobs are implemented in lunar core based on laravel job batches.
- I opted to use job batches (instead of single jobs) because of the features for execution flow of jobs and also support for grouping/naming them.
- The `job_batches` table is a slightly extended version of the laravel's official table, but is compatible with it.
- Also added a `JobBatch` model for working with job batches data in the admin hub.
- A basic job batches listing screen added into the hub setting (below the activity logs)
Screenshot from 2022-12-16 10-11-40

### Automatic price conversion

- The Price and Currency observers are responsible for dispatching price conversion job batches, when:
  - a non-default currency exchange rate changes
  - default currency price of a purchasable changes
- The jobs execute in background (as a job batch)
- a new config file `pricing.php` has been added (main parts are fully configurable).

#### possible improvements

There is still some places which can be improved:
For example, some events could be triggered in certain points of conversion process.
Another improvement could be in product/variant edit page (price section), to inform staff about the last auto-update operation.

Other things that are still missing:
- Docs
- Tests


alecritson:
Thanks for this @armezit , certainly looks interesting. Having an initial play around I got this error:

```
Call to undefined method Illuminate\Database\Eloquent\Casts\Attribute::make()
```

Steps to reproduce:

1) Created a new currency
2) Updated the exchange rate (see row in job batches)
3) Go to the background jobs page.

armezit:
> ```
> Call to undefined method Illuminate\Database\Eloquent\Casts\Attribute::make()
> ```

@alecritson Fixed it now.

The make() method was introduced in laravel 9.2. Replaced it with older syntax to support laravel 8.x too.
Also added a unit test for JobBatch model.

alecritson:
Looks like there are still a couple of issues, possibly with L8

When updating prices on a product: https://flareapp.io/share/Vme4Z2z5#F98

When viewing the background jobs page: https://flareapp.io/share/B5ZwXrgm#F106

armezit:
Yes I see. Seems with L8 we need some refactors.
I'll check the overall process on L8 and write more tests

armezit:
Added more unit tests, which covers most of the code.
Also fixed reported issues (incompatibilities with Laravel 8).

glennjacobs:
We still need to review this properly, but priorities have prevented it. Due to this and file conflicts, I have converted to Draft for the moment.

glennjacobs:
We are looking to take a different approach for this, which allows dynamically converted prices. If we decide to incorporate this functionality, it will be in 1.x now, so I will close this PR.

PR #786, opened 2022-12-26
Check language code unique when updating

rubenvanerk:
When updating a language and changing the code to an existing one, a database [exception is thrown](https://flareapp.io/share/J7oq6z3m#F64).

This PR adds a rule to check for the uniqueness of the code.

### DIFF
--- a/packages/admin/src/Http/Livewire/Components/Settings/Languages/LanguageShow.php
+++ b/packages/admin/src/Http/Livewire/Components/Settings/Languages/LanguageShow.php
@@ -2,6 +2,7 @@
 
 namespace Lunar\Hub\Http\Livewire\Components\Settings\Languages;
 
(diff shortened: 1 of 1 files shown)


alecritson:
Thanks for this @rubenvanerk I think this is correct however in the hub we tend to use a string when defining rulesets, In would it be possible to do this here to keep things consistent. Thanks.

rubenvanerk:
I did notice that, but decided against using a string in favor of readability. I will update this though! :+1:

alecritson:
@rubenvanerk I agree it's more readable and I think at some point it would be good to update rulesets to this for sure.

alecritson:
Thanks @rubenvanerk looks good!

alecritson (approved):

PR #787, opened 2022-12-26
Slightly improve UX for image manager

rubenvanerk:
- Show a tooltip that a primary image can't be deleted
- Reduce opacity of delete button for primary image

Before:
image

After:
image


alecritson:
Thanks @rubenvanerk since we're close to releasing `0.2` and this is more of an enhancement imo I feel like this should just target `main` to be included in that release. What do you think @glennjacobs ?

glennjacobs:
> Thanks @rubenvanerk since we're close to releasing `0.2` and this is more of an enhancement imo I feel like this should just target `main` to be included in that release. What do you think @glennjacobs ?

Sorry, only just seen this. Yes, I'd agree. Target `main`

rubenvanerk:
See https://github.com/lunarphp/lunar/pull/806

PR #789, opened 2022-12-30
fix images key

wychoong:
issues:

- after uploaded images, saving on second time cause validation issue due to `$key` is overwritten by `Str::random()` and failed to properly unset `$this->imageUploadQueue`
- PR #592 seems to suggest to use `Str::random` as array key and blade already updated to use, however some part of the code is resetting the images assoc. array back to array causing `removeImage` unable to search image sortKey and deleting the first image / unable to find index 0 as PHP silently handle `$arr[false]` as `$arr[0]`


alecritson (approved):
Uploaded an image and saved then re-saved, on `main` this does error as suggested, using this branch seems to fix the issue.

glennjacobs (approved):

PR #794, opened 2023-01-10
[0.2] Feat - Menu layout refactor

alecritson:
This PR looks to introduce a number of tweaks to the side menu in the hub. Currently we have issues such as:

- Currently we have duplicate logic for displaying desktop and mobile menu's
- Collapsing the menu into icon's which tooltips doesn't save much screen space and also introduces some UX issues when trying to show nested menu items
- Item's such link links are a bit duplicated within loops leading to level of maintenance overhead we should avoid
- The menu doesn't scroll if the number of menu items is longer than the window height, making it unusable.
- When on a sub page the parent menu item was not shown as active and it wasn't as obvious what page you were viewing

What this PR looks to solve...

- Moved menu display into one component so we only have to update in one place
- Menu will no longer collapse into icons, instead the user can decide to completely hide the menu, freeing up the entire available screen space, this should persist between page loads
- The side menu is now scrollable when the items overflow
- There is a slight shift in the layout so the top bar that goes across is now fixed, however the padding has been reduced to minimise screen size issues.

Misc updates

- Changed font to Poppins

With menu
image

On a sub page with active highlighting
image

Menu hidden
image


glennjacobs:
@alecritson looks good. All I'd say is the JS open/close could do with a nice transition. 

`transition-all duration-500` might do it?

alecritson:
> @alecritson looks good. All I'd say is the JS open/close could do with a nice transition.
> 
> `transition-all duration-500` might do it?

Added a small animation

glennjacobs:
> > @alecritson looks good. All I'd say is the JS open/close could do with a nice transition.
> > `transition-all duration-500` might do it?
> 
> Added a small animation

Can we do the same for the Products sub-menu, so that also has a subtle animation?

glennjacobs (approved):

PR #796, opened 2023-01-11
Change of UI on discount limitations to use popover

ryanmitchell:
This PR changes the discount limitations UI to use pop-overs/slide-outs for collection and brand selection, rather than the tree and list components currently in place.

As part of this PR a new brand search component has been introduced.


alecritson (dismissed):

alecritson:
Thanks @ryanmitchell I think this is a good change!

glennjacobs (approved):

PR #797, opened 2023-01-11
Fix issue when discount relation dates are null

ryanmitchell:
This PR fixes an issue when starts_at and ends_at dates are null on channelables and customer group relations to discounts, causing the discount to not be selectable.


ryanmitchell:
Tests are running fine for me locally, could the error be a Github issue?

alecritson:
@ryanmitchell I've updated this PR from `main` which should hopefully have a fix for the test failing (was unrelated to this PR)

My only question from this would be, if a Discount doesn't have a start date, should it be considered as valid? @glennjacobs

ryanmitchell:
The change doest affect discount start/end dates - its the the start/end for the groups and channels, which arent editable through the UI so therefore default to NULL.

alecritson:
@ryanmitchell I might be looking in the wrong place, when editing a discount I can see this section:

image

When clicking the the link text I get the prompt to choose dates:

image


Or am I misunderstanding?

ryanmitchell:
No I missed that they were clickable to set the dates.

But if you dont set a date they still default to NULL (as I did), so either thats acceptable or the channel/group start date needs to be defaulted to the same as the discount one?

alecritson:
Yeah I'm with you now, got a bit confused. This PR is correct as if you don't have a date set for channel or group it's considered always available.

alecritson (approved):

ryanmitchell:
Thanks, I should have been clearer!

glennjacobs (approved):

PR #803, opened 2023-01-12
fix: product page bottom bar not full width when side bar collapsed

wychoong:
product page bottom bar not full width when side bar collapsed

before
image

after
image


alecritson:
Thanks for this @wychoong I didn't see you did this PR until I did mine, the issue with this one (after I tried `w-full`) was that the bar goes over the menu, also it needed some tweaks to other views. Please see #805

PR #806, opened 2023-01-12
Improve UX for image manager

rubenvanerk:
#787 but targeting main


packages/admin/resources/views/components/input/toggle.blade.php
```diff
@@ -21,7 +21,7 @@
     class="{{ $on ? 'bg-green-500' : 'bg-gray-200' }} relative inline-flex flex-shrink-0 h-6 w-11 border-2 border-transparent rounded-full disabled:opacity-50 disabled:cursor-not-allowed cursor-pointer transition-colors ease-in-out duration-200 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-green-500"
     role="switch"
     aria-checked="false"
-    @if($disabled ?? true) disabled @endif
+    @disabled($disabled ?? true)
```
alecritson, line 24: Looks like this directive was added in Laravel 9, we're still trying to support L8 so features need to be backwards compatible.

alecritson (changes requested):

packages/admin/resources/views/components/input/toggle.blade.php
rubenvanerk: Sorry about that, reverted it!

alecritson (approved):
Thanks for this @rubenvanerk I spotted another instance of `@disabled` but correct it, looks good will get it in :)

PR #807, opened 2023-01-13
Fix/get_class of extended polymoprhic model

### LINKED ISSUE #767: Different "Url" morph class while loading and saving an extended model
- Lunar version: 0.1.5
- Laravel Version: 9.x
- PHP Version: 8.1
- Database Driver & Version:

### Actual Behaviour:

If we extend a model which has `HasUrls` trait, then Url morph class would be different while loading and saving the model in admin hub.

### Steps To Reproduce:

1. Extend a model with `HasUrls` trait (for example: `Lunar\Models\Product`) and register it:
```php
$models = collect([
    ...
    \Lunar\Models\Product::class => \App\Models\Product::class,
]);
ModelManifest::register($models);
```

2. Save the model through admin hub. The morph class (`element_type` field) of Url would be `App\Models\Product`.

3. Reload saved product in admin hub. `urls` property is empty, because `HasModelExtending::getMorphClass()` returns `Lunar\Models\Product` and would result to an empty query result.

### NOTE: 
Seems the problem is here:
https://github.com/lunarphp/lunar/blob/8d32958c53b673b4fcfc9efafef5c19dc325b551/packages/admin/src/Http/Livewire/Traits/HasUrls.php#L186

Shouldn't `get_class` be replaced with `$model->getMorphClass()` ?

wychoong:
fixes #767 

replace #749

### DIFF
--- a/packages/admin/src/Actions/Pricing/UpdatePrices.php
+++ b/packages/admin/src/Actions/Pricing/UpdatePrices.php
@@ -60,7 +60,7 @@ class UpdatePrices
             'customer_group_id' => $groupId,
             'tier' => $tier,
             'priceable_id' => $owner->id,
-            'priceable_type' => get_class($owner),
+            'priceable_type' => $owner->getMorphClass(),
         ]);
 
         $priceModel->save();
--- a/packages/admin/src/Actions/Pricing/UpdateTieredPricing.php
+++ b/packages/admin/src/Actions/Pricing/UpdateTieredPricing.php
@@ -72,7 +72,7 @@ class UpdateTieredPricing
             'customer_group_id' => $groupId,
             'tier' => $tier,
             'priceable_id' => $owner->id,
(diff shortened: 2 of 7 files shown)


wychoong:
@alecritson i updated a few more places that have `get_class`, while not being to test all as I do not have all models extended. Also I changed the base to `main` as there are usage of `get_class` in new files from `main` branch

also there are 2 places I'm not sure should it be changed
https://github.com/lunarphp/lunar/blob/619fa1a2aa8f4c9193c5acf873dd6bb9e8348afa/packages/admin/src/Views/Components/ModelUrl.php#L35

https://github.com/lunarphp/lunar/blob/619fa1a2aa8f4c9193c5acf873dd6bb9e8348afa/packages/admin/src/Http/Livewire/Traits/HasSlots.php#L156

alecritson (approved):

glennjacobs (approved):

PR #813, opened 2023-01-17
Add products to discount limitations

ryanmitchell:
Following on from #796 this PR allows discounts to be limited by product (note: not product variant).

It follows the same approach set out elsewhere and makes use of the DiscountPurchasable table using a 'limitation' type. 

Some of the files that appear changed are from 796, so the diff should clean up once its merged.


glennjacobs:
All looks good, but I feel a test should be introduced in the core to cover the new product criteria.

ryanmitchell:
Test added :)

glennjacobs (approved):

packages/core/src/DiscountTypes/Discount.php
```diff
@@ -115,6 +116,12 @@ private function getEligibleLines(Cart $cart)
             });
         }
 
+        if ($productIds->count()) {
+            $lines = $lines->reject(function ($line) use ($productIds) {
+                return ! $productIds->contains(get_class($line->purchasable->product).'::'.$line->purchasable->id);
+            });
+        }
```
wychoong, line 123: @ryanmitchell correct me if I'm wrong, this line is getting the class name of the `\Lunar\Models\Product` but concatenate with product variant id?

ryanmitchell: Yes it should be updated to be `$line->purchasable->product->id`... i was originally targeting product variants and later changed.

PR #814, opened 2023-01-18
Increment discount uses as part of order creation

ryanmitchell:
This PR adds incrementing the `uses` column of discounts as part of the order creation process. 

This is likely not the right approach, but is intended to kick start a conversation as to how is best to achieve this. 

Things that should be considered:
- Should discounts be associated with an order so that you can know which orders had which discount applied? (maybe as an order line)
- Should incrementing only be done once payment has been received, and if so how do we approach that.


glennjacobs:
Could you include a test?

ryanmitchell:
Test added

glennjacobs (approved):

PR #826, opened 2023-01-26
When DB::table() is used make sure connection is passed

ryanmitchell:
Just noticed a few DB methods missing `::connection()` when updating to 0.2-rc1.


packages/admin/src/Http/Livewire/Components/Orders/OrdersTable.php
```diff
@@ -83,9 +83,9 @@ public function build()
             SelectFilter::make('tags')->options(function () {
                 $tagTable = (new Tag)->getTable();
 
-                $tags = DB::table(
-                    config('lunar.database.table_prefix').'taggables'
-                )->join($tagTable, 'tag_id', '=', "{$tagTable}.id")
+                $tags = DB::connection(config('lunar.database.connection'))
```
alecritson, line 86: Will `connection()` through an error if this isn't set in the config 🤔

ryanmitchell: Nope it should fallback to the default connection:
https://github.com/laravel/framework/blob/e6846df19eed9d9dede821d15e02315653319652/src/Illuminate/Database/DatabaseManager.php#L90

alecritson (approved):

glennjacobs (approved):

PR #834, opened 2023-01-31
[0.2] Hotfix - fix pluck on null

### LINKED ISSUE #830: Using pluck() on null on discount creation page
- Lunar version: 0.2.0-rc1
- Laravel Version: 9.48.0
- PHP Version: 8.1.3

### Expected Behaviour:
Show the create form of discounts

### Actual Behaviour:
Throws error
```
Call to a member function pluck() on null
```
pointing to "packages/admin/resources/views/partials/forms/discount/limitations.blade.php"
https://github.com/lunarphp/lunar/blob/937d946f17eb2fc9c79a3bbc6eaa5304118c4885/packages/admin/resources/views/partials/forms/discount/limitations.blade.php#L144-L147

### Steps To Reproduce:
Open discounts tab then click on "Create"
The install was upgraded from a previous version of lunar (upgrade guide has been followed) and is seeded with the demo store datas (not using demo store only its seeder)

wychoong:
fixes #830 

bug introduce in #813


alecritson:
Thanks @wychoong didn't realise how far out pint was so i've push that up, if you can update your branch it should be easier to see what the changes are.

wychoong:
> Thanks @wychoong didn't realise how far out pint was so i've push that up, if you can update your branch it should be easier to see what the changes are.

done. maybe use GHA for the pint 😜

packages/admin/src/Http/Livewire/Components/Discounts/AbstractDiscount.php
```diff
@@ -116,18 +116,15 @@ public function mount()
     {
         $this->currency = Currency::getDefault();
 
-        $this->selectedBrands = $this->discount->brands->map(fn ($brand) => $this->mapBrandToArray($brand));
-        $this->selectedCollections = $this->discount->collections->map(fn ($collection) => $this->mapCollectionToArray($collection));
+        $this->selectedBrands = $this->discount->brands->map(fn ($brand) => $this->mapBrandToArray($brand)) ?? collect();
```
alecritson, line 119: Is there a reason for `?? collect()` since the relationship should always return a collection instance?

wychoong: should be no, will remove it

alecritson (approved):

PR #839, opened 2023-02-01
Fix validation of attributes of type TranslatedText

fakeheal:
Applying additional validation rules (specified when adding/updating an attribute) to attributes of type `TranslatedText` is wrong. It is applying the rule to the field that is [a collection of items](https://github.com/lunarphp/lunar/blob/main/packages/core/src/FieldTypes/TranslatedText.php#L59). So if the rule `string` was added to the attribute it would validate the collection/array of items against it and **always** fail.

**Note**: If you're testing it and your attribute has the `RichText` feature enabled and is also required, the js plugin used for [WYSIWYG adds `<p><br></p>`](https://github.com/quilljs/quill/issues/1235) as content even when all the text is deleted and it looks like the required rule isn't working.

### DIFF
--- a/packages/admin/src/Http/Livewire/Traits/WithAttributes.php
+++ b/packages/admin/src/Http/Livewire/Traits/WithAttributes.php
@@ -187,22 +187,31 @@ trait WithAttributes
             }
 
             $validation = $attribute['validation'] ? explode(',', $attribute['validation']) : [];
-
             $field = $attribute['signature'];
 
-            if (($attribute['required'] ?? false) || ($attribute['system'] ?? false)) {
-                if ($attribute['type'] == TranslatedText::class) {
-                    // Get the default language and make that the only one required.
-                    $field = "{$attribute['signature']}.{$this->defaultLanguage->code}";
-                }
+            $isRequired = ($attribute['required'] ?? false) || ($attribute['system'] ?? false);
 
+            // TranslatedText values are in an array, apply rules to each item of the array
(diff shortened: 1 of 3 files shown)


alecritson:
Thanks for this @fakeheal . I'd be good to see a small test to ensure this is kicking in as expected.

fakeheal:
@alecritson Sure! Is there a guide or something for starting up the existing tests or/and any guidelines?

alecritson:
> @alecritson Sure! Is there a guide or something for starting up the existing tests or/and any guidelines?

We have some tests already in place i.e. https://github.com/lunarphp/lunar/blob/main/packages/admin/tests/Unit/Http/Livewire/Components/Products/ProductCreateTest.php#L89 so you can use these as a base if that helps. 

Also on Livewire's site there is some helpers for testing https://laravel-livewire.com/docs/2.x/testing#introduction

If you're unsure or get stuck let us know and we can help

fakeheal:
@alecritson Thanks for the links. 

I just cloned my fork, ran `composer install` and then (using php8.1) ran `phpunit` (with `--filter` to "target" `ProductCreateTest`). Hope that's the correct way!

I added some test cases to the existing file testing mix of `FieldType`s that are required/optional + additional rules.

alecritson (approved):

alecritson:
Thanks for this @fakeheal We're just finalising the sign off then we'll get this merged!

glennjacobs (approved):

PR #840, opened 2023-02-01
[0.2] Hotfix - Show delete discount 

wychoong:


alecritson:
Thanks @wychoong Can you target the `0.2` branch please. :)

alecritson:
@wychoong Looks like the UI is a bit off, do you think this should appear in the left column rather than spanning across the whole page?

wychoong:
> @wychoong Looks like the UI is a bit off, do you think this should appear in the left column rather than spanning across the whole page?

image

done. I also moved the Save button to bottom panel to align with other pages. and refactored the bottom panel to a separate component

alecritson (approved):

glennjacobs (approved):

PR #841, opened 2023-02-01
[0.2] Fix/discount by product

wychoong:
issues:
1. discount's conditions checking
     - if cart has coupon but discount doesn't have coupon code won't apply, but discount without code should auto apply
     - minSpend not using correct precision
2. fixedValue discount using hardcoded precision `* 100`
3. small refactor and introduce `checkDiscountConditions` method to allow other discountType to extend

### DIFF
--- a/packages/core/src/DiscountTypes/AbstractDiscountType.php
+++ b/packages/core/src/DiscountTypes/AbstractDiscountType.php
@@ -65,7 +65,7 @@ abstract class AbstractDiscountType implements DiscountTypeInterface
         $cartCoupon = strtoupper($cart->coupon_code ?? null);
         $conditionCoupon = strtoupper($this->discount->coupon ?? null);
 
-        $validCoupon = $conditionCoupon ? ($cartCoupon === $conditionCoupon) : true;
+        $validCoupon = $cartCoupon && $conditionCoupon ? ($cartCoupon === $conditionCoupon) : true;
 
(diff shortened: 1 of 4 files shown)


packages/core/src/DiscountTypes/AbstractDiscountType.php
```diff
@@ -38,4 +40,41 @@ public function markAsUsed(): self
...
+
+    /**
+     * Return the eligible lines for the discount.
+     *
+     * @param  Cart  $cart
+     * @return Illuminate\Support\Collection
+     */
+    protected function getEligibleLines(Cart $cart): Collection
+    {
+        return $cart->lines;
```
alecritson, line 52: Looks like we're just returning the lines here, shouldn't there be some logic in here?

wychoong: no, its intended. This is because the usage in `checkDiscountConditions`.

the purpose is:
`checkDiscountConditions`: default conditions checking whether the discount are applicable (coupon code, min spend, max uses)
`getEligibleLines`: get the eligible lines for the discount

since this class is the base, so this function got no logic.
and using `Discount` type as example, it should be override in the class if it has more logic (brand, product, collection etc)

alecritson: Gotcha, sorry I didn't see it was on the Abstract class 🙈

glennjacobs (approved):

PR #845, opened 2023-02-06
Fix/error when creating discount

ryanmitchell:
This PR fixes an error when creating a discount - as selectedProducts is otherwise not available.


wychoong:
this is covered by #834

PR #847, opened 2023-02-07
[0.2] Hotfix - Discount issues

ryanmitchell:
Theres overlap with PR #841 here but I wanted to push up as conversation on it seems to have stalled.

This fixes up a few issues I've found on discounts:

1. Cache getEligibleLines to avoid re-running queries and logic
2. Handle there not being a condition coupon, which per the UI indicates the discount should be added automatically
3. Applies currency factoring

One thing I've noticed in testing is that percentage discounts don't apply a cart level discount value (`cartDiscountAmount`) in the way fixed discounts do. Is this intentional?

### DIFF
--- a/packages/core/src/DiscountTypes/Discount.php
+++ b/packages/core/src/DiscountTypes/Discount.php
@@ -33,7 +33,7 @@ class Discount extends AbstractDiscountType
         $cartCoupon = strtoupper($cart->coupon_code ?? null);
         $conditionCoupon = strtoupper($this->discount->coupon ?? null);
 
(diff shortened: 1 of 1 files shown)


alecritson (approved):

glennjacobs (approved):

alecritson:
@ryanmitchell Looks like there's some conflicts since the other PR was merged in. Do you have casting your eyes over it, I don't want to try myself and get it wrong 🙈

ryanmitchell:
Shouldnt be needed now the other has merged.

PR #855, opened 2023-02-14
Add documentation for lunar models to improve IDE assistance and enable better PhPstan support

kris914g:
This pull request adds PhpDocs properties to the Lunar models, specifying the types of their properties. This will enable IDEs to provide better code completion and type hinting for these models, making it easier for developers to work with them. It will also help prevent errors caused by incorrect data types in tools such as PhpStan. For example, without these PhpDocs properties, developers might encounter errors such as 'Access to an undefined property Lunar\Models\Order::$placed_at.' when analyzing the code with PhpStan. Adding these properties will improve the overall quality of the codebase and reduce the likelihood of bugs. Overall, this change will enhance the developer experience and improve the reliability of the Lunar project.

No code is affected since this PR only includes comments.

### DIFF
--- a/packages/admin/resources/views/partials/orders/lines.blade.php
+++ b/packages/admin/resources/views/partials/orders/lines.blade.php
@@ -9,7 +9,7 @@
                 @endif
 
                 <div class="flex-shrink-0">
-                    @if ($thumbnail = $line->purchasable->getThumbnail())
+                    @if ($thumbnail = $line->purchasable?->getThumbnail())
                         <x-hub::thumbnail :src="$thumbnail->getUrl('small')" />
                     @else
                         <x-hub::icon ref="photograph"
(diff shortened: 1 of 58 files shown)


alecritson:
Thanks for the PR, what did you use to generate this docs? Just thinking in the future how we keep this maintained.

kris914g:
> Thanks for the PR, what did you use to generate this docs? Just thinking in the future how we keep this maintained.

I created the model docs manually, which took some time. Unfortunately, there are no existing tools specifically for creating models in Laravel package development. While tools like https://github.com/barryvdh/laravel-ide-helper exist, which verify models against the database, they are only compatible with Laravel projects and not packages sadly.

glennjacobs (approved):

PR #857, opened 2023-02-14
[0.2] Hotfix - Add missing soft deletes traits

alecritson:


glennjacobs:
If we're going to start soft deleting these models, then how will they be restored?

alecritson:
We'd need to update the UI to handle it, however as it stands we've added the `deleted_at` columns but not applied the trait, which is wrong IMO since other models have `deleted_at` and do use the trait.

PR #859, opened 2023-02-15
Fix/brand table search

wychoong:
- enable Brand table search
- fix `save search` button do not show if its not enabled
- add `Brand` to search config

currently when applying search on table doesn't reset the page pagination, this causes issue when applying search in X page returns no result, due to pagination offset. should this be handled by table builder or manually


alecritson:
Hey @wychoong can we get this targeting `0.3` please?

wychoong:
> Hey @wychoong can we get this targeting `0.3` please?

done

PR #869, opened 2023-02-17
fix route name

wychoong:
1. enhance menu handle to support `array` 
  - collection page not showing active menu state

 **before**
image

**after**
image

2. change product type route name

**before:**
image

**after:**
image

### DIFF
--- a/packages/admin/src/Menu/MenuLink.php
+++ b/packages/admin/src/Menu/MenuLink.php
@@ -31,7 +31,7 @@ class MenuLink implements MenuItem
     /**
      * The handle for the menu link.
      *
-     * @var string
+     * @var string | array
      */
     public $handle;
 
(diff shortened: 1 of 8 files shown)


packages/admin/src/Menu/MenuLink.php
```diff
@@ -146,6 +146,8 @@ public function renderIcon($attrs = null)
      */
     public function isActive($path)
     {
-        return Str::startsWith($path, $this->handle);
+        return (bool) collect($this->handle)->first(function ($handle) use ($path) {
```
alecritson, line 149: If `$this->handle` is a string, will `collect()` explode the string into single characters? (just trying to remember past experience with similar)

wychoong: a quick test with tinker it will return as collection properly
image

https://github.com/laravel/framework/blob/975807b3d9bc2e6fc60e2a9a74cc49589a0d9585/src/Illuminate/Collections/Traits/EnumeratesValues.php#L966-L985
the (array) cast should handle it nicely

alecritson (approved):

PR #870, opened 2023-02-17
Buy X get Y: Ensure we dont add discounts for partial quantities

ryanmitchell:
When using a Buy X get Y discount i noticed if your quantity was more than 1 then it was possible to get `partial` discounts adding (e.g. 0.5), which would then throw an error on the price data type as it was getting a float rather than an int.

See config:

Screenshot 2023-02-17 at 15 57 23

This PR fixes it.

### DIFF
--- a/packages/core/src/DiscountTypes/BuyXGetY.php
+++ b/packages/core/src/DiscountTypes/BuyXGetY.php
@@ -107,7 +107,11 @@ class BuyXGetY
 
             $remainder = $rewardLine->quantity % $remainingRewardQty;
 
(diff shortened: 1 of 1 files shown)


alecritson:
Thanks @ryanmitchell We have been rewriting the discount logic and I've included this PR [here](https://github.com/lunarphp/lunar/pull/871/files#diff-5d96816d28333ea27ac0a577f24252dcd57d55c6083709d5f4046815ed5ab55eR110) which I hope is okay. I appreciate you sorting this though!

ryanmitchell:
No problem!

PR #871, opened 2023-02-20
[0.2] Hotfix - Fix discount logic

alecritson:
This PR looks to add a few changes/fixes to the way discounts work.

### Rename `Discount` type to `AmountOff`

The discount type `Discount` has now been renamed to `AmountOff` to add clarity to what it's actually doing.

### Fixed price discounts now distribute across lines

Currently fixed price discounts will be subtracted from the calculated sub total, now fixed price discounts will distribute across all eligible cart lines, for example:

Consider a fixed price discount of $10

#### Current

The discount would apply across the cart as a whole, so:

```php
$cart->discountTotal->value = 1000
```

#### New

```php
$cart->discountTotal->value = 1000

$firstCartLine->discountTotal->value = 334
$secondCartLine->discountTotal->value = 333
$thirdCartLine->discountTotal->value = 333
```

### Cart Line changes

Currently in order to display what a cart line sub total would be you'd have to add logic into your templates like so:

```php
@php
  $discountedSubTotal = new Price($line->subTotal->value - $line->discountTotal->value, $line->cart->currency, 1);
@endphp

<span class="strike">{{ $line->subTotal->formatted() }}</span> {{ $discountedSubTotal->formatted() }}
```

Now there is an additional `$subTotalDiscounted` property on the cart line:

```php
<span class="strike">{{ $line->subTotal->formatted() }}</span> {{ $line->subTotalDiscounted->formatted() }}
```

### Cart Pipeline changes

An additional `CalculateTax` action has been added to the pipelines, this allows us to sub total the cart, handle any discounts and then calculate the tax at the end without having to worry about overwriting what would have previously been calculated in previous pipelines.


glennjacobs:
Now that discounts are spread across the cart lines, should we look to remove `cartDiscountAmount`?

https://github.com/lunarphp/lunar/blob/hotfix/discount-amount-off-fix/packages/core/src/Models/Cart.php#L101

glennjacobs (approved):

PR #878, opened 2023-02-23
fix: use `::make` to create models in Hub for new resources

### LINKED ISSUE #877: Indexing products with custom models on creation fails with TypeError
- Lunar version: 0.2-rc1
- Laravel Version: 9.19
- PHP Version: 8.1
- Database Driver & Version: mysql 8

### Expected Behaviour:

The Observer's method `indexing()` to receive the custom Product model and not `Lunar\Models\Product`.

### Actual Behaviour:

**TypeError**:
App\Observers\ProductObserver::indexing(): Argument #1 ($product) must be of type App\Models\Product, Lunar\Models\Product given, called in /var/www/html/vendor/laravel/framework/src/Illuminate/Events/Dispatcher.php on line 441

### Steps To Reproduce:

1. Create `Product` model that extends `Lunar\Models\Product`
1. Add in `AppServiceProvider`:
```php
ModelManifest::register(collect([
    LunarProduct::class => Product::class,
]));
```
3. Create a `ProductObserver` with the following contents:
```php
<?php

namespace App\Observers;


use App\Models\Product;

class ProductObserver
{
    /**
     * Called when we're about to index the product
     **/
    {
        dd($product);
    }

    /**
     * Called when we are setting up the index via
     * php artisan lunar:meilisearch:update
     * */
    public function searchSetup(Product $product)
    {
        
    }
}
```
4. Add `Product::observe(ProductObserver::class);` to the `boot()` method of `EventServiceProvider`.
4. Attempt creating a product.

**Note:** Editing existing products fires the `indexing` event and no errors are shown.

fakeheal:
Draft PR that fixes #877.

I am not 100% sure I understand the entirety of the functionality that allows us to use custom models for Products, Customers, Orders, etc, but based on the comment of [benjaminhull on stackoverflow](https://stackoverflow.com/questions/32193613/can-you-create-a-new-model-instance-without-saving-it-to-the-database#comment132528449_53452394), if we replace the way we are creating new models using the constructor of model classes with `::make()`, the `newInstance` method in the trait `HasModelExtending` is called and the correct model is returned.

If someone provides a list of all "swappable" models I can go through the entire hub and replace `new Model([... attributes...])` with `Model::make([...attributes...])`.

### DIFF
--- a/packages/admin/src/Http/Livewire/Components/Products/ProductCreate.php
+++ b/packages/admin/src/Http/Livewire/Components/Products/ProductCreate.php
@@ -16,7 +16,7 @@ class ProductCreate extends AbstractProduct
      */
     public function mount()
     {
-        $this->product = new Product([
+        $this->product = Product::make([
             'status' => 'draft',
             'product_type_id' => ProductType::first()->id,
         ]);


glennjacobs:
> If someone provides a list of all "swappable" models

I believe all models in the core are "swappable".

wychoong:
i feel the model swapping is kind of fragile 😅 . can you make sure after the change it will not impact the existing record morph relationship. especially its the main Product model, like #767

glennjacobs:
> i feel the model swapping is kind of fragile 😅 . can you make sure after the change it will not impact the existing record morph relationship. especially its the main Product model, like #767

Yeah, I was never a massive fan the way models are overridden, however, it was a quick solution at the time.

I wouldn't be against suggestions on how we could do it better. I guess at a basic level we want a register for the classes and then a way of getting that reference. But previous quick attempts were pretty horrid.

glennjacobs:
Closing due to inactivity. Feel free to re-open if the update progresses.

PR #884, opened 2023-03-03
Discount breakdown

ryanmitchell:
This PR adds the concept of `discount breakdowns` to the cart, similar to tax breakdowns.

When a discount is added to the cart, it allows you to see what items were affected by that discount, and what the price of the individual discount is. This allows developers to add later pipelines to the cart that perform (for example) tax apportioning.

### DIFF
--- a/docs/src/.vuepress/config.js
+++ b/docs/src/.vuepress/config.js
@@ -127,6 +127,7 @@ module.exports = {
           '/extending/shipping',
           '/extending/tables',
           '/extending/taxation',
+          '/extending/validations',
         ]
       }
     ],
--- a/packages/core/src/Base/ValueObjects/Cart/DiscountBreakdowValue.php
+++ b/packages/core/src/Base/ValueObjects/Cart/DiscountBreakdowValue.php
@@ -1,0 +1,18 @@
+<?php
+
+namespace Lunar\Base\ValueObjects\Cart;
+
+use Illuminate\Support\Collection;
+use Lunar\DataTypes\Price;
+use Lunar\Models\Discount;
+
+class DiscountBreakdownValue
+{
+    public function __construct(
+        public Price $price,
+        public Collection $lines,
+        public Discount $discount,
+    ) {
+        //
+    }
+}
--- a/packages/core/src/Base/ValueObjects/Cart/DiscountBreakdown.php
+++ b/packages/core/src/Base/ValueObjects/Cart/DiscountBreakdown.php
@@ -1,0 +1,25 @@
+<?php
+
+namespace Lunar\Base\ValueObjects\Cart;
+
+use Illuminate\Support\Collection;
+
+class DiscountBreakdown
+{
+    public function __construct(
+        public ?Collection $discounts = null
+    ) {
+        $this->discounts = $discounts ?: collect();
+    }
+
+    /**
+     * Add a discount breakdown.
+     *
+     * @param  DiscountBreakdownValue  $discountBreakdownValue
+     * @return void
+     */
+    public function addDiscount(DiscountBreakdownValue $discountBreakdownValue)
+    {
+        $this->discounts->push($discountBreakdownValue);
+    }
+}
--- a/packages/core/src/DiscountTypes/BuyXGetY.php
+++ b/packages/core/src/DiscountTypes/BuyXGetY.php
@@ -122,7 +122,7 @@ class BuyXGetY
             }
             
             $affectedLines->push((object) [
-                'line' => $cartLine,
+                'line' => $rewardLine,
                 'quantity' => $qtyToAllocate,
             ]);
             
--- a/packages/core/src/DiscountTypes/BuyXGetY.php
+++ b/packages/core/src/DiscountTypes/BuyXGetY.php
@@ -2,6 +2,7 @@
 
 namespace Lunar\DiscountTypes;
 
+use Lunar\Base\ValueObjects\Cart\DiscountBreakdownValue;
 use Lunar\DataTypes\Price;
 use Lunar\Models\Cart;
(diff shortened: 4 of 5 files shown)


glennjacobs:
@ryanmitchell yes, this feels more of a "Lunar" approach, nice idea.

ryanmitchell:
Ok - this should be ready for a proper check over now - all the logic is in place that I think needs to be. 
Also let me know what other areas you would want test coverage for and I'll work on that.

packages/core/database/migrations/2023_03_03_100001_add_discount_breakdown_to_orders_table.php
```diff
@@ -0,0 +1,22 @@
...
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+use Lunar\Base\Migration;
+
+class AddDiscountBreakdownToOrdersTable extends Migration
+{
+    public function up()
+    {
+        Schema::table($this->prefix.'orders', function (Blueprint $table) {
+            $table->json('discount_breakdown')->nullable();
```
alecritson, line 12: Can we put this column in a suitable place, maybe before `discount_total` so it follows the same pattern as `tax_breakdown` / `tax_total`

packages/core/src/Actions/Carts/CreateOrder.php
```diff
@@ -120,7 +122,31 @@ public function execute(
                 ]);
             }
 
-            $order->lines()->createMany($orderLines->toArray());
+            $cartLinesMappedToOrderLines = [];
+            foreach ($orderLines as $orderLine) {
+                $orderLineModel = $order->lines()->create(collect($orderLine)->except(['cart_line_id'])->all());
+
+                if (isset($orderLine['cart_line_id'])) {
```
alecritson, line 129: Do we need `isset` since we're always mapping `cart_line_id` above?

packages/core/src/Base/Casts/DiscountBreakdown.php
```diff
@@ -0,0 +1,73 @@
...
+    /**
+     * Prepare the given value for storage.
+     *
+     * @param  \Illuminate\Database\Eloquent\Model  $model
+     * @param  string  $key
+     * @param  \Illuminate\Support\Collection  $value
+     * @param  array  $attributes
+     * @return array
+     */
+    public function set($model, $key, $value, $attributes)
```
alecritson, line 50: Looks like there is some repetition from the `createOrder` action, is there a way we can use this casting in there so we don't have to maintain in two places?

packages/core/src/DiscountTypes/AmountOff.php
```diff
@@ -88,6 +93,15 @@ private function applyFixedValue(array $values, Cart $cart): Cart
                 $cart->currency,
                 1
             );
+
+            // work out what share of the discount each line takes
+            $percentageShareOfDiscount = $line->subTotal->value / $linesSubtotal;
```
alecritson, line 98: Looks like we've set some variables here but we're not actually using them anywhere else?

packages/core/src/Actions/Carts/CreateOrder.php
ryanmitchell: Yeah we do as the shipping rows dont have a cart line

packages/core/src/Base/Casts/DiscountBreakdown.php
ryanmitchell: Im not sure of any way other than making a helper function and changing both places to use that?

packages/core/database/migrations/2023_03_03_100001_add_discount_breakdown_to_orders_table.php
ryanmitchell: Yeah thats done

packages/core/src/Base/Casts/DiscountBreakdown.php
alecritson: If it's being cast, wouldn't eloquent handle it when the model is saved to the database?

ryanmitchell: Maybe I'm misunderstanding, but Im not sure how you see it working. Is there a similar example elsewhere I can reference? The added complication in this case is changing each 'line' from a cart line to to a order line.

packages/core/src/Base/Casts/DiscountBreakdown.php
```diff
@@ -0,0 +1,73 @@
...
+    {
+        if (! $value instanceof Collection) {
+            return [
+                $key => json_encode($value),
+            ];
+        }
+
+        return [
+            $key => json_encode(collect($value)->map(function ($discountLine) {
```
alecritson, line 59: Would it be better to use the [toJson()](https://laravel.com/docs/10.x/collections#method-tojson) method available on collections

ryanmitchell: Thats changed

packages/core/src/Base/Casts/DiscountBreakdown.php
ryanmitchell: I've updated it. It doesnt save a huge amount of duplication - the data still needs mapped into a pretty similar format.

alecritson:
@ryanmitchell Thanks for your work on this, do you have an idea of what tests you will include and when so we can get this finished up?

ryanmitchell:
Docs and changelog both added

alecritson (approved):

wychoong:
@alecritson with this PR in, does that means `addApplied` and `getApplied` in DiscountManager are obsolete?

Issue #882, opened 2023-02-28
Saving an order causes a tax breakdown error


ryanmitchell:
Using rc-0.2

We have a process where after creating an order we sync with a 3rd party warehousing service and add a id to the `meta` so the two are kept in sync:

```php
            $meta = $order->meta ?? (object)[];
            $meta->pearl_id = 'xxxx';
            $order->meta = $meta;
            $order->save();
```

However now that we're using the Lunar tax service, we're getting an error at that point:

```
Lunar\Exceptions\InvalidDataTypeValueException
Value was "object" expected "int"
```

coming from `vendor/lunarphp/lunar/packages/core/src/Base/Casts/TaxBreakdown.php:27`

It seems the tax breakdown is being saved to the database as:

```json
[{"total": {"value": 0, "unitQty": 1, "currency": {"id": 1, "code": "GBP", "name": "British Pound", "default": 1, "enabled": 1, "created_at": "2023-01-18T08:26:30.000000Z", "updated_at": "2023-01-19T07:18:34.000000Z", "exchange_rate": "1.0000", "decimal_places": 2}}, "identifier": "T0", "percentage": 0, "description": "UK"}, {"total": {"value": 55, "unitQty": 1, "currency": {"id": 1, "code": "GBP", "name": "British Pound", "default": 1, "enabled": 1, "created_at": "2023-01-18T08:26:30.000000Z", "updated_at": "2023-01-19T07:18:34.000000Z", "exchange_rate": "1.0000", "decimal_places": 2}}, "identifier": "T20", "percentage": 20, "description": "UK"}]
```

instead of 

```json
[{"total": 0, "identifier": "T0", "percentage": 0, "description": "UK"}, {"total": 55, "identifier": "T20", "percentage": 20, "description": "UK"}]
```

Maybe something to do with the casting? Is there anything we can do about this?

glennjacobs:
@alecritson could you take a look at this? Would be good to get a fix into 0.2 before it's released.

ryanmitchell:
Closed by #887

PR #887, opened 2023-03-06
Fix tax breakdown on save

ryanmitchell:
When saving a cast-ed tax break down value if total is a price then save its value. This avoids any serialised/json-ed objects being saved to the database.

Fixes: #882

### DIFF
--- a/packages/core/src/Base/Casts/TaxBreakdown.php
+++ b/packages/core/src/Base/Casts/TaxBreakdown.php
@@ -63,7 +63,7 @@ class TaxBreakdown implements CastsAttributes, SerializesCastableAttributes
      */
     public function serialize($model, $key, $value, $attributes)
     {
-        return json_encode($value->map(function ($rate) {
+        return $value->map(function ($rate) {
             if ($rate->total instanceof Price) {
                 $rate->total = (object) [
                     'value' => $rate->total->value,
@@ -72,7 +72,6 @@ class TaxBreakdown implements CastsAttributes, SerializesCastableAttributes
                 ];
             }
             return $rate;
-        })
-        ->values());
+        })->toJson();
     }
 }
--- a/packages/core/src/Base/Casts/TaxBreakdown.php
+++ b/packages/core/src/Base/Casts/TaxBreakdown.php
@@ -54,10 +54,13 @@ class TaxBreakdown implements CastsAttributes, SerializesCastableAttributes
     }
     
     /**
-    * Get the serialized representation of the value.
-    *
-    * @param  array<string, mixed>  $attributes
-    */
+     * Get the serialized representation of the value.
+     *
+     * @param  \Illuminate\Database\Eloquent\Model  $model
+     * @param  string  $key
+     * @param  \Illuminate\Support\Collection  $value
+     * @param  array<string, mixed>  $attributes
+     */
     public function serialize($model, $key, $value, $attributes)
     {
         return json_encode($value->map(function ($rate) {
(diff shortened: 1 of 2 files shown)


ryanmitchell:
Hmm this actually seems to be throwing errors in the test suite no matter what approach I take.
Maybe the fix is to make [DataTypes\Price jsonable](https://laravel.com/api/6.x/Illuminate/Contracts/Support/Jsonable.html#method_toJson)?

ryanmitchell:
Got to the bottom of this - it seems Livewire is serialising the data which was causing the issue. You'll see I've fixed it up by implementing `[SerializesCastableAttributes](https://laravel.com/docs/10.x/eloquent-mutators#array-json-serialization)`

alecritson:
Thanks for taking a look @ryanmitchell , are the current tests sufficient for this? Feels like we might need something to prove this will work consistently.

ryanmitchell:
Well the live wire unit tests were failing for me without this fix. Do we maybe need to check the raw value from the database? I can add a test for it.  On 9 Mar 2023, at 15:08, Alec Ritson ***@***.***> wrote:﻿
Thanks for taking a look @ryanmitchell , are the current tests sufficient for this? Feels like we might need something to prove this will work consistently.

ryanmitchell:
I've added some test coverage to make sure the string 'currency' is not in the database value

packages/core/src/Base/Casts/TaxBreakdown.php
```diff
@@ -42,7 +42,37 @@ public function get($model, $key, $value, $attributes)
...
+     * Get the serialized representation of the value.
+     *
+     * @param  \Illuminate\Database\Eloquent\Model  $model
+     * @param  string  $key
+     * @param  \Illuminate\Support\Collection  $value
+     * @param  array<string, mixed>  $attributes
+     */
+    public function serialize($model, $key, $value, $attributes)
+    {
+        return json_encode($value->map(function ($rate) {
```
alecritson, line 66: Here's another place I think we can use the `toJson` helper?

alecritson:
Just to clarify then, the reason we was seeing the price object in the tax breakdown was due to Livewire? I assume from viewing the order in the hub?

packages/core/src/Base/Casts/TaxBreakdown.php
ryanmitchell: Updated

ryanmitchell:
Yeah that’s it - and would make sense why ti errored for our site too as we use livewire on the checkout pages.

packages/core/src/Base/Casts/TaxBreakdown.php
```diff
@@ -42,7 +42,37 @@ public function get($model, $key, $value, $attributes)
     public function set($model, $key, $value, $attributes)
     {
         return [
-            $key => json_encode($value),
+            $key => json_encode(collect($value)->map(function ($rate) {
+                if (! is_array($rate)) {
```
alecritson, line 46: Here we're checking if it's not an array, but could `rate` be anything other than an object which would cause the second statement to throw an exception?

ryanmitchell: Ideally it should never be an array - its a similar issue to the Discounts in that CreateOrder is sending arrays rather than the same format as returned by ->get()

alecritson (approved):

SebastianBotez:
Hello guys, Looks like there is still a problem on tax breakdown using Lunar 0.8.
The error occurs when this code is executed:
```
$order->update([
    'discount_breakdown' => $discountBreakdown,
]);
```
`Object of class stdClass could not be converted to string`

Any ideea hot to fix it?

PR #889, opened 2023-03-09
Add shipping tax correctly

ryanmitchell:
Shipping tax addition was using the wrong property.


alecritson:
Thanks @ryanmitchell Should this be targeting `0.2`?

ryanmitchell:
I've updated that

alecritson (approved):

PR #892, opened 2023-03-13
Add max user uses to discounts

ryanmitchell:
This PR adds the ability to limit the number of times a user can apply a discount code. 

It adds a new relation between discounts and users with a pivot table
It tracks when a logged in user uses a discount
It adds a new field to the admin hub to "Max uses per user", when this has a value if there is no logged in user then it will fail to add the discount.

### DIFF
--- a/packages/core/database/migrations/2023_03_03_100001_add_max_uses_per_user_to_discounts_table.php
+++ b/packages/core/database/migrations/2023_03_03_100001_add_max_uses_per_user_to_discounts_table.php
@@ -9,7 +9,7 @@ class AddMaxUsesPerUserToDiscountsTable extends Migration
     public function up()
     {
         Schema::table($this->prefix.'discounts', function (Blueprint $table) {
-            $table->mediumInteger('max_uses')->unsigned()->nullable()->after('max_uses');
+            $table->mediumInteger('max_uses_per_user')->unsigned()->nullable()->after('max_uses');
         });
     }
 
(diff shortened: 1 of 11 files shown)


glennjacobs:
@ryanmitchell can you avoid using `Auth::user()` and instead use `$cart->user` ?

packages/core/database/migrations/2023_03_03_100001_add_max_uses_per_user_to_discounts_table.php
```diff
@@ -0,0 +1,22 @@
...
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+use Lunar\Base\Migration;
+
+class AddMaxUsesPerUserToDiscountsTable extends Migration
+{
+    public function up()
+    {
+        Schema::table($this->prefix.'discounts', function (Blueprint $table) {
+            $table->json('max_uses_per_user')->nullable()->after('max_uses');
```
alecritson, line 12: Should this be a json field?

packages/core/src/DiscountTypes/AbstractDiscountType.php
```diff
@@ -39,6 +41,10 @@ public function markAsUsed(): self
     {
         $this->discount->uses = $this->discount->uses + 1;
 
+        if ($user = Auth::user()) {
```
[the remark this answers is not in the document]
ryanmitchell: Thanks - should be updated now

packages/core/database/migrations/2023_03_03_100001_add_max_uses_per_user_to_discounts_table.php
ryanmitchell: Definitely not.

alecritson:
@ryanmitchell Is this all finished up and ready?

ryanmitchell:
Yeah it is

PR #895, opened 2023-03-17
Cart meta cast

wychoong:
model `meta` attribute is cast into object by eloqeunt, however, the `Cart->add` is expecting `$meta` to be `array` type

this causes issue when trying to copy cartLines directly

eg: 
```php
 $this->checkoutCart->addLines($this->cart->lines);
```

would like to change the model $casts to other than `object` but seems none of the eloquent casting is suitable without impacting the usage of `$cart->meta->meta_field`

another way would be in userland to map all lines to cast `meta` to array but that isn't convenient to clone from one cart to another


alecritson:
Should this be targetting `0.2`?

wychoong:
> Should this be targetting `0.2`?

Done

glennjacobs (approved):

PR #897, opened 2023-03-17
calculate cartline unit price based on cart user, rather than Auth::user

kylekanderson:
Currently, the GetUnitPrice class that's used to calculate CartLine unit prices in the Cart Pipeline relies on `Auth::user()` to fetch the CustomerGroups used to calculate the CartLine's pricing.

However, there are instances where `Auth::user()` would not yield the correct CustomerGroups that apply to the cart (when running a queued job, for example), as there is no current authenticated user in that context.

Instead, this Pull Request relies on the cart's user to get the applicable CustomerGroups.


alecritson (dismissed):

alecritson:
Thanks @kylekanderson I think this would be useful in `0.2`, can we target that branch please.

kylekanderson:
> Thanks @kylekanderson I think this would be useful in `0.2`, can we target that branch please.

Sure thing. I just updated this to target `0.2`.

glennjacobs:
@alecritson I'm happy with this, can you sort the merge conflict and merge?

alecritson (approved):

PR #899, opened 2023-03-23
Optimize shipping modifiers

wychoong:
- enhance `getShippingOption` to check if shipping option is set on the address before looping options..
- added a method `addOptions` to ShippingManifest to directly set `options` without checking existing options
    -  in our project we define the shipping price by day and slot (eg: 23/3/2023 9am-12pm, $5), this can results to a lot of options and current `addOption` is not ideal. the new `addOptions` directly set `$options` without check exist & loop
- add `getOptionUsing` closure to further speed up option retrieving
    - continue of the above scenario, currently every time to retrieve an option, it requires to define all options then looping thru the collection. which is redundant and might be slow if a lot options. this new closure allows direct retrieve by identifier
    - ```php
        ShippingManifest::getOptionUsing(function (Cart $cart, $identifier): ShippingOption {
            $slot = DeliverySlot::fromIdentifier($identifier);

            return $slot->toShippingOption($cart->currency);
        });
```


alecritson:
@wychoong Could this target `0.3` please and I'm going to set it to draft until there are some tests and then we can look to review and get it in.

wychoong:
the changes is a mess after changing the base, closing this for #1074

PR #901, opened 2023-03-24
remove duplicate refresh method on CachesProperties trait

kylekanderson:
The `packages/core/src/Base/Traits/CachesProperties.php` file contains a duplicated `refresh()` method. This breaks composer when attempting to install the package. Both instances of the method were exactly the same, so I just removed one of them.

### DIFF
--- a/packages/core/src/Base/Traits/CachesProperties.php
+++ b/packages/core/src/Base/Traits/CachesProperties.php
@@ -13,21 +13,6 @@ trait CachesProperties
             $model->restoreProperties();
         });
     }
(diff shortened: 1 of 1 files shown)


alecritson:
Good spot! Thank you. I'll change the target to `0.2` and get this up asap

alecritson:
@kylekanderson Looks like somethings up with your origin, if you could sync it up then we can get this in, or close and reopen a new PR whichever suits.

glennjacobs (approved):

PR #909, opened 2023-03-31
Stop popover closing on click away

ryanmitchell:
See following screen recording for a example of the issue caused by this:

https://user-images.githubusercontent.com/51899/229046628-34421ad5-5242-499e-8b7e-b77faf6ae83f.mov

This PR removes the click away function - there is already a close button so its how the popover is closed.


alecritson:
@glennjacobs I'm okay with this, how about you? Also would it be in `0.3` ?

glennjacobs:
> @glennjacobs I'm okay with this, how about you? Also would it be in `0.3` ?

Fine by me, and happy for any simple bug fixes to be incorporated into 0.3.

ryanmitchell:
Yep and yep - both done

PR #924, opened 2023-04-03
Fix incorrect totals due to discounts

ryanmitchell:
The Calculate pipeline hadn't been updated since discounts stopped affecting line totals, so they were generating the wrong totals.

This PR changes the approach to calculating the final cart total.

### DIFF
--- a/packages/core/src/Pipelines/Cart/Calculate.php
+++ b/packages/core/src/Pipelines/Cart/Calculate.php
@@ -17,8 +17,8 @@ class Calculate
     {
         $discountTotal = $cart->lines->sum('discountTotal.value');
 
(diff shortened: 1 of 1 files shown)


ryanmitchell:
Ah I forgot you want discount before subtotal. But when I change that other tests fail.
Either way, use this PR as highlighting there is an issue - you probably have a better approach to fixing it.

alecritson:
@ryanmitchell Is this something that could be solved by adding `subtotalDiscounted` to the cart? I've had a bit of testing on a site we're developing and the totals (with tax) seemed ok.

Otherwise are you able to provide a failing test?

PR #936, opened 2023-04-05
Fix accumulating discount breakdowns

ryanmitchell:
As cart calculation can be called multiple times in a request the discount breakdowns were duplicating.

This PR fixes that by removing any for that discount before adding.

### DIFF
--- a/packages/core/src/DiscountTypes/AbstractDiscountType.php
+++ b/packages/core/src/DiscountTypes/AbstractDiscountType.php
@@ -97,12 +97,6 @@ abstract class AbstractDiscountType implements DiscountTypeInterface
      */
     protected function addDiscountBreakdown(Cart $cart, DiscountBreakdown $breakdown)
     {
-        if (! $cart->discountBreakdown) {
-            $cart->discountBreakdown = collect();
-        }
-
-        $cart->discountBreakdown = $cart->discountBreakdown->reject(fn ($b) => $b->discount == $breakdown->discount);
-
         $cart->discountBreakdown->push($breakdown);
 
         return $this;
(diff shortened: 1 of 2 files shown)


packages/core/src/DiscountTypes/AbstractDiscountType.php
```diff
@@ -101,6 +101,8 @@ protected function addDiscountBreakdown(Cart $cart, DiscountBreakdown $breakdown
             $cart->discountBreakdown = collect();
         }
 
+        $cart->discountBreakdown = $cart->discountBreakdown->reject(fn ($b) => $b->discount == $breakdown->discount);
```
alecritson, line 104: What is this ultimately checking for uniqueness? For example you could have multiple discounts applied to the same cart of the same type which are perfectly valid.

ryanmitchell: Its not the 'type' of discount but the discount model - that should only be applied once per cart?

ryanmitchell: I suppose the other way to fix it would be to reset `discountBreakdown` in discount manager :: apply()?

ryanmitchell:
Any thoughts on this @alecritson ?

packages/core/src/Pipelines/Cart/ApplyDiscounts.php
```diff
@@ -15,6 +15,8 @@ final class ApplyDiscounts
      */
     public function handle(Cart $cart, Closure $next)
     {
+        $cart->discountBreakdown = collect([]);
```
alecritson, line 18: This feels a lot cleaner I think, nice and simple. You happy with this @glennjacobs ?

glennjacobs: Seems fine to me, if you're happy then feel free to approve.

alecritson (approved):

PR #938, opened 2023-04-05
BuyXGetY - fix count for how many condition qty to allocate

ryanmitchell:
The count for how many condition qty to allocate was incorrect, this PR fixes that.


glennjacobs:
I think we'd want a test for this, please.

ryanmitchell:
Yep - on it. Meant to open the PR as a draft.

glennjacobs (approved):

PR #949, opened 2023-04-17
Fixed `AmountOff` to share by percentage

ryanmitchell:
We're running into issues where the items in certain carts combined with discounts were giving negative subtotals due to how the amount off (fixed value) discount was being calculated.

I've re-worked the discount to take a different approach. I've also added test coverage for one of the basket combinations that caused us issues.


packages/core/src/DiscountTypes/AmountOff.php
```diff
@@ -110,9 +106,17 @@ private function applyFixedValue(array $values, Cart $cart): Cart
...
-            $line = $cart->lines->first(function ($line) use ($remaining) {
-                return (bool) (($line->subTotal->value - $line->discountTotal->value) - $remaining);
+                        
+            // prioritise sharing the remaining over eligible lines
+            $line = $lines->first(function ($line) use ($remaining) {
+                return $line->subTotalDiscounted->value - $remaining > 0;
             });
+            
+            if (! $line) {
+                $line = $cart->lines->first(function ($line) use ($remaining) {
```
alecritson, line 116: Should we be checking over all lines here? Shouldn't this just be eligible lines?

ryanmitchell: I think it should just be eligible but it was originally all lines so I didnt want to change it!

alecritson: Yeah looks like a mistake there 😅 Happy for it to be just eligible lines and then maybe it can be added to the logic above it to keep it concise

ryanmitchell: No problem - thats updated

packages/core/src/DiscountTypes/AmountOff.php
```diff
@@ -61,30 +61,26 @@ private function applyFixedValue(array $values, Cart $cart): Cart
...
         foreach ($lines as $line) {
-            if ($line->subTotal->value < $roundedChunk) {
-                $amount = $roundedChunk - ($roundedChunk % $line->subTotal->value);
-            } else {
-                $amount = $roundedChunk;
-            }
-
-            // If this discount already has a greater discount value
-            // don't add this one as they already have a better deal.
-            if ($line->discountTotal->value > $amount) {
```
alecritson, line 86: I can see this was removed, was there a reason behind this? If there is a previous discount which is better than this one, we don't want this one to apply I don't think.

ryanmitchell: The way it's working now is to compound the discount... so if you already have a discounted price then it uses that to work out what share of the discount you take. 

Ultimately you are still getting the same 'value' out of the discount as its shared across all the lines.

alecritson: I think myself and glenn spoke about this and not having discounts stack as it adds complexity, what do you think @glennjacobs ?

ryanmitchell: Just a heads up that if that’s the intention it’s not in place on the percentage discount - they compound.

glennjacobs: @alecritson Compounding discounts do indeed add a lot of complexity. It would be easier to avoid it.

@ryanmitchell is this something you need for your project? Trying to understand what impact it might have if we didn't support compounding discounts.

ryanmitchell:
I don’t need it.On 25 Apr 2023, at 10:58, Glenn Jacobs ***@***.***> wrote:﻿
@glennjacobs commented on this pull request.



In packages/core/src/DiscountTypes/AmountOff.php:
>          $remaining = $value;
 
         $affectedLines = collect();
 
         foreach ($lines as $line) {
-            if ($line->subTotal->value < $roundedChunk) {
-                $amount = $roundedChunk - ($roundedChunk % $line->subTotal->value);
-            } else {
-                $amount = $roundedChunk;
-            }
-
-            // If this discount already has a greater discount value
-            // don't add this one as they already have a better deal.
-            if ($line->discountTotal->value > $amount) {

@alecritson Compounding discounts do indeed add a lot of complexity. It would be easier to avoid it.
@ryanmitchell is this something you need for your project? Trying to understand what impact it might have if we didn't support compounding discounts.

ryanmitchell:
Do you guys plan to work on this or do you need me to update the code here to avoid compounding?

alecritson:
@ryanmitchell I think if you update what you changed and then we can take a look and see if anything needs changing on a new PR or something.

ryanmitchell:
No problem - thats done - https://github.com/lunarphp/lunar/pull/949/commits/f4d2fd94d03f91f0165f0a282e0e62d616f9ea9e

ryanmitchell:
Seems like can_apply_multiple_different_discounts needs to be updated (or removed) as it is a test that assumes compounding?

glennjacobs:
> Seems like can_apply_multiple_different_discounts needs to be updated (or removed) as it is a test that assumes compounding?

@alecritson could you take a look at this?

alecritson:
> > Seems like can_apply_multiple_different_discounts needs to be updated (or removed) as it is a test that assumes compounding?
> 
> @alecritson could you take a look at this?

From what I can see the compounding is due to it being a percentage, not fixed price, which from what I remember you greenlit? If you could cast your eyes on it too that would be great.

ryanmitchell:
Any thoughts on this? Its still causing issues for us on our live sites.

glennjacobs:
@ryanmitchell Alec and I have discussed this one and we're going to run with your suggestion. Could you re-target this PR to `main` and we'll look to get it in, thanks.

glennjacobs (changes requested):
Retarget to `main`

ryanmitchell:
Closing for #1207

PR #955, opened 2023-04-18
Hotfix [0.2] discount priority 

wychoong:
discount should apply from high to low 

`priority => 1: Low, 5: Medium, 10: High`

### DIFF
--- a/packages/core/src/Managers/DiscountManager.php
+++ b/packages/core/src/Managers/DiscountManager.php
@@ -165,7 +165,7 @@ class DiscountManager implements DiscountManagerInterface
                     $query->whereNull("{$joinTable}.ends_at")
                         ->orWhere("{$joinTable}.ends_at", '>', now());
                 });
-        })->orderBy('priority')->get();
+        })->orderBy('priority', 'desc')->get();
     }
 
     /**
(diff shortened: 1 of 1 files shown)


alecritson:
Has something happened by changing the order that's caused the tests to fail? Just wondering if there is a deeper issue.

wychoong:
> Has something happened by changing the order that's caused the tests to fail? Just wondering if there is a deeper issue.

the ordering definitely gonna impact the outcome of the calculation, but before fixing the test, we need to establish whether the current order is wrong

alecritson:
> > Has something happened by changing the order that's caused the tests to fail? Just wondering if there is a deeper issue.
> 
> the ordering definitely gonna impact the outcome of the calculation, but before fixing the test, we need to establish whether the current order is wrong

Yeah I think we do want to order High > Medium > Low

wychoong:
> > > Has something happened by changing the order that's caused the tests to fail? Just wondering if there is a deeper issue.
> > 
> > 
> > the ordering definitely gonna impact the outcome of the calculation, but before fixing the test, we need to establish whether the current order is wrong
> 
> Yeah I think we do want to order High > Medium > Low

should be ready now, added `orderBy('id')` for when discount is same priority

glennjacobs (approved):

PR #958, opened 2023-04-20
Fix - ensure product association exists

ryanmitchell:
If a product that has been added as an association has later been deleted then it throws an error in the admin hub on the product it was associated with.
 
This change works around it.

### DIFF
--- a/packages/admin/src/Http/Livewire/Components/Products/AbstractProduct.php
+++ b/packages/admin/src/Http/Livewire/Components/Products/AbstractProduct.php
@@ -694,6 +694,10 @@ abstract class AbstractProduct extends Component
         $this->associations = $this->product->associations
             ->merge($this->product->inverseAssociations)
             ->map(function ($assoc) {
+                if (! $assoc->target) {
+                    return;
+                }
+
                 $inverse = $assoc->target->id == $this->product->id;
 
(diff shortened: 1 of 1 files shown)


alecritson:
My initial thought was if the product has been soft deleted, should we be showing it still but "greyed" out or otherwise show it's been deleted 🤔 but then maybe it just disappearing and reappearing is just as okay, what do you think @glennjacobs ?

glennjacobs:
> My initial thought was if the product has been soft deleted, should we be showing it still but "greyed" out or otherwise show it's been deleted 🤔 but then maybe it just disappearing and reappearing is just as okay, what do you think @glennjacobs ?

From a user's perspective, if they've deleted something then they expect it to be completely gone. So not showing is correct I think.

However, if a user restores a deleted product should it restore all its associations also? I don't think there's much harm and it's probably more useful that it does, rather than not. 

So I think this is fine.

alecritson:
@glennjacobs If you're happy with this then lets get it merged.

glennjacobs (approved):

SebastianBotez:
If the product was deleted and you try to edit a product that has the deleted product associated in the associations tabs it trows an error: `Trying to get property id of null`

PR #960, opened 2023-04-20
Add shipping breakdown

alecritson:
This PR will add a new `shipping_breakdown` to orders and carts. This will allow developers to add additional shipping costs and provide a more complete breakdown similar to discounts and taxes.

The idea being that you may have outside factors which affect the cost of shipping outside of what's provided by the manifest, such as split delivery or additional charges for awkward items, this should facilitate that.

### DIFF
--- a/packages/core/tests/Unit/Models/CartTest.php
+++ b/packages/core/tests/Unit/Models/CartTest.php
@@ -570,7 +570,10 @@ class CartTest extends TestCase
         ]);
     }
 
-    /** @test */
+    /**
+    * @test
+    * @group thisone
+    */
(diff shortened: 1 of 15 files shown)


packages/core/tests/Unit/Models/CartTest.php
```diff
@@ -372,7 +372,10 @@ public function can_update_existing_cart_line()
         ]);
     }
 
-    /** @test */
+    /**
+    * @test
+    * @group thisone
```
glennjacobs, line 377: `thisone`

glennjacobs:
@alecritson can you fix the conflicts?

PR #963, opened 2023-04-24
Hotfix [0.2] - Improve discount extending

### LINKED ISSUE #959: Using a custom discount type in the admin hub is broken
If you make a custom discount type eg something as simple as

```php
<?php
namespace App\Lunar\DiscountTypes;
class AmountOffRetail extends \Lunar\DiscountTypes\AmountOff { }
?>
```

And register it using the Discounts Facade, then try to use it in the Lunar hub you get:

"Call to member function rules() on null" in partials/forms/discount.php coming from AbstractDiscount line 442.

The issue seems to be the hub is using Editing/DiscountTypes as the source of truth for what components are available, not the Discount facade? Maybe each discount needs a getComponent() method, rather than using this?

alecritson:
Closes #959

### DIFF
--- a/docs/admin-hub/extending.md
+++ b/docs/admin-hub/extending.md
@@ -344,7 +344,7 @@ class ProductUrls
 If you have registered your own discount types in the core, you will likely want to provide an interface so authenticated staff members can add the data required.
 
 :::warning
-You need to make sure you have registered your discount with the Lunar core beforehand.
+You need to make sure you have [registered your discount with the Lunar](/core/extending/discounts) core beforehand.
 :::
 
 ```php
--- a/docs/admin-hub/extending.md
+++ b/docs/admin-hub/extending.md
@@ -363,8 +363,9 @@ namespace App\Http\Livewire\Components;
 
 use Illuminate\Support\Facades\DB;
 use Lunar\Models\Discount;
+use Lunar\Hub\Http\Livewire\Components\Discounts\Types\AbstractDiscountType;
 
-class CustomDiscount extends AbstractDiscountType
+class CustomDiscountComponent extends AbstractDiscountType
 {
     /**
      * The instance of the discount.
--- a/docs/admin-hub/extending.md
+++ b/docs/admin-hub/extending.md
@@ -338,3 +338,97 @@ class ProductUrls
     \Lunar\Models\Product::class => \App\Storefront\ProductUrls::class,
 ],
 ```
+
+## Discounts
+
+If you have registered your own discount types in the core, you will likely want to provide an interface so authenticated staff members can add the data required.
+
+:::warning
+You need to make sure you have registered your discount with the Lunar core beforehand.
+:::
+
+```php
+use Lunar\Hub\Facades\DiscountTypes;
+
+DiscountTypes::register(DiscountType::class, DiscountTypeComponent::class);
+```
+
+This should then appear when editing a discount and the user has chosen your custom discount type.
+
+
+```php
+<?php
+
+namespace App\Http\Livewire\Components;
+
+use Illuminate\Support\Facades\DB;
+use Lunar\Models\Discount;
+
+class CustomDiscount extends AbstractDiscountType
+{
+    /**
+     * The instance of the discount.
+     *
+     * @var Discount
+     */
+    public Discount $discount;
+
+    /**
+     * {@ineheritDoc}.
+     */
+    public function rules()
+    {
+        return [
+            'discount.data' => 'array',
+        ];
+    }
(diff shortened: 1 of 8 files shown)


docs/admin-hub/extending.md
```diff
@@ -338,3 +338,97 @@ class ProductUrls
...
+<?php
+
+namespace App\Http\Livewire\Components;
+
+use Illuminate\Support\Facades\DB;
+use Lunar\Models\Discount;
+
+class CustomDiscount extends AbstractDiscountType
+{
+    /**
```
wychoong, line 369:
```suggestion
use Lunar\Models\Discount;
use Lunar\Hub\Http\Livewire\Components\Discounts\Types\AbstractDiscountType;

class CustomDiscountComponent extends AbstractDiscountType
{
    /**
```

docs/admin-hub/extending.md
```diff
@@ -338,3 +338,97 @@ class ProductUrls
...
+You need to make sure you have registered your discount with the Lunar core beforehand.
+:::
+
+```php
+use Lunar\Hub\Facades\DiscountTypes;
+
+DiscountTypes::register(DiscountType::class, DiscountTypeComponent::class);
+```
+
+This should then appear when editing a discount and the user has chosen your custom discount type.
```
wychoong, line 356:
```suggestion
```

docs/admin-hub/extending.md
```diff
@@ -338,3 +338,97 @@ class ProductUrls
...
+
```
wychoong, line 357:
```suggestion
Create a Livewire component to handle your custom discount type
```

docs/admin-hub/extending.md
```diff
@@ -338,3 +338,97 @@ class ProductUrls
...
+     *
+     * @return \Illuminate\View\View
+     */
+    public function render()
+    {
+        return view('my-custom-discount-ui')
+            ->layout('adminhub::layouts.base');
+    }
+}
+```
```
wychoong, line 434:
``````suggestion
```

Register the component corresponding to the discount type

```php
use Lunar\Hub\Facades\DiscountTypes;
use App\Http\Livewire\Components\CustomDiscountComponent;
use App\DiscountTypes\MyCustomDiscountType;

DiscountTypes::register(MyCustomDiscountType::class, CustomDiscountComponent::class);
 ```

The component should then appear when user has chosen the custom discount type.
``````

File: docs/admin-hub/extending.md [hunk 5]
```diff
@@ -338,3 +338,97 @@ class ProductUrls
     \Lunar\Models\Product::class => \App\Storefront\ProductUrls::class,
 ],
 ```
+
+## Discounts
+
+If you have registered your own discount types in the core, you will likely want to provide an interface so authenticated staff members can add the data required.
```
wychoong, line 344: should add a link to the discount doc

wychoong (commented):
hope these add some clarity

wychoong:
`Lunar\Hub\Editing\DiscountTypes` seems should be merged into `DiscountManager` and have the `DiscountManager::register(DiscountType::class, DiscountTypeComponent::class)` handle both registration with 1 method, it's breaking so maybe for 0.4?

alecritson:
> `Lunar\Hub\Editing\DiscountTypes` seems should be merged into `DiscountManager` and have the `DiscountManager::register(DiscountType::class, DiscountTypeComponent::class)` handle both registration with 1 method, it's breaking so maybe for 0.4?

We want to keep the core and hub classes separated, really the core shouldn't have any knowledge of the hub as much as possible.

wychoong:
> keep the core and hub classes separated

that make sense.

glennjacobs (approved):

ryanmitchell:
@alecritson Just testing this out as part of the latest 0.2 release.

I'm getting `Call to undefined method App\Models\User::authorize()` when using a custom component. I remember we had this issue with slots too, but I cant remember how we resolved it?

alecritson:
@ryanmitchell Was the issue related to the naming of the component namespace? i.e. it needs to have `lunar` in order to be picked up by the middleware?

ryanmitchell:
Yeah that was it - thanks

ryanmitchell:
A further issue - when using discounts with a coupon code they fail validation as [CouponValidator](https://github.com/lunarphp/lunar/blob/main/packages/core/src/Base/Validation/CouponValidator.php) is limited to `AmountOff` types only.

Would you intend that devs have to use their own validation class to add to validation logic? Or does coupon validator need a way to register other validations?

PR #964, opened 2023-04-24
Feat [0.3] - Storefront session

alecritson:
This PR looks to add a `StorefrontSession` helper for storefronts to set things like channels, customer groups and currencies. It can be expanded going forward but this is a start for the essentials.


glennjacobs:
@alecritson if we have Storefront Session pulled out into a "Storefront" section, should we have the same for the Cart Session?

glennjacobs:
@alecritson what happens if a Customer belongs to multiple Customer Groups?

alecritson:
> @alecritson if we have Storefront Session pulled out into a "Storefront" section, should we have the same for the Cart Session?

Yeah can pull it out on a separate PR, was done this way because there wasn't a suitable place for this.

alecritson:
> @alecritson what happens if a Customer belongs to multiple Customer Groups?

It should still match right, unless you wanted to specify multiple customer groups to make it really specific.

alecritson:
> > > @alecritson what happens if a Customer belongs to multiple Customer Groups?
> > 
> > 
> > It should still match right, unless you wanted to specify multiple customer groups to make it really specific.
> 
> Just thinking, if a User is logged in and the associated Customer record belongs to multiple Customer Groups, do we not need to consider them all for our queries?

We can certainly update the methods so it handles multiple.

glennjacobs:
@alecritson docs need updating to show setting multiple customer groups.

glennjacobs (approved):

PR #966, opened 2023-04-25
Hotfix [0.2] - Fix discount saving

### LINKED ISSUE #953: Can't remove expiry date from discount once entered
- Lunar version: 0.2.4
- Laravel Version: 9.52.5
- PHP Version: 8.1.10

### Expected Behaviour:
Can clear expiry date on a discount.

### Actual Behaviour:
Can't clear expiry date on a discount. Cross button highlights the date but does not clear it.
Manually highlighting it and deleting the text content does not seem to update the field value. Saving with a blank field does not remove the date once entered.

### Steps To Reproduce:
Create a discount sitewide, automatic discount without an expiry date (almost definitely not an exact requirement but both examples I had were in this configuration). Save.
Edit the discount, adding an expiry date. Save.
Attempt to remove the expiry date and fail to do so.

alecritson:
Closes #953


packages/admin/src/Http/Livewire/Components/Discounts/DiscountCreate.php
```diff
@@ -48,7 +48,7 @@ public function rules()
             'discount.max_uses' => 'nullable|numeric|min:0',
             'discount.max_uses_per_user' => 'nullable|numeric|min:0',
             'discount.priority' => 'required|min:1',
-            'discount.starts_at' => 'date',
+            'discount.starts_at' => 'required|date',
             'discount.coupon' => 'nullable',
             'discount.ends_at' => 'nullable|date|after:starts_at',
```
wychoong, line 53:
```suggestion
            'discount.ends_at' => 'nullable|date|after:discount.starts_at',
```
should be same as DiscountShow?

alecritson: Good spot thanks!

glennjacobs (approved):

PR #967, opened 2023-04-25
Feat [0.3] - Add storefront scopes

alecritson:


glennjacobs:
@alecritson I think this looks pretty good. Do we need to consider currencies? If we do, it could be a follow-up PR anyhow.

alecritson:
@glennjacobs I think it could come in a later release when we do more price work. I was however thinking of adding parameters so you can specify dates instead of insisting on `now()`, this way you could list products which are "due to be available" so to speak.

glennjacobs:
> @glennjacobs I think it could come in a later release when we do more price work. I was however thinking of adding parameters so you can specify dates instead of insisting on `now()`, this way you could list products which are "due to be available" so to speak.

Probably worth doing it now, whilst we think of it? As long as it defaults to `now()` if not specified.

glennjacobs:
@alecritson docs need updating to show setting multiple customer groups.

alecritson:
> @alecritson docs need updating to show setting multiple customer groups.

I've added some initial docs

glennjacobs (approved):

PR #969, opened 2023-04-25
Hotfix [0.2] fix cart address test assert

wychoong:
the test when asserting the CartAddress in database with Address instance is including the timestamps, and this causes issue when timestamp doesnt match

image

this should solve it by only extracting the fillable attributes used by the action and assert

### DIFF
--- a/packages/core/tests/Unit/Actions/Carts/AddAddressTest.php
+++ b/packages/core/tests/Unit/Actions/Carts/AddAddressTest.php
@@ -9,6 +9,7 @@ use Lunar\Models\Cart;
 use Lunar\Models\CartAddress;
 use Lunar\Models\Currency;
 use Lunar\Tests\TestCase;
+use function Livewire\invade;
 
 /**
  * @group lunar.actions
@@ -18,6 +19,27 @@ class AddAddressTest extends TestCase
 {
     use RefreshDatabase;
 
+    protected AddAddress $action;
+
+    protected array $fillableAttributes;
+
+    protected string $cartAddressTable;
+
+    public function setUp(): void
+    {
+        parent::setUp();
+
+        $this->action = new AddAddress;
+
+        $this->fillableAttributes = invade($this->action)->__get('fillableAttributes');
+
+        $this->cartAddressTable = (new CartAddress)->getTable();
+    }
+
+    protected function getCartAddressAttributesFromAddress(Address $address): array
+    {
+        return collect($address->getAttributes())->only($this->fillableAttributes)->toArray();
+    }
     /**
      * @test
      */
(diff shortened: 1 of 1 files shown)


ryanmitchell:
Might it be simpler to just freeze time in the tests? https://laravel-code.tips/you-can-freeze-time-in-tests/

wychoong:
> Might it be simpler to just freeze time in the tests? https://laravel-code.tips/you-can-freeze-time-in-tests/

Yes and no. Freezing hide the error but the test still asserting the unnecessary attributes

alecritson:
Thanks @wychoong for this, I agree it's a problem but I think this adds a level of complexity I don't think is required. If the timestamps are causing the issue we should just remove them from the check completely.

wychoong:
> Thanks @wychoong for this, I agree it's a problem but I think this adds a level of complexity I don't think is required. If the timestamps are causing the issue we should just remove them from the check completely.

The complexity is the refactoring or the livewire/invade?

I can think of few ways
- revert the changes, then
- Make the fillable public, or duplicate into the test,
- then collect()->only() for the assert

alecritson:
@wychoong I think I would agree with the approach @ryanmitchell mentioned (freezing time) as this would then solve the problem for any future tests. @glennjacobs ?

glennjacobs:
> @wychoong I think I would agree with the approach @ryanmitchell mentioned (freezing time) as this would then solve the problem for any future tests. @glennjacobs ?

Freeze time should work well, and we could use it in other areas potentially. Would it be too risky to put this on the main `TestCase`?

wychoong:
> Would it be too risky to put this on the main `TestCase`

IMO freezing time is kind of dirty hack and may blind us from actual issue in the future. A not so good example is pivot model dates, are there risk of asserting the wrong ones, I not sure as not enough experience with tests 😓 

What this PR trying to fix is to assert only attributes we interested with. personally would prefer manually unset the attributes or duplicate the fillable over freeze time, but i did the `invade` way to avoid duplication and lost sync

glennjacobs:
> > Would it be too risky to put this on the main `TestCase`
> 
> IMO freezing time is kind of dirty hack and may blind us from actual issue in the future. A not so good example is pivot model dates, are there risk of asserting the wrong ones, I not sure as not enough experience with tests 😓
> 
> What this PR trying to fix is to assert only attributes we interested with. personally would prefer manually unset the attributes or duplicate the fillable over freeze time, but i did the `invade` way to avoid duplication and lost sync

We're only talking about the current time as determined by `now()`, I don't see any issues. If we do need to consider datetime, then we'd be specifically setting it.

wychoong:
> > > Would it be too risky to put this on the main `TestCase`
> > 
> > 
> > IMO freezing time is kind of dirty hack and may blind us from actual issue in the future. A not so good example is pivot model dates, are there risk of asserting the wrong ones, I not sure as not enough experience with tests 😓
> > What this PR trying to fix is to assert only attributes we interested with. personally would prefer manually unset the attributes or duplicate the fillable over freeze time, but i did the `invade` way to avoid duplication and lost sync
> 
> We're only talking about the current time as determined by `now()`, I don't see any issues. If we do need to consider datetime, then we'd be specifically setting it.

how would you like to proceed with this? if to go with freeze time then can close this

wychoong:
oops just saw your PR, closing this

glennjacobs:
> oops just saw your PR, closing this

Beat me to it.

PR #990, opened 2023-05-14
Fix - set database connection if specified in config

nick-nds:
PR [#235](https://github.com/lunarphp/lunar/pull/235) introduces a new feature to the BaseModel class. This feature allows users to set a custom database connection for the package. By implementing this change, all models within the package will utilize the specified custom database connection if one is provided.

However, I have noticed that when performing database transactions through the Database Manager Facades, it continues to utilize the default connection. In my opinion, it would be beneficial if the Database Manager also utilized the custom connection specified in the configuration file for these transactions. One of the many instances is when Adding/Updating permissions for Staff. 

Thus I have changed the default database connection if specified in the config.

### DIFF
--- a/packages/admin/src/AdminHubServiceProvider.php
+++ b/packages/admin/src/AdminHubServiceProvider.php
@@ -9,6 +9,7 @@ use Illuminate\Support\Facades\Config;
 use Illuminate\Support\Facades\Event;
 use Illuminate\Support\Facades\Gate;
 use Illuminate\Support\Facades\Route;
+use Illuminate\Support\Facades\DB;
 use Illuminate\Support\ServiceProvider;
 use Livewire\Livewire;
 use Lunar\Hub\Auth\Manifest;
@@ -181,6 +182,10 @@ class AdminHubServiceProvider extends ServiceProvider
             $this->app->singleton($tableBuilder, function ($app) use ($tableBuilder) {
                 return new $tableBuilder;
             });
+        }
+
+        if ($connection = config('lunar.database.connection', false)) {
+            DB::setDefaultConnection($connection);
         }
     }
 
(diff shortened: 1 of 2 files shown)


nick-nds:
> Do you mean when you are using DB::xxx() in your own code? If so you need to specify DB::connection('my_connection')->xxx()
> 
> Personally I wouldnt want Lunar to be changing my default database setting.

No, I am referencing the code in Lunar itself. For instance, when adding/updating Staff permissions at `lunarphp/lunar/packages/admin/src/Http/Livewire/Components/Settings/Staff/AbstractStaff.php` when calling function `syncPermissions` we make a `DB::transaction`. This transaction uses a default connection in spite of a connection value provided in lunar database config. So in this case when adding/updating permissions for Staff the code attempts to run query using default connection but the Staff model and StaffPermission model are using different connection.

ryanmitchell:
Then they should be updated to use `::connection()` too - I must have missed those.

nick-nds:
> Then they should be updated to use `::connection()` too - I must have missed those.

I noticed `DB::transaction` is used throughout the codebase and thought specifying `::connection()` to all those places might be tedious and error-prone. So I specified the Database to use the custom connection if exists in ServiceProvider.

glennjacobs:
> > @nick-nds Maybe I'm missing something here, but you're setting Laravel's default DB connection inside of the Lunar package. That surely will affect the entire app, not just Lunar? Seems wrong to me.
> 
> I think you are right. But the issue still exists. Let me look into it more. I need to come up with a different solution. One way is to apply ` ::connection` to all instances of `DB` throughout the package. But that seems tedious. I will look for a different solution and will let you know if I come up with something.

It might be tedious, but it's probably the only way.

We could look to make our own DB facade with the connection already pre-set perhaps?

glennjacobs:
Closing as this approach isn't correct. Please make a new PR when you have a new approach.

PR #995, opened 2023-05-16
Hotfix [0.2] - Fixed visibility of the save search button when $canSaveSearches is s…

DmitryZelonkin:
Steps to reproduce:
1. Open Products page.
2. Type something in the search input and save search button will appear.
3. Open lunarphp/lunar/packages/admin/src/Http/Livewire/Components/Products/Tables/ProductsTable.php and set $canSaveSearches to false.
4. Refresh page and type something in the search input again.

Actual result: the save search button will appear.
Expected result: the search button will not appear.

### DIFF
--- a/utils/livewire-tables/resources/views/index.blade.php
+++ b/utils/livewire-tables/resources/views/index.blade.php
@@ -99,7 +99,7 @@
                             </div>
                         @endif
 
-                        @if ($this->hasSearchApplied)
+                        @if ($this->canSaveSearches)
                             <x-l-tables::button x-on:click="savingSearch = true">
                                 Save Search
                             </x-l-tables::button>


wychoong:
@alecritson i think this is slightly overlapped with what I tried to fix too in #859

DmitryZelonkin:
> @alecritson i think this is slightly overlapped with what I tried to fix too in #859

Yes, but why we need to have the second check for $hasSearchApplied? In my opinion there is no need to check for $hasSearchApplied because we have a property that controls it - $canSaveSearches.

wychoong:
> > @alecritson i think this is slightly overlapped with what I tried to fix too in #859
> 
> Yes, but why we need to have the second check for $hasSearchApplied? In my opinion there is no need to check for $hasSearchApplied because we have a property that controls it - $canSaveSearches.

I believe the original check is to only show the save button when there is search input value.

However only checking on that alone is insufficient/incorrect as you pointed out, hence I have the second check to remain the behavior and also make sure it respect the $canSaveSearches

DmitryZelonkin:
> > > @alecritson i think this is slightly overlapped with what I tried to fix too in #859
> > 
> > 
> > Yes, but why we need to have the second check for $hasSearchApplied? In my opinion there is no need to check for $hasSearchApplied because we have a property that controls it - $canSaveSearches.
> 
> I believe the original check is to only show the save button when there is search input value.
> 
> However only checking on that alone is insufficient/incorrect as you pointed out, hence I have the second check to remain the behavior and also make sure it respect the $canSaveSearches

I get the idea, thank you!

PR #1005, opened 2023-05-18
Hotfix - Ensure $rate is an object when serializing TaxBreakdown

kylekanderson:
There have been times where the `$rate` variable in this instance is an array, rather than an object, thus breaking the property accessors in this serialize method.

This commit ensures that $rate will always be an object, by checking if it is an array, and if so, converting it to an object.

### DIFF
--- a/packages/core/src/Base/Casts/TaxBreakdown.php
+++ b/packages/core/src/Base/Casts/TaxBreakdown.php
@@ -66,7 +66,8 @@ class TaxBreakdown implements CastsAttributes, SerializesCastableAttributes
     public function serialize($model, $key, $value, $attributes)
     {
         return $value->map(function ($rate) {
-            if (is_array($rate)) $rate = (object) $rate;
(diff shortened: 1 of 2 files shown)


glennjacobs:
@alecritson do you think we need a test for this?

packages/core/src/Base/Casts/TaxBreakdown.php
```diff
@@ -66,6 +66,7 @@ public function set($model, $key, $value, $attributes)
     public function serialize($model, $key, $value, $attributes)
     {
         return $value->map(function ($rate) {
+            if (is_array($rate)) $rate = (object) $rate;
```
alecritson, line 69: We don't tend to use the shorthand version, can we switch it for the "usual" syntax, thanks.

alecritson:
> @alecritson do you think we need a test for this?

Yeah I think we should have a test for this.

glennjacobs (approved):

PR #1078, opened 2023-05-24
Fix scout search

wychoong:
for some reason, in my setup scout search only return up to 20 result, thus causing the table pagination never get more than 1 page

- engine: meilisearch:v1.1.0
- `Model->perPage` should be laravel default 15
- meilisearch index setting: `"pagination":{"maxTotalHits":1000}"`

Found the reason for 20 records, meilisearch default value https://www.meilisearch.com/docs/reference/api/search#number-of-results-per-page

this PR changes `Model::search` calls to pass in `hitsPerPage` option to return more result so that eloquent pagination can work properly

### DIFF
--- a/packages/admin/src/Http/Livewire/Components/Brands/BrandsTable.php
+++ b/packages/admin/src/Http/Livewire/Components/Brands/BrandsTable.php
@@ -50,7 +50,11 @@ class BrandsTable extends LunarTable
         return Brand::with(['thumbnail'])
             ->when($this->query, function ($query) {
                 $this->resetPage();
-                $query->whereIn('id', Brand::search($this->query)->keys());
+                $query->whereIn('id', Brand::search($this->query)
+                    ->query(fn ($query) => $query->select('id'))
+                    ->options([
+                        'hitsPerPage' => 1000,
+                    ])->keys());
             })
             ->withCount(['products'])
             ->orderBy('name')
--- a/packages/admin/src/Tables/Builders/CustomersTableBuilder.php
+++ b/packages/admin/src/Tables/Builders/CustomersTableBuilder.php
@@ -21,7 +21,11 @@ class CustomersTableBuilder extends TableBuilder
         $query = Customer::query();
 
         if ($this->searchTerm) {
-            $query->whereIn('id', Customer::search($this->searchTerm)->keys());
+            $query->whereIn('id', Customer::search($this->searchTerm)
+                ->query(fn ($query) => $query->select('id'))
+                ->options([
+                    'hitsPerPage' => 500,
+                ])->keys());
         }
 
         return $query->paginate($this->perPage);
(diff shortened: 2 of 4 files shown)


alecritson:
I'm not near a computer atm but the first thing that springs to mind, is `hitsPerPage` a Meilisearch specific parameter?

wychoong:
> I'm not near a computer atm but the first thing that springs to mind, is `hitsPerPage` a Meilisearch specific parameter?

[Algolia](https://www.algolia.com/doc/api-reference/api-parameters/hitsPerPage/)  have that too but I’m not able to test the before and after change

From scout source code it seems other engines are not using it so no effect on them

On algolia [Doc](https://www.algolia.com/doc/api-reference/api-parameters/hitsPerPage/)  it mentions can set global settings for `hitsPerPage` to change the default: 20, but not meilisearch 

so another approach I can think of is `::search(‘term’)->paginate(1000)->page(1)`
So that the engine will handle their own perPage equivalent

glennjacobs:
I'm going to close this, as I don't think we should have driver-specific settings in the core. Feel free to continue to discuss this here and open another PR if we have a better approach.

wychoong:
@glennjacobs I think I found the root cause to this issue and a fix that is not driver specific

**issue:**
we are not using Scout's pagination but instead return scout result's IDs into eloquent `whereIn` and paginate on eloquent.
https://github.com/lunarphp/lunar/blob/0b54f4d269e15b60b4b4bd12ba9dfef86eb5d112/packages/admin/src/Tables/Builders/ProductsTableBuilder.php#L24-L45
the issue with this is we are always on the first page of the Scout search, and is subject to search engines default per page limit
suppose we should able to set the value on the engine but as mentioned earlier
> On algolia [Doc](https://www.algolia.com/doc/api-reference/api-parameters/hitsPerPage/) it mentions can set global settings for `hitsPerPage` to change the default: 20, but not meilisearch

also there is no such config in Scout.

a potential fix would be to rewrite the `getData` in table builder as seen in the above snippet to use scout's `->paginate($this->perPage)`, but I not sure how well it works with the additional filters and queryExtenders etc, and also the switch to Filament's table builder

for a driver agnostic approach and quick fix, would be similar to this current PR changes but instead of using the `options()` api, I would suggest to make use of scout's `take()` api and let it handle it
```diff
-$query->whereIn('id', Product::search($this->searchTerm)->keys());
+$query->whereIn('id', Product::search($this->searchTerm)
+                ->take(1000) // arbitrary large per page size
+                ->query(fn ($query) => $query->select('id')) // to limit the retrieval
+                ->keys());
```

let me know what do you think

glennjacobs:
@wychoong sounds reasonable, but does this work with Meilisearch and changing it's default response amount? If so, then great.

wychoong:
yes it works fine with Meilisearch, not tested with Algolia but the api is supported in Scout engines

this with Meilisearch


https://github.com/lunarphp/lunar/assets/67364036/b3b5fe73-81af-4bc4-9de8-a631885ed8d3

PR #1079, opened 2023-05-25
Allow disabling of migrations

OscarKolsrud:
This pull request makes a small modification to how the admin hub and the core of Lunar registers migrations with Laravel. We now do a check of the `lunar-hub.database.disable_migrations` config boolean for the admin hub and `lunar.database.disable_migrations` config boolean for the core if we should register these migrations or not. The default option of this config option is `false`, both in the updated config file layouts and as the fallbacks in the ServiceProviders.

Featurewise this now offers Lunar more flexibility for using with third party packages where you may need to publish the migrations to a different path than "standard" and then override. F.ex. when using in a multi tenant setup.


OscarKolsrud:
@glennjacobs Is there a time frame where I could hope this gets merged?

glennjacobs:
This PR requires changelog entries and also documentation for the new configuration option.

PR #1080, opened 2023-05-25
Bug fixes, extra shipping check and doc enhancement

dewebdesigns:
I have fixed several bugs, added oz to the weight measurements, and removed the default sorting feature for the measurements. It is now sorted based on how it's set in the config file. 

I have updated the docs accordingly, and added an option to convert to the imperial system for the developers in the USA. There is rounding issues going from kg to oz, and it's easier to have it set to 1 pound to 16 ounces due to the simplicity of the math, like how metric uses a base 10.

I have also added an extra shipping check to make sure the shipping option is not null on the cart's shipping address with a test.


glennjacobs:
Thank you for the contributions, but we must only have one distinct feature per PR.

Could you please open new PRs with the shipping updates separate?

docs/core/reference/products.md
```diff
@@ -236,6 +237,64 @@ You can configure the available UOM's in the `lunar/shipping.php` config file. H
 - gal
 - floz
 
+### Sorting dropdown measurements in the admin hub
+
+You can adjust the order of the `lunar/shipping.php` config file array to change the order of the dropdown measurements in the admin hub.
+
+### Changing to the imperial system
```
wychoong, line 244: this section is redundant, the previous mention of sorting in config is more than enough

PR #1081, opened 2023-05-26
Add short name to product option

wychoong:
Its nullable thus optional

the only change to UI is when editing the Product Option
image


glennjacobs:
Rather than short name, I feel "Label" would be a better option?

I'm not sure I like the idea of it being optional. It would cause problems in storefronts. 

Maybe we should make it required, and include something to copy the name value across when migrating?

wychoong:
> Rather than short name, I feel "Label" would be a better option?

Ya sounds better.


> Maybe we should make it required, and include something to copy the name value across when migrating?

I'm fine to work on this if thats the preferred direction

glennjacobs:
@wychoong I think I'd like to avoid frontend IF statements checking to see if we have a label, otherwise display the name. 

Part of me thought in the hub UI, maybe the level could auto-populate as you type into the Name field and then if it needs to be different you can change it.

wychoong:
> @wychoong I think I'd like to avoid frontend IF statements checking to see if we have a label, otherwise display the name.

Yup sure, just thinking how to communicate with the dev to use `name` or `label` for the storefront.

> Part of me thought in the hub UI, maybe the level could auto-populate as you type into the Name field and then if it needs to be different you can change it.

Do you think it's necessary to populate on frontend during creation? 
reason being currently we can create new option in Settings and Product page, but both only require the `name`, and `handle` is only updated in Option edit

Setting (dialog)
image

Product (slideover)
image

glennjacobs:
@wychoong if like handle, we auto-generate the Label on creation, they can come back and edit it later if no good?

wychoong:
@glennjacobs I mean, during creation, do you want a Label input field which auto-populate with Name input? or you just need `$option->label = $option->name` during creation

glennjacobs:
> @glennjacobs I mean, during creation, do you want a Label input field which auto-populate with Name input? or you just need `$option->label = $option->name` during creation

The latter.

wychoong:
> > @glennjacobs I mean, during creation, do you want a Label input field which auto-populate with Name input? or you just need `$option->label = $option->name` during creation
> 
> The latter.

sure thing.

alecritson:
@wychoong Have you finished with this and is it ready for us to take a full review?

wychoong:
> @wychoong Have you finished with this and is it ready for us to take a full review?

Yup, ready

glennjacobs:
We have an issue here in that MySQL 5.7 doesn't seem to be able to detect an empty JSON field. This means the state action doesn't populate the new "label" field.

wychoong:
> We have an issue here in that MySQL 5.7 doesn't seem to be able to detect an empty JSON field. This means the state action doesn't populate the new "label" field.

I don't have mysql db at the moment, does changing `where('label', '')` to `whereNull('label')` helps?

glennjacobs:
> > We have an issue here in that MySQL 5.7 doesn't seem to be able to detect an empty JSON field. This means the state action doesn't populate the new "label" field.
> 
> I don't have mysql db at the moment, does changing `where('label', '')` to `whereNull('label')` helps?

Unfortunately not. MySQL 5.7 only gives a proper response when doing `whereRaw('JSON_KEYS(label) IS NULL')`.
And then we've not even considered Postgres. I think maybe we simply check for MySQL as the driver and use that statement instead for the moment.

We're talking about getting GitHub Actions set-up to test in MySQL and Postgres, so that will identify any issues we don't know about.

wychoong:
> > > We have an issue here in that MySQL 5.7 doesn't seem to be able to detect an empty JSON field. This means the state action doesn't populate the new "label" field.
> > 
> > 
> > I don't have mysql db at the moment, does changing `where('label', '')` to `whereNull('label')` helps?
> 
> Unfortunately not. MySQL 5.7 only gives a proper response when doing `whereRaw('JSON_KEYS(label) IS NULL')`. And then we've not even considered Postgres. I think maybe we simply check for MySQL as the driver and use that statement instead for the moment.
> 
> We're talking about getting GitHub Actions set-up to test in MySQL and Postgres, so that will identify any issues we don't know about.

while the idea of having GitHub Actions for DB test is good, i think it won't help much as the we are not writing much test for DB related. I would say handle it case to case basis would be easier.

I used an online mysql 8.0 test site and "`json` is null" works fine, is the issue you mention a 5.7 specific? we should also consider that it's eol is Oct 2023, which is not too far

glennjacobs:
> > > > We have an issue here in that MySQL 5.7 doesn't seem to be able to detect an empty JSON field. This means the state action doesn't populate the new "label" field.
> > > 
> > > 
> > > I don't have mysql db at the moment, does changing `where('label', '')` to `whereNull('label')` helps?
> > 
> > 
> > Unfortunately not. MySQL 5.7 only gives a proper response when doing `whereRaw('JSON_KEYS(label) IS NULL')`. And then we've not even considered Postgres. I think maybe we simply check for MySQL as the driver and use that statement instead for the moment.
> > We're talking about getting GitHub Actions set-up to test in MySQL and Postgres, so that will identify any issues we don't know about.
> 
> while the idea of having GitHub Actions for DB test is good, i think it won't help much as the we are not writing much test for DB related. I would say handle it case to case basis would be easier.
> 
> I used an online mysql 8.0 test site and "`json` is null" works fine, is the issue you mention a 5.7 specific? we should also consider that it's eol is Oct 2023, which is not too far

Yeah, I did mention to Alec that maybe we stop supporting MySQL 5.7, not sure how many it would affect if we changed to MySQL 8.0 only.

wychoong:
> 
> Yeah, I did mention to Alec that maybe we stop supporting MySQL 5.7, not sure how many it would affect if we changed to MySQL 8.0 only.

I see. Maybe for this PR we just add a manual step in upgrade guide to highlight this and get feedback how many are on 5.7, since the field is optional/non-breaking and this state action is just nice to have

wychoong:
> > > Yeah, I did mention to Alec that maybe we stop supporting MySQL 5.7, not sure how many it would affect if we changed to MySQL 8.0 only.
> > 
> > 
> > I see. Maybe for this PR we just add a manual step in upgrade guide to highlight this and get feedback how many are on 5.7, since the field is optional/non-breaking and this state action is just nice to have
> 
> I think with 0.4 we'll simply change the requirements to MySQL 8.0+, as we make strong use of JSON fields 5.7 is holding us back.

yeah, that would be great

glennjacobs:
@wychoong Do you want to update the query to check for either null or a blank string, and then I think we can get it merged.

wychoong:
> @wychoong Do you want to update the query to check for either null or a blank string, and then I think we can get it merged.

sure, will look into that in a bit

glennjacobs:
> > @wychoong Do you want to update the query to check for either null or a blank string, and then I think we can get it merged.
> 
> sure, will look into that in a bit

d/w, I'll put it in now, so I can move on :-)

glennjacobs (approved):

PR #1083, opened 2023-05-26
Feature/add oz to weight

dewebdesigns:
Added oz to the weight array for shipping, and also fixed the bug that had pounds lower than a kg. The conversion was in reverse.


docs/core/reference/products.md
```diff
@@ -236,6 +237,60 @@ You can configure the available UOM's in the `lunar/shipping.php` config file. H
...
+To convert this to the imperial system, you would change it to look like this:
+
+```php
+'weight' => [
+    'oz' => [
+        'format' => '1.00 oz',
+        'unit' => 16.00,
+    ],
+    'lbs' => [
+        'format' => '1.00 lbs',
```
alecritson, line 278: The format appears wrong, unless I'm missing something? Other examples have `1,0.00 kg` see https://cartalyst.com/manual/converter/8.x#formatting

docs/core/reference/products.md
```diff
@@ -236,6 +237,60 @@ You can configure the available UOM's in the `lunar/shipping.php` config file. H
...
+        'format' => '1.00 kg',
+        'unit' => 0.453592,
+    ],
+    'g' => [
+        'format' => '1.00 g',
+        'unit' => 453.592,
+    ],
+],
+```
+This sorts it in the admin hub automatically since you changed the order of the array, so oz will be first followed by lbs.
```
alecritson, line 291: Could we be a bit clearer on "This sorts it" as it might not be immediately obvious for non native english speakers.

PR #1084, opened 2023-05-26
Fix bug that sets the default weight to kg instead of using a length …

dewebdesigns:
Update the default weight unit. mm is a length attribute. Setting to kg to keep it on par with the metric system currently.


alecritson:
Thanks @dewebdesigns Could we get a changelog entry please then we can get this in.

alecritson (approved):

PR #1085, opened 2023-05-26
Sort by the order it's pulled from in the cart/shipping config file

dewebdesigns:
Remove the default sort order so the order of the shipping.php arrays are reflected in the admin hub shipping dropdowns.


glennjacobs:
Requires a changelog entry.

PR #1087, opened 2023-05-26
Allow prices to be stored including tax

glennjacobs:
This enhancement allows store owners to enter prices including tax. It allows for pricing such as £9.99 which was previously impossible to obtain due to rounding from ex. tax pricing.

Currently work in progress.

- [x] Tests
- [x] Documentation updates
- [x] Changelog
- [x] Review cart totals


alecritson:
When editing a product and pricing is set to include tax it still says "(excl. Tax)" when editing 
image

glennjacobs:
> When editing a product and pricing is set to include tax it still says "(excl. Tax)" when editing image

Oh, is this in the hub?

alecritson:
Using the livewire starter kit, I set the prices to include tax and then added an item to the cart, when I came to checkout the sub total for the cart was `$10.99`, which is the price including tax.

image

alecritson:
> > When editing a product and pricing is set to include tax it still says "(excl. Tax)" when editing image
> 
> Oh, is this in the hub?

Yeah it is

alecritson:
> > Using the livewire starter kit, I set the prices to include tax and then added an item to the cart, when I came to checkout the sub total for the cart was `$10.99`, which is the price including tax.
> > image
> 
> Yeah, I didn't know whether to change that or not, as the customer is expecting prices inc. vat now.

Would that mean discounts would need to be considered as well, at least in the UI to state the amounts entered should either be inc or ex tax?

glennjacobs:
> > > Using the livewire starter kit, I set the prices to include tax and then added an item to the cart, when I came to checkout the sub total for the cart was `$10.99`, which is the price including tax.
> > > image
> > 
> > 
> > Yeah, I didn't know whether to change that or not, as the customer is expecting prices inc. vat now.
> 
> Would that mean discounts would need to be considered as well, at least in the UI to state the amounts entered should either be inc or ex tax?

In the hub UI do you mean?

alecritson:
> > > > Using the livewire starter kit, I set the prices to include tax and then added an item to the cart, when I came to checkout the sub total for the cart was `$10.99`, which is the price including tax.
> > > > image
> > > 
> > > 
> > > Yeah, I didn't know whether to change that or not, as the customer is expecting prices inc. vat now.
> > 
> > 
> > Would that mean discounts would need to be considered as well, at least in the UI to state the amounts entered should either be inc or ex tax?
> 
> In the hub UI do you mean?

Yeah

alecritson (approved):

PR #1092, opened 2023-05-31
Spatie Laravel-permission

wychoong:
Replace Lunar permission with `spatie/laravel-permission` 

- [x] Automated tests

Existing Lunar StaffPermissions are migrated over

- [x] Not sure how to handle of dropping `staff_permission` table and model


packages/admin/database/migrations/2023_05_31_100000_create_permission_tables.php
```diff
@@ -0,0 +1,141 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+use Spatie\Permission\PermissionRegistrar;
+
+class CreatePermissionTables extends Migration
```
alecritson, line 8: Is this provided by the Spatie package? A concern would be if developers already have this installed will there be a class name conflict?

packages/admin/src/AdminHubServiceProvider.php
```diff
@@ -519,12 +526,32 @@ protected function registerAuthGuard()
      */
     protected function registerPermissionManifest()
     {
-        Gate::after(function ($user, $ability) {
-            // Are we trying to authorize something within the hub?
+        Gate::after(function (Staff $user, $ability) {
             $permission = $this->app->get(Manifest::class)->getPermissions()->first(fn ($permission) => $permission->handle === $ability);
```
alecritson, line 530: If Spatie is taking over permissions, do we need the manifest anymore?

packages/admin/src/Console/Commands/SyncRolesPermissions.php
```diff
@@ -0,0 +1,55 @@
+<?php
+
+namespace Lunar\Hub\Console\Commands;
+
+use Illuminate\Console\Command;
+use Lunar\Hub\Auth\Manifest;
+use Spatie\Permission\Models\Permission;
+use Spatie\Permission\Models\Role;
+
+class SyncRolesPermissions extends Command
```
alecritson, line 10: What would the purpose of this command be going forward? One concern would be this would just be hanging around in future versions, would it be better on the state class?

packages/admin/database/migrations/2023_05_31_100000_create_permission_tables.php
wychoong: Yes this is from spatie. The conflict can happen but I not sure what would be the better approach

packages/admin/src/Console/Commands/SyncRolesPermissions.php
wychoong: IMO the Manifest is a good api to extend permissions. And with this command we can make the deployment of new permissions easier by just running this, as each permission in spatie package require it to be created upfront.

alecritson: Got it, do you think in that case that the actual logic for doing this should be in it's own action class rather than a command (which can execute that action) so we don't have to litter the codebase with `Artisan::call` ?

packages/admin/database/migrations/2023_05_31_100000_create_permission_tables.php
alecritson: If we're publishing the migration ourselves we could prefix like `LunarCreatePermissionsTables` and check if the table already exists or if the permissions package handles the migrations then do we even need this class?

wychoong: The package requires to publish migration file during installation. So I think we need to include this file, but maybe add a check if the table exist

packages/admin/src/Console/Commands/SyncRolesPermissions.php
wychoong: So we maintain the command but extract it out, and here just execute the action class?

packages/admin/database/migrations/2023_05_31_100000_create_permission_tables.php
wychoong: changed the migration file to
- anonymous class ( avoid class name conflict )
- add table exist check. no table creation if exist

packages/admin/src/AdminHubServiceProvider.php
wychoong: after giving it some thoughts, it should be remain to serve as a scope to only check with permissions registered and managed via the hub. but is possible to just remove it

packages/admin/src/Console/Commands/SyncRolesPermissions.php
wychoong: the logic is extracted to action class

glennjacobs:
> Not sure how to handle of dropping `staff_permission` table and model

@wychoong I've taken a look at your code and you are storing the old permissions as a backup. So can we not simply add a migration to remove the old table?

glennjacobs:
@wychoong I've converted to draft for the moment, as we need to decide what we're doing with the old permissions table and there's some menu related tests which are failing.

wychoong:
> @wychoong I've converted to draft for the moment, as we need to decide what we're doing with the old permissions table and there's some menu related tests which are failing.

anything you need from my side?

glennjacobs:
> > @wychoong I've converted to draft for the moment, as we need to decide what we're doing with the old permissions table and there's some menu related tests which are failing.
> 
> anything you need from my side?

If you can fix the tests and add a migration to remove the old permissions table, that should do it.

packages/admin/database/migrations/2023_05_31_110000_drop_staff_permissions_table.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+use Illuminate\Support\Facades\Schema;
+use Lunar\Base\Migration;
+
+class CreateStaffPermissionsTable extends Migration
```
glennjacobs: This should probably be called "Remove" not "Create"

wychoong: done

alecritson:
@wychoong I tried upgrading to this on the starter kit, I did the following:

```bash
php artisan migrate
```

I logged into the hub and created a staff member, attempted to assign permissions and got:

image


So I checked the database and there was nothing populated, then I looked in the PR and ran:

```sh
php artisan lunar:hub:permissions
```

Which seemed to have no effect and still have the same error.

Could it be a case that the upgrade guide needs improving or is something going wrong?

wychoong:
> @wychoong I tried upgrading to this on the starter kit, I did the following:
> 
> ```shell
> php artisan migrate
> ```
> 
> I logged into the hub and created a staff member, attempted to assign permissions and got:
> 
> image
> So I checked the database and there was nothing populated, then I looked in the PR and ran:
> 
> ```shell
> php artisan lunar:hub:permissions
> ```
> 
> Which seemed to have no effect and still have the same error.
> 
> Could it be a case that the upgrade guide needs improving or is something going wrong?

silly mistake.
it should be handled by `lunar:hub:install` after fresh installation

alecritson (approved):

PR #1093, opened 2023-05-31
Fix menu permissions

wychoong:
Currently `MenuItem` with permission is not hidden from sidebar/menu when staff have no permission to access which results to getting 403 on clicking.

This PR looks to fix:
- add `gate()` to menu registration
- hide `Group/Section` label when no Item is accessible
- Setting menu now redirect to first accessible route instead of always `hub.attributes.index`
- fix Staff permission when remove `Settings` perm remove other permissions
    - eg: Enable `Settings` and `Manage Orders`, save, refresh page permissions still correct, Disable `Settings`, save and refresh, `Manage Orders` permission removed as well

*****

- [x] Documentation updates (n/a)
- [x] Automated tests


packages/admin/src/Menu/MenuSlot.php
```diff
@@ -108,17 +122,55 @@ public function getItems(): Collection
...
      */
     public function getGroups()
     {
-        return $this->groups;
+        return $this->groups->filter(function ($group) {
+            return ! $group->gate || Auth::user()->can($group->gate);
+        });
+    }
+
+    public function hasLinks(): bool
```
alecritson, line 142: Do you think we can refactor a bit to reduce this?

alecritson (commented):
Thanks @wychoong Will have a pull down and play around this but looks promising!

packages/admin/src/Menu/MenuSlot.php
wychoong: I don't think there is a meaningful way to refactor these as `items`, `groups` and `sections` are stored separately properties

packages/admin/src/Menu/MenuSlot.php
```diff
@@ -108,17 +122,55 @@ public function getItems(): Collection
...
+        }
+
+        if ($this->getSections()->filter(fn ($section) => $section->hasLinks())->count()) {
+            return true;
+        }
+
+        return false;
+    }
+
+    public function getFirstLink()
```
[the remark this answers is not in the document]
wychoong: same

packages/admin/src/Menu/MenuSlot.php
wychoong: I'm open to work on if you can give me some pointer to start with

PR #1096, opened 2023-06-06
Fix livewire middleware persistency 

wychoong:
Authenticate middleware doesn't persist after initial request

https://laravel-livewire.com/docs/2.x/security#persistent-middleware

Not sure what is the easy way to replicate this, I have this route register in the hub

```php
 Route::middleware([
    'web',
    Lunar\Hub\Http\Middleware\Authenticate::class,
])->group(function () {
    Route::middleware(['can:permission'])
      ->group(function () {
              Route::get('route', MyLivewireComponent::class);
      });
});
```

after the initial request, it doesn't resolve with the correct `staff` guard and causes `Gate::after(function ($user, $ability)` to receive a `App\Models\User` instead of `Lunar\Models\Staff`

- [] Automated tests
- [x] Changelog entries (hub)


alecritson:
Thanks @wychoong Could we get a quick changelog entry so it's clear what's changing.

wychoong:
> Thanks @wychoong Could we get a quick changelog entry so it's clear what's changing.

done

glennjacobs (approved):

PR #1097, opened 2023-06-08
Feat - Refactor Order creation

alecritson:
This PR aims to refactor and improve Orders and how they relate to Carts. Currently a Cart can only have 1 order associated via the `order_id` on the `carts` table, which was fine for most cases but it's a bit restrictive, also the `CreateOrder` action has all the logic within, which means if you want to customise any aspect of this, you have to replace the action entirely, which isn't ideal if you only want to do a small tweak...

The changes listed below will likely be breaking changes so this will likely go in the next `0.X` release.

## Changes to the Cart > Order relationship.

Currently a cart has an `order_id` column, this will change so that the `order` table now has the `cart_id` column, this will then also introduce the concept of "draft" and "complete" orders.

### Draft Orders

These are orders which have been created from a cart that are yet to be considered "paid" for.

### Complete Orders

These are orders which are considered "paid"

The logic which determines whether an order is complete is the presence of a date in the `placed_at` column.

### Accessing the relationship

Previously.

```php
$cart->order
```

```php
// This is relationship which will accept an ID of the draft you want to fetch.
$cart->draftOrder
$cart->draftOrder(1);
```

```php
// Return all completed orders.
$cart->completedOrders

// Return the first completed order.
$cart->completedOrder

// Return the completed order by it's id
$cart->completedOrder(1);
```

It's also possible to check whether the cart has completed orders with a little helper

```php
$cart->hasCompletedOrders();
```

## CreateOrder action changes.

The biggest change to this Action is that currently it's very much "create an order every time you call me", which is not very useful, so now the action will handle whether there is an existing order we can just simply update. **Note, this action will specifically look for a draft order to update, completed orders will not be touched**. The process is as follows:

- Find a **draft** order if given an ID
- If no ID is passed or not found, return an instance of a new order to create.
- If the cart already has completed order associated, check whether to allow for multiple orders.
- If multiple orders are not allowed, throw an `DisallowMultipleCartOrdersException`
- If the above passes, carry on.

As mentioned at the start, the logic for this is currently all within the `CreateOrder` action class, which is not in keeping with the direction we want to go in terms of offering flexibility for developers. This has now be refactored into pipelines, much like when calculating the cart totals.

These pipelines are stored in `config/lunar/orders.php` and the default values will be:

```php
'pipelines' => [
    'creation' => [
        Lunar\Pipelines\Order\Creation\FillOrderFromCart::class,
        Lunar\Pipelines\Order\Creation\CreateOrderLines::class,
        Lunar\Pipelines\Order\Creation\CreateOrderAddresses::class,
        Lunar\Pipelines\Order\Creation\CreateShippingLine::class,
        Lunar\Pipelines\Order\Creation\CleanUpOrderLines::class,
        Lunar\Pipelines\Order\Creation\MapDiscountBreakdown::class,
    ],
],
```

An example of what these classes could look like is:

```php
<?php

namespace App\Pipelines\Orders\Creation;

use Closure;
use Lunar\Models\Order;
use Lunar\Models\OrderAddress;

class AcmeOrderPipeline
{
    /**
     * @return Closure
     */
    public function handle(Order $order, Closure $next)
    {
        // ...
        return $next($order);
    }
}

```

In addition to this, there are some changes to the `CreateOrder` action, based on the draft/complete concept above.

Current execute method:

```php
public function execute(
    Cart $cart
)
```

New execute method:

```php
public function execute(
    Cart $cart,
    int $orderIdToUpdate = null
): self
```
- `$allowMultipleOrders` - Determine whether you want to allow another draft order to be created, if there is already a complete order associated to this cart.
- `$orderIdToUpdate` - Pass through the ID of the order you want to update.

### Changes to actions `then()` callback.

Previously, if you was to call the action and use a callback in the `then` method like so:

```php
(new CreateOrder)->execute(...)->then(function () {
  // ...
});
```

You wouldn't have access to the result of the action, which was a bit pointless, now you have access to whatever the result was, like so:

```php
(new CreateOrder)->execute(...)->then(function ($order) {
  // ...
});
```

### Added `orderId` to the `PaymentAuthorize` DTO

Old:

```php
public function __construct(
    public bool $success = false,
    public ?string $message = null
) {
```

New:

```php
public function __construct(
    public bool $success = false,
    public ?string $message = null,
    public ?int $orderId = null
) {
```

Notice we don't actually pass the Order model, this is because orders could potentially have many many lines which could cause problems when serialised and by the time the data gets to the job queue etc it could potentially be outdated, so it's up to your handlers to fetch the order to ensure it's the most up to date version.

### Added `PaymentAttemptEvent`

There is now a new event, which will be used by add ons (like Stripe) to dispatch when payment is attempted, emphasis on the word "attempted", this will trigger whether successful or not and an instance of `Lunar\Base\DataTransferObjects\PaymentAuthorize` will be passed through to the event so you can check logic accordingly like so:

```php
final public function handle(PaymentAttemptEvent $event)
{
    $paymentAuthorize = $event->paymentAuthorize;

    if ($paymentAuthorize->success) {
        $order = Order::find($paymentAuthorize->orderId);

        Mail::to($order->billingAddress->contact_email)->queue(
            new OrderConfirmationMail($order)
        );
    }
}
```

---

The `order_id` column is deprecated on `carts` for ease of migration/upgrade and figured it's better left to the next update to remove entirely.


packages/core/src/Models/Cart.php
```diff
@@ -235,21 +235,70 @@ public function billingAddress()
...
+                $query->where('id', $completedOrderId);
+            })->whereNotNull('placed_at');
+    }
+
+    /**
+     * Return the carts completed order.
+     *
+     * @return \Illuminate\Database\Eloquent\Relations\HasMany
+     */
+    public function completedOrders()
```
wychoong, line 288: should have `draftOrders` too?

packages/core/database/state/MigrateCartOrderRelationship.php
```diff
@@ -0,0 +1,41 @@
...
+
+    public function run()
+    {
+        $prefix = config('lunar.database.table_prefix');
+
+        if (! $this->canRun()) {
+            return;
+        }
+
+        DB::transaction(function () use ($prefix) {
```
wychoong, line 23: a concern here for store that is in productions and have lots of carts and orders
- maybe some pagination/chunking and have the `DB::transaction` within
- after copied the `cart->id` to `order`, set `cart.order_id` to null
- should we explore another way without using too much php loop like something 
```sql
-- using query builder to form
update orders set cart_id = (select id from carts where order_id = order.id) 
where exists (select 1 from carts where order.id = cart.order_id where cart.order_id is not null)

-- update cart order_id to null....
```

packages/core/src/Actions/AbstractAction.php
```diff
@@ -6,8 +6,13 @@
...
 {
+    /**
+     * The property to pass back to the callback
+     */
+    protected mixed $passThrough = null;
+
     public function then(Closure $callback)
     {
-        return app()->call($callback);
+        return $callback($this->passThrough);
```
wychoong, line 16: does this change means we can't use dependencies injection anymore in the callback?
maybe
```php
app()->call($callback, [
    'order' => $this->passThrough,
]);
```
the drawback is the `->then(fn($order)=> ..)` parameter must be `$order`

packages/core/src/Actions/Carts/CreateOrder.php
```diff
@@ -2,154 +2,43 @@
...
-                    })->values(),
-                    'tax_total' => $line->taxAmount->value,
-                    'total' => $line->total->value,
-                    'notes' => null,
-                    'meta' => $line->meta,
-                ];
-            });
-
-            $addresses = collect();
+    final public function execute(
```
wychoong, line 18: looking at `execute` is the only method in the action class, and customisation should be on the pipelines, should the `final` be at the class instead? and since the action can be swapped out from config, is it even necessary to be final 😅

wychoong:
just curious, in what use case the multiple draft and completed orders would be useful?

packages/core/src/Models/Cart.php
alecritson: You should be only be setting one draft order I believe.

wychoong: then the draft relationship shouldn't need orderId?

packages/core/database/state/MigrateCartOrderRelationship.php
alecritson: Thanks @wychoong I think the idea of using a single statement is good.

packages/core/src/Actions/AbstractAction.php
alecritson: Yeah although I'm open to that if people think the ability to DI on the callback is important.

packages/core/src/Actions/Carts/CreateOrder.php
alecritson: I've moved final on to the class, I think this is fine because if people want to swap the class then that's fine but they then shouldn't then extend the class, if you're replacing it then it should be your own implementation, otherwise use the pipelines.

packages/core/src/Models/Cart.php
glennjacobs: > then the draft relationship shouldn't need orderId?

It uses Order ID as an optional check to ensure you're getting the order you expect.

glennjacobs:
> just curious, in what use case the multiple draft and completed orders would be useful?

We don't want multiple draft orders, but there could be an occasion when someone might want to create multiple orders from the same cart. It's also a fail-safe so if a customer does somehow manage to create two orders from a single cart, maybe because of a coding mistake, we are still able to keep the two orders related to the cart, whereas currently if that happened you would only be aware of the one order on the cart model and the other would be orphaned.

glennjacobs:
@alecritson the order ID is optional isn't it?

> Find a draft order if given an ID

glennjacobs:
@alecritson should it not be...

- Find a draft order (optionally passing an order ID)
- If no draft order is found, return an instance of a new order to create.
- If the cart already has completed order associated, check whether to allow for multiple orders.
- If multiple orders are not allowed, throw an DisallowMultipleCartOrdersException
- If the above passes, carry on.

packages/core/src/Models/Cart.php
```diff
@@ -500,7 +549,7 @@ public function createOrder(): Order
         return app(
             config('lunar.cart.actions.order_create', CreateOrder::class)
```
glennjacobs, line 550: The CreateOrder action now accepts 2 additional parameters, but these aren't being used here?

alecritson:
> @alecritson should it not be...
> 
> * Find a draft order (optionally passing an order ID)
> * If no draft order is found, return an instance of a new order to create.
> * If the cart already has completed order associated, check whether to allow for multiple orders.
> * If multiple orders are not allowed, throw an DisallowMultipleCartOrdersException
> * If the above passes, carry on.

Am I missing something, is that not what it's doing?

alecritson:
> @alecritson the order ID is optional isn't it?
> 
> > Find a draft order if given an ID

Is it optional yup

glennjacobs:
> > @alecritson should it not be...
> > 
> > * Find a draft order (optionally passing an order ID)
> > * If no draft order is found, return an instance of a new order to create.
> > * If the cart already has completed order associated, check whether to allow for multiple orders.
> > * If multiple orders are not allowed, throw an DisallowMultipleCartOrdersException
> > * If the above passes, carry on.
> 
> Am I missing something, is that not what it's doing?

I believe so, just your comment suggested it wasn't. I think we're all good.

alecritson:
> @alecritson do we need to document the updated `$cart->createOrder()` function?

Changed up the docs a bit to reflect the changes

docs/core/reference/orders.md
```diff
@@ -2,36 +2,37 @@
 
 ## Overview
 
-As you'd expect, orders on an online system show what users have purchased. They are linked to a Cart and you can only have 1 order per cart in the database.
+As you'd expect, orders on an online system show what users have purchased. They are linked to a Cart and you can only
+have 1 order per cart in the database.
```
glennjacobs, line 6: This isn't true, you can now have multiple orders per cart.

docs/core/reference/orders.md
```diff
@@ -2,36 +2,37 @@
...
-|exchange_rate| The exchange rate between `currency_code` and `compare_currency_code`
-|placed_at|The datetime the order was considered placed.
-|meta|Any additional meta info you wish to store
-|created_at||
-|updated_at||
+| Field                 | Description                                                                                                        |
+|:----------------------|:-------------------------------------------------------------------------------------------------------------------|
+| id                    |                                                                                                                    |
+| user_id               | If this is not a guest order, this will have the users id                                                          |
+| customer_id           | Can be `null`, stores customer                                                                                     |
```
glennjacobs, line 16: Shouldn't `cart_id` be in this list now?

glennjacobs:
> question: how is the `$allowMultipleOrders` and `$orderIdToUpdate` being set in the PaymentType taking the default OfflinePayment as example
> 
> https://github.com/lunarphp/lunar/blob/472a4a6df62a72271c60ffdf5c5f37c013c507f3/packages/core/src/PaymentTypes/OfflinePayment.php#L19
> 
> 
> as typically `$cart->createOrder` is being called inside PaymentType, as PaymentType integration can be from addons, so how user can control? would the `allowMultiple` more suitable to be a config? as unlikely a store would have different behaviour for this action on different location

We've had a long discussion on this and 99 times out of 100 multiple orders won't be required. If someone does need it, they can extend or make their own payment driver. Config didn't seem right, as it would really want to be per-payment driver.

glennjacobs (approved):

PR #1101, opened 2023-06-19
Bugfix: Warning message on capture view should take currency's decimal places into account

webcraft:
When you work with a currency with decimal_places != 2, the warning message ("You're about to capture an amount less than the total transaction value.") will be shown when it's not needed. The multiplier is hard coded to 100, but this should be based on the currency's decimal_places imho.

Similar problem occurs in the refund modal (incorrect amount shown), this is a bit more involved; and I've not yet found the time to create a PR. Will try to do this after feedback on this PR.


glennjacobs:
Looks good, just needs a changelog entry

webcraft:
@glennjacobs Done, thanks!

glennjacobs (approved):

PR #1102, opened 2023-06-20
Fix deprecation warnings due to passing null

ryanmitchell:
We are getting a lot of depreciation warnings on discounts due to passing null to strtoupper and bcmul, instead of the types they expect.

This PR sets the type they are expecting.


alecritson (approved):
Thanks @ryanmitchell for sorting.

PR #1104, opened 2023-06-20
Cache relations to avoid duplicate queries

ryanmitchell:
You'll probably hate this approach, but I prefer to bring solutions than issues so it at least should give you something to think about.

We're getting a load of duplicate queries on our live 0.3 sites, which is obviously something we're keen to avoid! 

The approach is to use the attribute accessor methods to `Blink` the first call to the relation, and intentionally not make it specific to the model, so that other models dont need to do the same query. Where this is really useful is for relations like tax classes, channels and other models that are referenced by many other models.


glennjacobs:
I'm not keen on this. I think performance can be achieved in other ways, such as eager loading or caching models using observers, so going to close it. 

If you feel there's still a problem to address, then happy to discuss it with examples.

PR #1105, opened 2023-06-20
Change index searchable to override key value

wychoong:
currently the `isset` checking avoid multi assignment of the same key but this is unnecessary as it can cause other undesired scenario.

for example:
```php
// i have this observer setup and is working
    {
        $channels = $product->channels()
            ->select('handle', 'enabled')
            ->pluck('enabled', 'handle')
            ->toArray();

        foreach ($channels as $channel => $enabled) {
            $product->addSearchableAttribute('channel_' . $channel, (int) $enabled);
        }
    }
```

``` php
// and i have this to create product

$product = Product::create([
    'attribute_data' => $attributeData,
    'product_type_id' => $productType->id,
    'status' => 'published',
    'brand_id' => $brand,
]);

$product->scheduleChannel($channels);

$product->searchable();
```

the issue here is when `Product` is created/saved it is retrieving the not-exist/old `Channel`
and the `isset` is preventing `->searchable()` to update the `additionalSearchFields` array

removing the `isset` won't break anything and is letting the array getting latest value if its called multiple times

- [] Automated tests
- [x] Changelog entries (core)


alecritson:
Thanks @wychoong Happy to have this in, can you please add a quick change-log entry? Thanks.

wychoong:
> Thanks @wychoong Happy to have this in, can you please add a quick change-log entry? Thanks.

done

glennjacobs (approved):

PR #1108, opened 2023-06-22
Improve customer group and channel scope empty checks

alecritson:
Currently the `channel` and `customerGroup` scopes only check if the passed values are not `null`, which doesn't account for whether empty arrays or collections are passed (which are accepted parameters).

The result currently would be if you passed an empty array/collection you would get nothing back.

This PR looks to add additional checks which consider those additional types.


glennjacobs (changes requested):
Test coverage needed

packages/core/src/Base/Traits/HasChannels.php
```diff
@@ -83,29 +83,28 @@ public function activeChannels()
...
-     * @param Channel|iterable $channel
-     *
+     * @param  Builder  $query
+     * @param  Channel|iterable  $channel
      * @return Builder
      */
     public function scopeChannel($query, Channel|iterable $channel = null, DateTime $startsAt = null, DateTime $endsAt = null)
     {
-        if (!$channel) {
+        if (empty($channel) || (is_a($channel, Collection::class) && $channel->isEmpty())) {
```
wychoong, line 99: i think Laravel `blank` helper function is cleaner here
https://github.com/laravel/framework/blob/b4ab62986000eaa411a22dbeb112b21c10420fdc/src/Illuminate/Support/helpers.php#L42-L62

packages/core/src/Base/Traits/HasCustomerGroups.php
```diff
@@ -73,38 +67,35 @@ protected function validateScheduling(Collection $models)
...
-     * @param CustomerGroup|string $customerGroup
-     *
+     * @param  Builder  $query
+     * @param  CustomerGroup|string  $customerGroup
      * @return Builder
      */
     public function scopeCustomerGroup($query, CustomerGroup|iterable $customerGroup = null, DateTime $startsAt = null, DateTime $endsAt = null)
     {
-        if (!$customerGroup) {
+        if (empty($customerGroup) || (is_a($customerGroup, Collection::class) && $customerGroup->isEmpty())) {
```
wychoong, line 98: same

alecritson: Good shout!

glennjacobs (approved):

glennjacobs (approved):

Issue #1103, opened 2023-06-20
Discount performance


ryanmitchell:
- Lunar version: 0.3.0
- Laravel Version: 10.13.5
- PHP Version: 8.1
- Database Driver & Version: MySQL 8

### Expected Behaviour:
Do you have any plans to improve discount performance? At the moment it's structured that all active discounts in the appropriate channels and groups are loaded in every page load, which is resulting in over 1000 extra database queries for us. As you'd expect that comes with a performance hit.

Could the query be optimised to only return those that have no coupon required or where the coupon matches? 

Another optimisation would be to provide a way of those only being loaded in on certain pages - eg we don't display the cart value on every page, so we dont actually need to load in discounts anywhere except for the basket and checkout pages.

alecritson:
Hi @ryanmitchell I've started a draft PR which I'm hoping will address these performance issues, I tested on one of our client sites and it did indeed take the number of queries down quite considerably. I'd be interesting to know your thoughts.

ryanmitchell:
Thanks @alecritson. This has helped a huge amount! Back down to a reasonable number of queries again.
There are a few duplicate queries coming from shipping tables, but I can pick those up with you separately.

alecritson:
@ryanmitchell This has been merged into `main` and will be in `0.4`

PR #1109, opened 2023-06-22
Discount query improvements

alecritson:
Based on #1103 this PR looks to optimise the query when fetching discounts. Beforehand we just loaded all available discounts, regardless of whether they were actually needed and then left it up to the individual types to figure out whether they should apply, this will be quite taxing on sites with large amounts of discounts. This PR looks to change the following:

- Removed the custom logic for checking customer groups / channels in favour of the recently added scopes.
- Added `usable` scope to discounts so we can query based on whether `uses` has exceeded `max_uses` if present.
- Added `products` scope to discounts to query based on whether the discount is limited to certain product ids.
- The `Discounts::getDiscounts()` method now accepts a nullable `Cart` model to be passed.
- If the cart exists, we only fetch discounts which have the products in the cart is conditions or no conditions at all.
- If the cart has a coupon, we only fetch discounts which match that coupon, or if the discount requires no coupon.

This PR relies on the tweak on #1108


glennjacobs:
@alecritson looks good, much tidier.

glennjacobs (approved):

PR #1008, opened 2023-05-20
Make cart aware of customer

wychoong:
- [x] Documentation updates
- [x] Automated tests
- [x] Changelog entries (core and hub)

PR #1125, opened 2023-06-23
Make cart aware of customer

wychoong:
replace #1008


packages/admin/CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Added Label field when editing product options.
 - Ability to add and sort child collections when editing a collection.
+- Added `customer` relationship to Cart model
```
glennjacobs, line 13: Nothing has changed in the admin package, I don't think these are needed?

packages/core/src/Managers/StorefrontSessionManager.php
```diff
@@ -120,6 +130,44 @@ public function initChannel()
...
+            return $this->customer;
+        }
+
+        $customer_id = $this->sessionManager->get(
+            $this->getSessionKey().'_customer'
+        );
+
+        if (! $customer_id) {
+            if (Auth::check() && is_lunar_user(Auth::user())) {
+                $user = Auth::user();
```
alecritson, line 148: One thing that comes to mind is by `Auth::user()` we are locking the user association to the authenticated user, I don't think we should assume that's what storefronts want to do. Maybe instead allow the user to be set and then check against that? It would be relatively simple to set the user on the storefront using middleware.

It would also mean in the `setUser` method we could check it's a lunar user and we don't have to worry about it ever again.

glennjacobs: The Cart Session Manager uses the auth'd user, so I don't see why this can't also.

However, should it be changed to work the same as the Cart Session Manager? https://github.com/lunarphp/lunar/blob/0.3/packages/core/src/Managers/CartSessionManager.php

```
public function __construct(
        protected SessionManager $sessionManager,
        protected AuthManager $authManager,    // <<<<<< THIS
        protected $channel = null,
        protected $currency = null,
        public $cart = null
    ) {
        //
    }
```

wychoong: made the change to use `AuthManager`

packages/core/src/Managers/StorefrontSessionManager.php
```diff
@@ -120,6 +131,44 @@ public function initChannel()
...
+
+        $customer_id = $this->sessionManager->get(
+            $this->getSessionKey().'_customer'
+        );
+
+        if (! $customer_id) {
+            if ($this->authManager->check() && is_lunar_user($this->authManager->user())) {
+                $user = $this->authManager->user();
+
+                if ($customer = $user->customers()->orderBy('created_at', 'desc')->orderBy('id', 'desc')->first()) {
```
alecritson, line 151: Should this be in a scope? Could be useful and explain itself better I think.

packages/core/src/Managers/StorefrontSessionManager.php
```diff
@@ -142,6 +191,46 @@ public function setChannel(Channel|string $channel): self
...
+            $this->getSessionKey().'_customer',
+            $customer->id
+        );
+
+        if (
+            $this->authManager->check()
+            && is_lunar_user($this->authManager->user())
+            && ! $this->customerBelongsToUser($customer)
+        ) {
+            return $this;
```
alecritson, line 218: I feel like an exception would be better here so it's clear to the developer why the customer they want to set, can't be.

packages/core/src/Managers/StorefrontSessionManager.php
wychoong: done

packages/core/src/Managers/StorefrontSessionManager.php
wychoong: done

PR #1074, opened 2023-05-24
Optimize shipping modifiers

wychoong:
replace #899

- [x] Documentation updates
- [x] Automated tests
- [x] Changelog entries (core and hub)

PR #1126, opened 2023-06-23
Optimize shipping modifiers

wychoong:
replace #1074


alecritson:
Thanks @wychoong I think because we're changing an interface we'll need some upgrade guides added for this.

wychoong:
> Thanks @wychoong I think because we're changing an interface we'll need some upgrade guides added for this.

done

glennjacobs (approved):

PR #1128, opened 2023-06-23
Change pagination on AbstractProductType for availableAttributes

yanielbf:
I detected that when there are many attributes, all of these are not displayed in the Product type. This is happening because the AbstractProductType on line 210 is paging up to 25 items

https://discord.com/channels/497326121441034240/765213029163335730/1117888171871572148

image

With 25:
image

With 250
image

### DIFF
--- a/packages/admin/src/Http/Livewire/Components/Products/ProductTypes/AbstractProductType.php
+++ b/packages/admin/src/Http/Livewire/Components/Products/ProductTypes/AbstractProductType.php
@@ -207,6 +207,6 @@ abstract class AbstractProductType extends Component
                 fn ($query, $search) => $query->where("name->{$this->defaultLanguage->code}", 'LIKE', '%'.$search.'%')
             )->whereSystem(false)
             ->whereNotIn('id', $existing->pluck('id')->toArray())
-            ->paginate(25);
+            ->paginate(250);
     }
 }


wychoong:
IMO this is not the right solution, as we are assuming the max will be 250, might as well don't paginate at all. and the UI doesn't looks good with so many attributes in 1 go.

should have the pagination links at the bottom to navigate and maybe a search field to filter

alecritson:
Going to convert this to draft for the time being, although I believe it's an issue that needs solving, i'm not sure this is the approach as @wychoong said.

packages/admin/src/Http/Livewire/Components/Products/ProductTypes/AbstractProductType.php
```diff
@@ -207,6 +207,6 @@ public function getAvailableAttributes($type)
                 fn ($query, $search) => $query->where("name->{$this->defaultLanguage->code}", 'LIKE', '%'.$search.'%')
             )->whereSystem(false)
             ->whereNotIn('id', $existing->pluck('id')->toArray())
-            ->paginate(25);
+            ->paginate(250);
```
glennjacobs, line 210: Should we simply just change this to `->get()` for the moment?

glennjacobs:
Please can you target `main` for this PR? thanks

alecritson:
Appreciate your work on this @yanielbf I tried to help move this along by rebasing to main but `0.3` is quite far behind `main` and since it would have been a simple change, the effort to rebase wouldn't be worth it I don't think. So for that reason there is a new PR #1208 just to get things going.

PR #1130, opened 2023-06-27
Remove type declaration for Staff

### LINKED ISSUE #1129: Lunar\Hub\AdminHubServiceProvider::Lunar\Hub\{closure}(): Argument #1 ($user) must be of type Lunar\Hub\Models\Staff
- Lunar version: 0.3.1

### Expected Behaviour:
The change made here is a breaking change:
https://github.com/lunarphp/lunar/commit/4c4e429abbe2781160a14c7086393fa42918dda6

By adding Staff $model to the Gate::after it has caused Statamic's control panel on our installs to no longer work (they error out). I'd imagine it also conflicts with other Gate::afters() from other packages.

Can you check if the request is inside the Lunar hub before applying the Gate?

alecritson:
Closes #1129

### DIFF
--- a/packages/admin/src/AdminHubServiceProvider.php
+++ b/packages/admin/src/AdminHubServiceProvider.php
@@ -3,7 +3,6 @@
 namespace Lunar\Hub;
 
 use Illuminate\Routing\Events\RouteMatched;
-use Illuminate\Support\Facades\Auth;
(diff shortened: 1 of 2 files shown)


glennjacobs (approved):

wychoong:
Should this be enhanced to be in Lunar scope only or is that not possible?

PR #1132, opened 2023-07-03
Stock Manager

glennjacobs:
This feature adds stock management to Lunar. It comes with two drivers; Simple + Null.

The Simple stock driver will simply decrement the stock+backorder values on ProductVariants. The Null driver will completely ignore stock requests.

You can make your own drivers if you want to integrate with other inventory systems.

TODO:

- [x] Config for reservation duration
- [x] Reserve stock from a Cart model
- [x] Return stock information for display
- [ ] Job to purge expired reservations
- [ ] Move reserved stock to Order when placed from a Cart
- [ ] Dispatch on Orders
- [ ] Dispatch Orders within the hub
- [ ] Config to auto-dispatch when an Order is "placed"
- [ ] Config to disable editing of stock figures in the hub
- [ ] Documentation
- [ ] Changelog


packages/core/src/Drivers/SimpleStockDriver.php
```diff
@@ -0,0 +1,153 @@
...
+
+        $reservedCount = StockReservation::where('variant_id', '=', $purchasable->id)
+            ->where('expires_at', '>', now())
+            ->sum('quantity');
+
+        $stockInfo = new StockInfo($purchasable->stock, $purchasable->backorder);
+
+        $stockInfo->stock -= $reservedCount;
+
+        if ($stockInfo->stock < 0) {
```
alecritson, line 33: Will stock ever go into negative numbers? And should it be `<=` or just `!$stockInfo->stock` ?

packages/core/src/Drivers/SimpleStockDriver.php
```diff
@@ -0,0 +1,153 @@
...
+    /**
+     * Get the available stock information, after deducting reservations.
+     */
+    public function availableStock(Purchasable $purchasable): StockInfo
+    {
+        // SimpleStock driver only supports ProductVariants
+        $this->checkIsVariant($purchasable);
+
+        $reservedCount = StockReservation::where('variant_id', '=', $purchasable->id)
+            ->where('expires_at', '>', now())
```
alecritson, line 26: Should date be passable? Using `now()` here if we're in a loop or long running process each instance of `now()` could potentially be different.

Also means we can get available stock for a given date if needed.

packages/core/src/Drivers/SimpleStockDriver.php
```diff
@@ -0,0 +1,153 @@
...
+            ->whereNot(function (Builder $query) use ($line) {
+                $query->where('stockable_type', '=', $line::class)
+                    ->where('stockable_id', '=', $line->id);
+            })
+            ->sum('quantity');
+
+        $totalAvailable = $line->purchasable->stock + $line->purchasable->backorder - $reservedCount;
+
+        return $totalAvailable >= $line->quantity;
```
alecritson, line 65: Should this return a DTO like the other methods, just thinking if there is no stock but it can be back ordered due to `purchasable` being `always` you may have some available but also some that need to be backordered?

packages/core/src/Drivers/SimpleStockDriver.php
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    /**
+     * Reserve stock for the cart/order line.
+     */
+    public function reserveStock(ReservesStock $line, string $location = null): bool
+    {
+        // Check if we have enough stock available to reserve
+        if (! $this->checkStock($line)) {
```
alecritson, line 74: Again, if the item can be backordered should we not be allowing the maximum available to be reserved but then have the overflow going into backorder? Maybe they allow split delivery

packages/core/src/Drivers/SimpleStockDriver.php
```diff
@@ -0,0 +1,153 @@
...
+
+        if (! $reservation) {
+            return true;
+        }
+
+        if (! $quantity) {
+            $quantity = $reservation->quantity;
+        }
+
+        if ($reservation->quantity > $quantity) {
```
alecritson, line 106: Am I correct in thinking if the quantity you want to release is more than or equal to the available quantity on the reservation, nothing happens aside from the expiry being set?

packages/core/src/Drivers/SimpleStockDriver.php
```diff
@@ -0,0 +1,153 @@
...
+     */
+    public function dispatchStock(ReservesStock $line, int $quantity = null): bool
+    {
+        // Remove/expire the reservation and decrement from stock figure
+
+        if (! $quantity) {
+            $quantity = $line->quantity;
+        }
+
+        if (! $this->releaseStock($line, $quantity)) {
```
alecritson, line 130: The `releaseStock` method above only returns `true` statements so how would this condition be met?

packages/core/src/Models/Cart.php
```diff
@@ -609,4 +609,14 @@ public function checkFingerprint($fingerprint)
             );
         });
     }
+
+    /**
+     * Reserves stock for all of the Cart's lines.
+     */
+    public function reserveStock(): void
+    {
+        $this->lines->each(function ($line) {
```
alecritson, line 618: Can be shortened to:

```php
$this->lines->each(
  fn($line) => $line->reserveStock()
);
 ```

packages/core/src/Drivers/SimpleStockDriver.php
glennjacobs: I don't see why you would want to check historic stock. The purpose is to check the current stock. If the stock has changed within a second or so, we still want to know.

packages/core/src/Drivers/SimpleStockDriver.php
glennjacobs: The check stock function isn't part of the interface. It might want to be a private function.

packages/core/src/Drivers/SimpleStockDriver.php
glennjacobs: This is a "simple" stock driver. It doesn't get into the complexities of inventory.

packages/core/src/Drivers/SimpleStockDriver.php
glennjacobs: Correct. By expiring the StockRervation it is in effect released.

glennjacobs:
After long discussions, we have decided to abandon this approach, in favour of a more comprehensive inventory system later down the line.

krzysztofgackowski:
oh no :(
I was relly looking forward to get that implemented.

glennjacobs:
> oh no :( I was relly looking forward to get that implemented.

Sorry about that. For the moment you'll have to manually control the stock levels.

PR #1134, opened 2023-07-03
Add helper functions to Order model

glennjacobs:
An idea to help devs understand the "placed" status of an order. If we don't have a `placed_at` value, we consider the order draft.

- [x] Change log
- [x] Docs

### DIFF
--- a/packages/core/src/Models/Order.php
+++ b/packages/core/src/Models/Order.php
@@ -351,8 +351,6 @@ class Order extends BaseModel
 
     /**
      * Determines if this is a draft order.
-     *
-     * @return bool
      */
     public function isDraft(): bool
     {
@@ -361,8 +359,6 @@ class Order extends BaseModel
 
     /**
      * Determines if this is a placed order.
-     *
-     * @return bool
      */
     public function isPlaced(): bool
     {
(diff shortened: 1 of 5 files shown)


packages/core/src/Models/Order.php
```diff
@@ -348,4 +348,20 @@ protected function getSearchableAttributes()
...
+    {
+        return $this->placed_at ? false : true;
+    }
+
+    /**
+     * Determines if this is a placed order.
+     */
+    public function isPlaced(): bool
+    {
+        return $this->placed_at ? true : false;
```
alecritson, line 365:
```php
!blank($this->placed_at)
```

packages/core/src/Models/Order.php
```diff
@@ -348,4 +348,20 @@ protected function getSearchableAttributes()
 
         return $data;
     }
+
+    /**
+     * Determines if this is a draft order.
+     */
+    public function isDraft(): bool
+    {
+        return $this->placed_at ? false : true;
```
alecritson, line 357:
```php
!$this->placedAt();
``` 

?

alecritson (approved):

PR #1138, opened 2023-07-07
Feature/exclude searchable attributes

dewebdesigns:
Added a new configuration field in the search.php file to exclude searchable attributes.

This makes it easy to exclude information you do not want included when indexing them.

I have also included a test, and updated the docs.

### DIFF
--- a/docs/.vitepress/config.js
+++ b/docs/.vitepress/config.js
@@ -46,7 +46,7 @@ export default defineConfig({
         ]
       },
       {
-        text: '0.3',
+        text: '0.4',
         items: [
           { text: 'Changelog', link: '/core/upgrading' },
           { text: 'Contributing', link: '/core/contributing' }
--- a/docs/core/reference/search.md
+++ b/docs/core/reference/search.md
@@ -38,6 +38,34 @@ php artisan lunar:search:index
 
 The command will import the records of the models listed in the `lunar/indexer.php` configuration file. Type `--help` to see the available options.
 
+## Exclude Attributes
+
+You can exclude any attribute on a Lunar model by adding the attribute to the respective model's array in the `config/search.php`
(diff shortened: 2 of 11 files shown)


glennjacobs:
@dewebdesigns you should be targeting the `main` branch for new features.

Also, I'm not sure config is the correct approach, but I agree there should be a well-documented way of changing the Scout indexing. One way is implementing your own model, which you can do already.

@alecritson do we want a specific way to update the scout indexing on models, or do we simply advise developers to extend the model?

wychoong:
> @dewebdesigns you should be targeting the `main` branch for new features.
> 
> Also, I'm not sure config is the correct approach, but I agree there should be a well-documented way of changing the Scout indexing. One way is implementing your own model, which you can do already.
> 
> @alecritson do we want a specific way to update the scout indexing on models, or do we simply advise developers to extend the model?

I think unless we found a solid way for model extension, I would advise not to. 

For this case I would suggest to use the observer way https://docs.lunarphp.io/core/extending/search.html
But currently the api is only for one way adding https://github.com/lunarphp/lunar/blob/0.4/packages/core/src/Base/Traits/Searchable.php

So I would suggest a mutator approach where you can add or remove as per needed.

glennjacobs:
I think that makes sense. Maybe we need remove and replace methods, or similar.

dewebdesigns:
I can take down the PR if you are planning on going in a different direction. I would like to request the following change at least from this PR:

Change `getSearchableAttributes` from `protected` to `public` on `Lunar\Models\Order`

I can put up a different PR, or one of you guys can add it to one of your current features - doesn't matter to me.

glennjacobs:
I think the direction @wychoong suggested is the best approach.

Issue #1139, opened 2023-07-10
Offline Payment referencing old cart->order relation


glennjacobs:
See https://github.com/lunarphp/lunar/blob/0b54f4d269e15b60b4b4bd12ba9dfef86eb5d112/packages/core/src/PaymentTypes/OfflinePayment.php#L18C54-L18C54

Most likely needs changing to `draftOrder()`.

Also need to ensure we have test coverage for this, as it should have been picked up.

PR #1140, opened 2023-07-10
Hotfix to Offline payment driver referencing old order relation

glennjacobs:
This update aims to fix the offline payment driver which was referencing the old cart->order relation, which no longer exists. See #1139 

- [ ] Tests
- [ ] Changelog


packages/core/src/Actions/Carts/ValidateCartForOrder.php
```diff
@@ -22,7 +22,7 @@ public function execute(
         Cart $cart
     ) {
         // Does this cart already have an order?
-        if ($cart->order) {
+        if ($cart->completedOrder()->first()) {
```
alecritson, line 25: Should this take into account allowing multiple orders?

glennjacobs: We have two options. If someone wants to support multiple orders for a single cart, we either make it a setting somehow, or they implement their own validator. Not sure which is best? Can we really be introducing a setting?

alecritson (dismissed):

alecritson (approved):

PR #1149, opened 2023-07-18
Prevent discount compounding

ryanmitchell:
On one of our live 0.3 sites we are seeing the same discounts being applied multiple times due to Cart::calculate() being called more than once in the same request. Great for the customer, not so great for the vendor.

To fix this cart lines ~~discountTotal and~~ subTotalDiscounted need to be removed before discounts are applied.


alecritson:
Thanks @ryanmitchell just having a little look at this, since we're calling `CalulateLines` before we calculate discounts, would it work to just remove where it's being set on there? https://github.com/lunarphp/lunar/blob/0.4/packages/core/src/Pipelines/Cart/CalculateLines.php#L40

ryanmitchell:
I've a feeling you already fixed this in 0.4 but I haven't had a chance to upgrade yet.
Can we leave this open for now and once we upgrade I'll update it?

alecritson:
Sure! I do have a vague memory of this issue on a site I was working on and trying to fix it, but wasn't sure 🤔

PR #1150, opened 2023-07-18
Fix Postgres support

### LINKED ISSUE #1143: PopulateProductOptionLabelWithName fails when using Postgress
- Lunar version: 0.4.0
- Laravel Version: v10.13.5
- PHP Version: 8.1.17
- Database Driver & Version: pgsql 

### Steps To Reproduce:
Just upgrade to 0.4.0 and run php artisan migrate:fresh --seed

Issue is due to 
protected function shouldRun()
{
    return ProductOption::where('label', '')->orWhereNull('label')->count() > 0;
}

You can't have '' value for json fields in postgres.. it should be 

return ProductOption::whereNull('label')->count() > 0;

glennjacobs:
Trying to fix #1143 so that Lunar installs without errors when using PostgreSQL.

Trying to add MySQL and PostgreSQL testing is causing a big issue, due to the nature of how the `refreshDatabase` trait works. Issues with JSON formatting, duplicate entries, etc all make the tests fail.

### DIFF
--- a/.github/workflows/db_mysql_tests.yml
+++ b/.github/workflows/db_mysql_tests.yml
@@ -30,4 +30,4 @@ jobs:
         composer install --no-scripts
 
     - name: Run Testsuite
-      run: vendor/bin/phpunit
+      run: DB_CONNECTION=mysql DB_HOST=mysql DB_DATABASE=test DB_USERNAME=root DB_PASSWORD=password vendor/bin/phpunit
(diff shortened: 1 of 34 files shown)


obayesshelton:
Any news on when this might be deployed. I cannot use this branch to test and use lunar/admin 

or at least I can't see a way to do so.

glennjacobs:
Testing in MySQL and Postgres was too much of a challenge. Maybe come back to it!

PR #1154, opened 2023-07-25
Change shipping option to match the ApplyShipping functon

dewebdesigns:
Use the Manifest function instead of the same code that fixed the bug in #1126

### DIFF
--- a/packages/core/src/Pipelines/Cart/CalculateTax.php
+++ b/packages/core/src/Pipelines/Cart/CalculateTax.php
@@ -105,12 +105,6 @@ class CalculateTax
 
     private function getShippingOption(Cart $cart)
     {
-        if (! $cart->shippingAddress) {
-            return null;
-        }
-
-        return ShippingManifest::getOptions($cart)->first(function ($option) use ($cart) {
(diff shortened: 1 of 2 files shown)


wychoong:
Opps it got missed out from the previous PR. 

I think would be good to extract it into trait and update the ApplyShipping.php as well

packages/core/src/Pipelines/Cart/CalculateTax.php
```diff
@@ -105,12 +105,6 @@ public function handle(Cart $cart, Closure $next)
 
     private function getShippingOption(Cart $cart)
```
alecritson, line 106: I think if we're just calling `ShippingManifest::getShippingOption($cart)` in this method then we can just do away with this private method altogether and just call it on line 62

alecritson (changes requested):

packages/core/src/Pipelines/Cart/CalculateTax.php
wychoong: that's a good one

PR #1157, opened 2023-07-27
Add remove searchable/sortable/filterable fields for search index

wychoong:
replace #1138

- [] Documentation updates
- [x] Automated tests
- [] Changelog entries (core and hub)

added remove and mutator api

### DIFF
--- a/docs/core/extending/search.md
+++ b/docs/core/extending/search.md
@@ -40,6 +40,9 @@ class OrderObserver
             'custom_field',
             $order->meta->custom_field
         );
+
+        // or to remove field
+        $order->removeSearchableAttribute('field_one');
     }
 
     /**
@@ -54,6 +57,15 @@ class OrderObserver
 
         $order->addSortableAttributes([
             'custom_field'
+        ]);
+
+        // or to remove fields
+        $order->removeFilterableAttributes([
+            'field_one',
+        ]);
+
+        $order->removeSortableAttributes([
+            'field_one',
         ]);
     }
 }
--- a/packages/core/src/Base/Traits/Searchable.php
+++ b/packages/core/src/Base/Traits/Searchable.php
@@ -14,7 +14,7 @@ trait Searchable
      *
      * @var array
      */
-    protected $additionalSearchFields = [];
+    protected $additionalSearchableFields = [];
 
     /**
      * Define the additional filterable fields.
@@ -120,7 +120,22 @@ trait Searchable
      */
     public function addSearchableAttribute($key, $value)
     {
-        $this->additionalSearchFields[$key] = $value;
+        $this->additionalSearchableFields[$key] = $value;
+    }
+
+    /**
+     * Remove an attribute from additional searchable fields.
+     *
+     * @param  string  $key
+     * @return void
+     */
+    public function removeSearchableAttribute($key)
+    {
+        $baseAttributes = $this->getSearchableAttributes();
+
+        if (! isset($baseAttributes[$key])) {
+            unset($this->additionalSearchableFields[$key]);
+        }
     }
 
     /**
@@ -136,7 +151,7 @@ trait Searchable
 
         return array_merge(
             $this->getSearchableAttributes(),
-            $this->additionalSearchFields
+            $this->additionalSearchableFields
         );
     }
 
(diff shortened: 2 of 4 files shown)


wychoong:
@ dewebdesigns can you test if this does what you need?

alecritson:
Wondering if there is a nicer way to allow searchable attribute extending, rather than having to override or mutate existing values, I think it'd be nicer if you can just specify everything in one place. My concern is that if these methods get called throughout the codebase then they're constantly fighting each other.

For example the developer calls `addSearchableAttribute('foobar')` and then by chance an add on (or another dev in the same project) calls `removeSearchableAttribute('foobar')` somewhere else, it's not obvious what's happening or why it's not being indexed.

wychoong:
> Wondering if there is a nicer way to allow searchable attribute extending, rather than having to override or mutate existing values, I think it'd be nicer if you can just specify everything in one place. My concern is that if these methods get called throughout the codebase then they're constantly fighting each other.
> 
> For example the developer calls `addSearchableAttribute('foobar')` and then by chance an add on (or another dev in the same project) calls `removeSearchableAttribute('foobar')` somewhere else, it's not obvious what's happening or why it's not being indexed.

maybe a class to define, or just model extension to override methods

alecritson:
> > Wondering if there is a nicer way to allow searchable attribute extending, rather than having to override or mutate existing values, I think it'd be nicer if you can just specify everything in one place. My concern is that if these methods get called throughout the codebase then they're constantly fighting each other.
> > For example the developer calls `addSearchableAttribute('foobar')` and then by chance an add on (or another dev in the same project) calls `removeSearchableAttribute('foobar')` somewhere else, it's not obvious what's happening or why it's not being indexed.
> 
> maybe a class to define, or just model extension to override methods

I was thinking a class you could easily define in the config, like we do for other things such as URL generators, I think it'd be good to give the developer total control so they can determine document size etc.

Something like:

```php
use Illuminate\Database\Eloquent\Builder;

class ProductIndexDocument
{
    public function getSearchableAttributes(): array
    {
        return [];
    }

    public function getSortableAttributes(): array
    {
        return [];
    }

    public function getFilterableAttributes(): array
    {
        return [];
    }

    public function extendQuery(Builder $query): Builder
    {
        return $query->with(['foo']);
    }
}
```

as an example, what do you think @glennjacobs ?

wychoong:
lunar can ship default fields, and how dev can remove fields from the defaults

as an example, currently I extend the Product model to do this 
```php
public function getSearchableAttributes()
    {
        $data = parent::getSearchableAttributes();

        return Arr::except($data, [
            'brand_id',
            'product_type_id',
        ]);
    }
```

and the index event 
```php
    {
        // add brand
        $brand = $product->brand;

        if ($brand) {
            $product->addSearchableAttribute('brand', $brand->name);
        }
    }
```

alecritson:
Yeah I think Lunar defaults could be better but should only provide basic attributes. Could even have a base class that's extended so you can choose whether to include what Lunar proposes or not.

alecritson:
This is going to be out the scope of `0.4` so would suggest targeting `main` :)

wychoong:
> This is going to be out the scope of `0.4` so would suggest targeting `main` :)

if Glenn agreed with that approach, can you take over for the implementation?

glennjacobs:
> > This is going to be out the scope of `0.4` so would suggest targeting `main` :)
> 
> if Glenn agreed with that approach, can you take over for the implementation?

I think I agree with the approach. We'd need to ship default implementations for the Models that are searchable and then allow the developer to switch them out for their own implementation. 

@alecritson would this be implemented on a base Model, so any Lunar Model could have scout searching configured?

alecritson:
> > > This is going to be out the scope of `0.4` so would suggest targeting `main` :)
> > 
> > 
> > if Glenn agreed with that approach, can you take over for the implementation?
> 
> I think I agree with the approach. We'd need to ship default implementations for the Models that are searchable and then allow the developer to switch them out for their own implementation.
> 
> @alecritson would this be implemented on a base Model, so any Lunar Model could have scout searching configured?

If we wanted any model to be searchable then we could do it on the base model, otherwise we'd need to be explicit but overrides would still need to be mapped in config I guess, which is fine.

glennjacobs:
@alecritson is going to work on a replacement PR today.

PR #1167, opened 2023-08-01
fix: extract getShippingOption to trait

wychoong:
can replace #1154 if trait is preferred


alecritson:
Given that calling `ShippingManifest::` is quite simple I don't think the trait is required. Thanks though! :)

PR #1168, opened 2023-08-01
Fix: SystemTaxDriver only adds tax amounts for the applicable tax zones

webcraft:
When a `TaxClass` has multiple `TaxRateAmount`s (e.g. one for every country), the `SystemTaxDriver` incorrectly applies every one of them, regardless of whether the `TaxZone` of the `TaxRateAmount` is relevant for the shipping address.

I believe this is a regression introduced in commit 809feeb7515203ded956138bcbd4a287cf07dbb2 when some code was wrapped in a `Blink::once()`. However, at the same time, the code inside was altered slightly, which changed the logic.

Please refer to the test I added in this PR. The `TaxClass` has 4 `TaxRateAmount`s, 2 of which should be applied (the default ones), and 2 of which should not be applied.

Running this test on the current `SystemTaxDriver` will fail.

My proposal is to revert the logic in `SystemTaxDriver` to what it was before the mentioned commit, while still being wrapped in `Blink::once()`.


glennjacobs:
Thanks, I'll take a good look this week.

alecritson (approved):

Issue #1156, opened 2023-07-27
When calculating shipping total the stored_inclusive_of_tax setting is ignored


jandol5:
- Lunar version: 0.4.0
- Laravel Version: 10.10.1
- PHP Version: 8.1

### Expected Behaviour:
The cart total should be calculated correctly when the pricing.stored_inclusive_of_tax is set to true or false.

### Actual Behaviour:
The cart total is calculated correctly only when pricing.stored_inclusive_of_tax is set to false. When it is set to true, the shipping tax is added up to the cart total together with shipping subtotal which already includes the tax.

### Fix:
In CalculateTax.php pipeline the lines [83-87](https://github.com/lunarphp/lunar/blob/0b54f4d269e15b60b4b4bd12ba9dfef86eb5d112/packages/core/src/Pipelines/Cart/CalculateTax.php#L83C1-L87C15) should be change from
<code>
$cart->shippingTotal = new Price(
&nbsp;&nbsp;&nbsp;&nbsp;$shippingSubTotal + $shippingTaxTotal?->value,
&nbsp;&nbsp;&nbsp;&nbsp;$cart->currency,
&nbsp;&nbsp;&nbsp;&nbsp;1
);
</code>
to
<code>
$shippingTotal = $shippingSubTotal;
if(!prices_inc_tax()) {
&nbsp;&nbsp;&nbsp;&nbsp;$shippingTotal += $shippingTaxTotal?->value;
}
$cart->shippingTotal = new Price(
&nbsp;&nbsp;&nbsp;&nbsp;$shippingTotal,
&nbsp;&nbsp;&nbsp;&nbsp;$cart->currency,
&nbsp;&nbsp;&nbsp;&nbsp;1
);
</code>
in order for price.stored_inclusive_of_tax setting to be respected accordingly.

If the fix is acceptable, I'm willing to prepare a PR.

glennjacobs:
A PR would be good with a test to show it working, thanks.

webcraft:
I can confirm the bug and the proposed fix.
Thanks @jandol5, please let me know if you need help with a PR.

jandol5:
Thanks @glennjacobs & @webcraft for your input, the PR #1170 is ready and it includes test.

PR #1170, opened 2023-08-02
Fix: cart total calculation when pricing.stored_inclusive_of_tax is set to true

jandol5:
Resolves issue [1156](https://github.com/lunarphp/lunar/issues/1156).


alecritson:
Thanks @jandol5 Could we get a changelog entry please?

glennjacobs (approved):

PR #1175, opened 2023-08-03
Fix/scout search page

wychoong:
for details see #1078 

- Documentation updates
- Automated tests


packages/admin/src/Tables/Builders/CustomersTableBuilder.php
```diff
@@ -21,7 +21,10 @@ public function getData(): iterable
         $query = Customer::query();
 
         if ($this->searchTerm) {
-            $query->whereIn('id', Customer::search($this->searchTerm)->keys());
+            $query->whereIn('id', Customer::search($this->searchTerm)
+                ->query(fn ($query) => $query->select('id'))
+                ->take(500)
```
alecritson, line 26: Would it not make sense to use `$this->perPage` so they're both consistent?

wychoong: No. using `$this->perPage` won't solve the issue as it only increase the return result from Scout (assuming per page is more than 20), not Scout pagination per page. And since it is using same value, it will only return a fixed amount from Scout and the table can never paginate.

current fix is to force Scout to return more results and use Eloquent pagination.

detailed finding is here https://github.com/lunarphp/lunar/pull/1078#issuecomment-1663274217

alecritson: Gotcha, I think this should be fine for now since we're going to be migrating to Filament in the future.

alecritson:
@wychoong This needs to target `main` I did try myself but it kept showing more changes, if you update from `main` and then change the target hopefully it'll sort itself out.

alecritson (approved):

PR #1176, opened 2023-08-03
Default tax rate amount based on tax class

se09deluca:
Issue:
The priceIncTax method of the Price class calculates the price with a default tax rate that is the sum of all tax rate amounts belonging to the default tax  zone. It doesn't consider the product's tax class.

Solution:
If priceable is present on the Price instance uses its tax class to determine the rate amount, otherwise uses the defaults tax rate amounts.


glennjacobs:
On initial inspection, I think that looks good. Could you add a test to cover this?

se09deluca:
Hi @glennjacobs  as you can see i added a test covering the updated functionality and some related fix. 
Now i don't know what is the problem with vercel's bot as i don't know enough.

If there's something else i need to do, please let me know.

glennjacobs:
Looks good, don't worry about Vercel.

alecritson:
@se09deluca Could you please add a changelog entry briefly describing what this PR adds :)

glennjacobs:
@alecritson would you agree we're going to lose performance from this due to the Blink cache name change? The change is correct, but I'm trying to think about how we can retain performance. As the cache name is based upon the purchasable ID, we're not going to get much benefit from the Blink cache.

glennjacobs:
Maybe we need the cache name to be based upon `$this->priceable?->taxClass` as that should already be loaded?

alecritson:
> @alecritson would you agree we're going to lose performance from this due to the Blink cache name change? The change is correct, but I'm trying to think about how we can retain performance. As the cache name is based upon the purchasable ID, we're not going to get much benefit from the Blink cache.

Yeah I would agree and also the name feels incorrect to me `getDefaultTaxRate` as it's not always returning the default from the DB so is there a wider issue?

se09deluca:
Since the [priceable is required](https://docs.lunarphp.io/core/reference/products.html#overview-1) on the price model i assumes it is always present. 
The rate amounts are now filtered only for the default tax zone.
The methods signature has been updated to better reflect how the rate amounts are calculated. priceInc

Do you see any others errors/problems ?

packages/core/src/Models/Price.php
```diff
@@ -107,18 +107,25 @@ public function priceIncTax()
...
+     */
+    protected function getPriceableTaxRate()
     {
-        return Blink::once('price_default_tax_rate', function () {
+        return Blink::once('price_tax_rate_'.$this->priceable->taxClass->id, function () {
             $taxZone = TaxZone::where('default', '=', 1)->first();
 
-            if ($taxZone) {
-                return $taxZone->taxAmounts->sum('percentage') / 100;
+            if ($taxZone && ! is_null($taxClass = $this->priceable->taxClass)) {
```
glennjacobs, line 125: Considering we're using the "purchasable", we should probably be using that interface's method `getTaxClass()`.

See <https://github.com/lunarphp/lunar/blob/0.4/packages/core/src/Base/Purchasable.php#L22-L25>

@alecritson do you agree?

alecritson: Yeah I think that's a good shout.

glennjacobs: @se09deluca can you make this change, then I think it's ready to be merged. Thanks.

se09deluca: Done! Thank you for the hint

packages/core/src/Models/Price.php
```diff
@@ -107,18 +107,25 @@ public function priceIncTax()
...
-    protected function getDefaultTaxRate()
+    /**
+     * Return the total tax rate amount within the predefined tax zone for the related priceable
+     *
+     * @return int|float
+     */
+    protected function getPriceableTaxRate()
     {
-        return Blink::once('price_default_tax_rate', function () {
+        return Blink::once('price_tax_rate_'.$this->priceable->taxClass->id, function () {
```
glennjacobs, line 122: you missed one ;-)

glennjacobs (approved):

PR #1187, opened 2023-08-07
Fix get shipping option in cart pipelines

wychoong:


dewebdesigns:
It would be great to merge this into the next release if possible @glennjacobs . I've been struggling with infinite loops when testing my checkout flow because of the shipping option bug in calculateTax.php that's fixed in this PR.

dewebdesigns:
Any update on getting this merged in @alecritson or @glennjacobs ?

alecritson (approved):

PR #1189, opened 2023-08-10
[Fix] Dashboard error when top selling product variant deleted

### LINKED ISSUE #1179: Deleting a ProductVariant breaks the Hub Dashboard
- Lunar version: 0.4.0
- Laravel Version: 10.17.1
- PHP Version: 8.2.8
- Database Driver & Version: MySQL v8

### Expected Behaviour:
Deleting a Product Variant should not break the Hub Dashboard view.

### Actual Behaviour:
Deleting a Product Variant breaks the Hub Dashboard view, with a handful of errors caused by the `purchasable` property not existing inside `$this->topSellingProducts`

### Steps To Reproduce:
1. Delete a Product Variant having the most (or 2nd most) Orders (can use the "Top Selling Products" widget on the Hub Dashboard to determine this)
2. Observe errors on the Hub Dashboard page

I think the solution here is to refactor the `getTopSellingProductsProperty` computed property to exclude null purchasables in the Dashboard Livewire component: https://github.com/lunarphp/lunar/blob/0b54f4d269e15b60b4b4bd12ba9dfef86eb5d112/packages/admin/src/Http/Livewire/Dashboard.php#L138

khatriafaz:
The issue:

When a product variant was deleted and it was in the top 2 selling products, the dashboard was broken.

The fix:

This fix adds a `whereHas` where it checks if the variant exists and only includes the products whose variant exists in the dashboard top selling items.

Fixes #1179

### DIFF
--- a/packages/admin/src/Http/Livewire/Dashboard.php
+++ b/packages/admin/src/Http/Livewire/Dashboard.php
@@ -152,6 +152,7 @@ class Dashboard extends Component
             now()->parse($this->from),
             now()->parse($this->to),
         ])->where('type', '!=', 'shipping')
+            ->whereHas('purchasable')
             ->groupBy('purchasable_type', 'purchasable_id')
             ->orderBy('count', 'desc')
             ->take(2)->get();


kylekanderson:
@khatriafaz - I'm getting the following error on the dashboard when testing this:

> Too few arguments to function Lunar\DataTypes\ShippingOption::__construct(), 0 passed in /vendor/laravel/framework/src/Illuminate/Database/Eloquent/Concerns/HasRelationships.php on line 793 and at least 5 expected

I've seen this before when trying to query against the purchasable relationship.

khatriafaz:
@kylekanderson I will try to reproduce and fix it.

I can work on this in a few days. ✌️

kylekanderson:
@khatriafaz Thanks for taking a look at this. I just submitted a PR that should resolve this issue (https://github.com/lunarphp/lunar/pull/1197). Feel free to have a look.

alecritson (approved):

kylekanderson:
@alecritson - just a heads up that this Pull Request was still resulting in an error on the Dashboard when a Top-Selling Product Variant is deleted. My Comment above (https://github.com/lunarphp/lunar/pull/1189#issuecomment-1675312186) has the details of that error. That was the reason for PR #1197.

khatriafaz:
@kylekanderson I tried checking it, but I don't see the error you are seeing.
Any specific scenario? I am using the built-in seeders to seed the database and then deleting one of the top-selling variants.

kylekanderson:
@alecritson @khatriafaz - Confirming this bug is still appearing in the latest 0.5 release.

It seems that this only happens when I have OrderLines with a type of `shipping`. An example of such an OrderLine is below:
2023-08-18 at 11 18 52

khatriafaz:
@kylekanderson Maybe this can go to a new issue. I think the issue is unrelated to the fix applied in this PR.
But still needs to be worked on.

kylekanderson:
@khatriafaz - Sure thing. Just opened Issue #1211 for this.

PR #1194, opened 2023-08-14
Allow hub to be disabled

ryanmitchell:
In some system setups the hub would not be available on the same URL as the shopfront, for example when its a white-labelled store front. However it is still necessary to include the hub composer requirement as add-ons extend it (adding views, slots, components etc) as well as providing front end functionality.

In this situation it would be useful to be able to stop the hub displaying by not letting it respond on any routes.

This PR adds a `lunar-hub.system.enable` config setting, that when set to `false` does not enable any hub routes.


ryanmitchell:
Got to love how pint highlights files that havent been changed by the PR 😕

wychoong:
> Got to love how pint highlights files that havent been changed by the PR 😕

Glenn updated the pint to use the `--dirty` flag in main branch. but tbh i did like pint just fix and commit the changes

alecritson:
Thanks @ryanmitchell is there any reason we couldn't add this logic to the service provider? i.e. https://github.com/lunarphp/lunar/blob/0.4/packages/admin/src/AdminHubServiceProvider.php#L203

ryanmitchell:
Nope, no reason not to do it there. I'll update later.

glennjacobs:
@ryanmitchell as this isn't a bug fix, can you please target `main`?

ryanmitchell:
Yeah no problem. Thats updated.

glennjacobs (approved):

PR #1197, opened 2023-08-15
Fix: Deleting a ProductVariant breaks the Hub Dashboard

kylekanderson:
The issue:

When a product variant was deleted and it was in the top 2 selling products, the dashboard was broken.

The fix:

This fix updates the Top Selling Products section of the Hub Dashboard to fall back to the `description` and `identifier` properties of the `OrderLine` model when the `$product->purchasable` (the `ProductVariant`) model has been deleted.

Fixes https://github.com/lunarphp/lunar/issues/1179


khatriafaz:
@kylekanderson This solution is even better than #1189 

Changes look good.

alecritson:
Thanks for the PR, I think in this case we'll go with #1189 since it's a bit simpler and since we're gearing up for filament I think it makes the most sense. Appreciate your work on this though!

PR #1198, opened 2023-08-16
fix: set ListField data to `[]` rather than `null`

fakeheal:
**Issue:** When a store has attributes of type `ListField` attached to product types and you try adding a product via the hub, switching between product types breaks the UI of all attributes of `ListField` type.

I think the problem is caused by setting `$value` (see the changes) to `null`, `ListField` should be set to `[]` (an empty array).

[Here's a short video of the problem](https://github.com/lunarphp/lunar/assets/1038697/faba3098-fe04-4b04-9c74-f4e92022a5d9), and [this is after my fix](https://github.com/lunarphp/lunar/assets/1038697/ca189018-baf5-4d29-9bca-5a481b023ba4).

### DIFF
--- a/packages/admin/resources/views/partials/forms/currency.blade.php
+++ b/packages/admin/resources/views/partials/forms/currency.blade.php
@@ -25,7 +25,7 @@
           :label="__('adminhub::inputs.default.label')"
           for="handle"
           :instructions="__('adminhub::inputs.default.instructions', [
-            'model' => 'channel',
+            'model' => 'currency',
           ])"
         >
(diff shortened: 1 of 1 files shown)


alecritson:
Thanks @fakeheal although it looks like the branch you created has gone a bit funky so it's hard to see just your changes.

fakeheal:
> Thanks @fakeheal although it looks like the branch you created has gone a bit funky so it's hard to see just your changes.

Yes, it's because my fork has only the main branch. I will convert it to a draft and fix it today. Sorry!

PR #1202, opened 2023-08-17
[Fix] Pricing not being deleted when variant is deleted

### LINKED ISSUE #1190: Prices are not being deleted when deleting ProductVariants
- Lunar version: 0.4.1
- Laravel Version: 10.10
- PHP Version: 8.2
- Database Driver & Version: MariaDB v10

### Expected Behaviour:
The corresponding pricing rows in the lunar_prices table should be deleted.

### Actual Behaviour:
The pricing rows are not being deleted and will stay put until you manually remove then from the database

### Steps To Reproduce:
Create a product and toggle the Variants option. Create the variants by selecting the product options for the variants and click "Use selected options" button.

Then either remove the variants individually or remove the all at once by toggling the variants option to off on the product page. Check the database lunar_prices table and you will see that the corresponding prices will be left in the database.

khatriafaz:
The issue:
The corresponding pricing rows in the `lunar_prices` table were not deleted when the variant was deleted either from the product show page by disabling variants or from the variant show page itself.

The fix:
This fix now deletes the record from the `lunar_prices` table for the deleted variant.

Closes #1190

### DIFF
--- a/packages/admin/src/Http/Livewire/Components/Products/AbstractProduct.php
+++ b/packages/admin/src/Http/Livewire/Components/Products/AbstractProduct.php
@@ -412,6 +412,7 @@ abstract class AbstractProduct extends Component
                 DB::transaction(function () use ($variantsToRemove) {
                     foreach ($variantsToRemove as $variant) {
                         $variant->values()->detach();
+                        $variant->prices()->delete();
                         $variant->forceDelete();
                     }
                 });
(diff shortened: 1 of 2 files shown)


alecritson (approved):

obayesshelton:
Hi, 

I'm using Postgres and this is still an issue... not sure if anyone has mentioned this.

PR #1204, opened 2023-08-17
fix: generate unique slugs using random string if necessary

fakeheal:
**Issue:** Creating a products with the same name, produces equal default URLs.

**Solution**: If slug exists, append random 5 character string <sup>1</sup>.

**Example:**
1. **Product**: Test Product, **Slug**: _test-product_
2. **Product**: Test Product, **Slug**: _test-product-abcd_

1: Another approach is to append the number of models matching the url, eg: test-product, test-product-1, test-product-2, etc.


glennjacobs:
Thanks @fakeheal 

Could we go with the other approach, using numbers? E.g. test-product, test-product-1, test-product-2, etc.

I feel like that would be a tidier approach and probably a bit more understandable. A random string might just serve to confuse.

glennjacobs (changes requested):
Change to use a numbering system instead.

fakeheal:
@glennjacobs Changed it to a numbering system (heavily inspired by [eloquent-sluggable](https://github.com/cviebrock/eloquent-sluggable)).

**Example:**
1. **Product**: Test Product, **Slug**: _test-product_
2. **Product**: Test Product, **Slug**: _test-product-2_
3. **Product**: Test Product, **Slug**: _test-product-3_

fakeheal (commented):
Changed it to a numbering system (heavily inspired by [eloquent-sluggable](https://github.com/cviebrock/eloquent-sluggable)).

**Example:**
1. **Product**: Test Product, **Slug**: _test-product_
2. **Product**: Test Product, **Slug**: _test-product-2_
3. **Product**: Test Product, **Slug**: _test-product-3_

glennjacobs (approved):

PR #1207, opened 2023-08-18
Fix amount off to share percentage take 2

ryanmitchell:
Replaces #949


glennjacobs:
@ryanmitchell any chance you could fix the failing tests so we can get this reviewed and merged before 0.6 is released?

glennjacobs (approved):

PR #1219, opened 2023-08-24
Fix price formatting for currencies with symbol to the right

adevade:
`trimTrailingZeros` did not work for Swedish currency, because our symbol is shown to the right.

This PR adds a test and updates the regex to support other currencies (or at least SEK).

---

I feel like a regex could be a little flaky for this functionality? 

Also in Sweden, if there's no value after the decimals, we usually don't show them at all. So `150,00 kr` is shown as `150 kr`. Do you think that users should write custom logic for stuff like that, or should it be handled by Lunar?

### DIFF
--- a/packages/core/src/DataTypes/Price.php
+++ b/packages/core/src/DataTypes/Price.php
@@ -110,7 +110,9 @@ class Price
         $formattedPrice = $formatter->format($value);
 
         if ($trimTrailingZeros) {
-            $formattedPrice = preg_replace('/(\.\d{'.$this->currency->decimal_places.'}\d*?)0+$/', '$1', $formattedPrice);
+            $decimalSeparator = $formatter->getSymbol(NumberFormatter::DECIMAL_SEPARATOR_SYMBOL);
+
+            $formattedPrice = preg_replace('/(\\'.$decimalSeparator.'\d{'.$this->currency->decimal_places.'}\d*?)0+(\s*\D*)$/', '$1$2', $formattedPrice);
         }
 
         return $formattedPrice;
--- a/packages/core/tests/Unit/DataTypes/PriceTest.php
+++ b/packages/core/tests/Unit/DataTypes/PriceTest.php
@@ -121,6 +121,23 @@ class PriceTest extends TestCase
     }
 
     /** @test */
+    function can_format_numbers_specifying_decimal_places_with_currency_suffix()
+    {
+        $currency = Currency::factory()->create([
+            'code' => 'SEK',
+            'decimal_places' => 2,
+        ]);
+
+        $dataType = new Price(15000, $currency, 1);
+        $this->assertEquals('150,00 kr', $dataType->formatted(locale: 'sv', decimalPlaces: 6, trimTrailingZeros: true));
+        $this->assertEquals('150,000000 kr', $dataType->formatted(locale: 'sv', decimalPlaces: 6, trimTrailingZeros: false));
+
+        $dataType = new Price(50050, $currency, 100);
+        $this->assertEquals('5,005 kr', $dataType->unitFormatted(locale: 'sv', decimalPlaces: 6, trimTrailingZeros: true));
+        $this->assertEquals('5,005000 kr', $dataType->unitFormatted(locale: 'sv', decimalPlaces: 6, trimTrailingZeros: false));
+    }
+
+    /** @test */
     public function can_handle_decimals_being_passed()
     {
         $currency = Currency::factory()->create([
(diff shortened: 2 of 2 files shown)


glennjacobs (approved):

glennjacobs:
> I feel like a regex could be a little flaky for this functionality?

Yeah. I'm not too sure what to replace it with though. It is ultimately just string manipulation.
 
> Also in Sweden, if there's no value after the decimals, we usually don't show them at all. So `150,00 kr` is shown as `150 kr`. Do you think that users should write custom logic for stuff like that, or should it be handled by Lunar?

This is a tricky one. Ultimately Lunar uses NumberFormatter::CURRENCY to format the values. It doesn't seem to have a consideration for decimal places, so we're on our own a bit.

Trying to cater for everyone's requirements would essentially be impossible I feel.

@alecritson made a good suggestion on Discord, which is for us to have the price formatting function split out into a class that is configurable. So then a developer is able to customise it to their own needs. I think that's the way we'd want to go, so we'll look to get that incorporated into the v0.6 release.

adevade:
Maybe we could get the formatted value without the currency symbols, do the regex on that string, and then add the currency symbols back? That way the regex is only running on the actual numeric value, giving it more consistency.

---

Further, I would actually expect `trimTrailingZeros: true` to trim off all decimals if there's only zeros present, for all currencies.
This is the way the `woocommerce_price_trim_zeros` filter [does it in WooCommerce](https://github.com/woocommerce/woocommerce/blob/879136404d3b5445c1f1fb81c5e179d5333c96f4/plugins/woocommerce/tests/legacy/unit-tests/formatting/functions.php#L613-L616). *(They also [do it with a regex](https://github.com/woocommerce/woocommerce/blob/879136404d3b5445c1f1fb81c5e179d5333c96f4/plugins/woocommerce/includes/wc-formatting-functions.php#L217-L225) btw, before adding the currency symbols like I suggested above.)*

Ex: $150.00 -> $150

Most currencies I've seen commonly use 2 decimal places, and below is a scenario where the attribute often doesn't seem to do anything.

Take the following example: 

```php
$currency = Currency::factory()->create([
    'code' => 'USD',
    'decimal_places' => 2,
]);

// Current behaviour
$price1 = new Price(15000, $currency, 1);
$price1->formatted(trimTrailingZeros: true);                    // '$150.00'
$price1->formatted(trimTrailingZeros: false);                   // '$150.00'
$price1->formatted(trimTrailingZeros: true, decimalPlaces: 3);  // '$150.00'
$price1->formatted(trimTrailingZeros: false, decimalPlaces: 3); // '$150.000'

$price2 = new Price(15001, $currency, 1);
$price2->formatted(trimTrailingZeros: true);                    // '$150.01'
$price2->formatted(trimTrailingZeros: false);                   // '$150.01'
$price2->formatted(trimTrailingZeros: true, decimalPlaces: 3);  // '$150.01'
$price2->formatted(trimTrailingZeros: false, decimalPlaces: 3); // '$150.010'

$price3 = new Price(15010, $currency, 1);
$price3->formatted(trimTrailingZeros: true);                    // '$150.10'
$price3->formatted(trimTrailingZeros: false);                   // '$150.10'
$price3->formatted(trimTrailingZeros: true, decimalPlaces: 3);  // '$150.10'
$price3->formatted(trimTrailingZeros: false, decimalPlaces: 3); // '$150.100'



// Expected behaviour (according to me)
$price1 = new Price(15000, $currency, 1);
$price1->formatted(trimTrailingZeros: true);                    // '$150'
$price1->formatted(trimTrailingZeros: false);                   // '$150.00'
$price1->formatted(trimTrailingZeros: true, decimalPlaces: 3);  // '$150'
$price1->formatted(trimTrailingZeros: false, decimalPlaces: 3); // '$150.000'

$price2 = new Price(15001, $currency, 1);
$price2->formatted(trimTrailingZeros: true);                    // '$150.01'
$price2->formatted(trimTrailingZeros: false);                   // '$150.01'
$price2->formatted(trimTrailingZeros: true, decimalPlaces: 3);  // '$150.01'
$price2->formatted(trimTrailingZeros: false, decimalPlaces: 3); // '$150.010'

$price3 = new Price(15010, $currency, 1);
$price3->formatted(trimTrailingZeros: true);                    // '$150.1'
$price3->formatted(trimTrailingZeros: false);                   // '$150.10'
$price3->formatted(trimTrailingZeros: true, decimalPlaces: 3);  // '$150.1'
$price3->formatted(trimTrailingZeros: false, decimalPlaces: 3); // '$150.100'
```

What do you think?

PR #1223, opened 2023-08-25
Feat - Scout configuration rewrite

alecritson:
This PR is off the back of #1157 and attempts to make it possible for developers to cleanly customise search indexing for any model which supports it.

Currently we require developers to use observers and implement methods such as `addSearchableAttribute($key, $value)` which wasn't very flexible in terms of adding extra customisation down the line. This new approach uses classes (which can be overridden) to determine what and how models are indexed.

All existing functionality should remain somewhat the same, but there are some notable things we now have the ability to do, which we couldn't before.

- Actually set attributes on the index which are set to `searchable`
- Set whether a model should be indexed
- Customise the index name
- Tap into the query behind the scenes to improve eager loading
- Customise the scout search key

The approach would be to create your own Indexer class and replace it's mapping in the config.

Going forward, if we wanted to customise anything else it would really just be a case of adding a method to the indexer and it'll be available without much of an upgrade path.

This approach will also negate the possible side effects currently from calls to `addSearchableAttribute($key, $value)`, which could be called from anywhere without the dev knowing explicitly where it's coming from.

Hopefully the initial doc updates in this PR will be enough to show how things are working and if the approach is good then I will carry on implementing and writing tests out plus the upgrade guide.

@wychoong It would be good to get your thoughts on this also, as the author of the initial PR.

### DIFF
--- a/docs/core/extending/search.md
+++ b/docs/core/extending/search.md
@@ -13,25 +13,31 @@ There are three things to consider when you want to extend the search:
 
 ### Default index values
 
-Eloquent models will use an indexer class to tell Scout how each model should be indexed. Lunar ships with
-an `EloquentIndexer` by default which will map the following:
+Eloquent models which use the `Lunar\Base\Traits\Searchable` trait will use an indexer class to tell Scout how each it
+should be indexed, if an indexer isn't mapped in the config the default `EloquentIndexer` (provided by Lunar) will be
+used.
+
+This class will map a basic set of fields to the search index:
 
 - The ID of the model
 - Any `searchable` attributes.
 
 If you are using the `mysql` driver, then the model's `->toArray()` method will simply be called.
 
-Some models, such as products, require a bit more information to be indexed and by default we have created extra index
-classes as follows:
+Some models require a bit more information to be indexed, such as SKU's, prices etc. For these scenarios, dedicated
+indexers have been created and are mapped in the config already.
 
-#### Products
+#### `Lunar\Search\ProductIndexer`
 
-The following additional fields are mapped for products:
+Fields which are indexed:
 
-- `status`
-- `product_type`
-- `brand` (if applicable)
-- `created_at`
+- The ID of the model
+- Any `searchable` attributes.
+- The product `status`
+- The product `product_type`
+- The `brand` (if applicable)
+- The ProductVariant `skus` related to the product.
+- The `created_at` timestamp
 
 ### Document mapping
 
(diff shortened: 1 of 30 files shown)


packages/core/src/Base/Traits/Searchable.php
```diff
@@ -2,142 +2,92 @@
...
-
-        return array_merge(
-            $this->getSearchableAttributes(),
-            $this->additionalSearchFields
+        $data = $this->indexer()->getDocument(
+            $this,
+            config('scout.driver')
         );
+
+        dd($data);
```
glennjacobs, line 90: Probably needs to go

docs/core/extending/search.md
```diff
@@ -2,65 +2,111 @@
...
+- The product `status`
+- The product `product_type`
+- The `brand` (if applicable)
+- The ProductVariant `skus` related to the product.
+- The `created_at` timestamp
+
+### Document mapping
+
+All indexers are mapped in `config/search.php` under `document_indexers`, if a model isn't mapped here then it will
+simply use the default `ELoquentIndexer`. To change how each model is indexed, simply map it like so:
```
glennjacobs, line 45: "ELoquentIndexer"

docs/core/extending/search.md
```diff
@@ -2,65 +2,138 @@
...
-     **/
-    public function indexing(Order $order)
+    // Scout method to return the index name.
+    public function searchableAs(Model $model): string
     {
-        $order->addSearchableAttribute(
-            'custom_field',
-            $order->meta->custom_field
-        );
+        return $model->getKey();
```
wychoong, line 70:
```suggestion
        return 'custom_index_name';
```

wrong example

wychoong:
this approach looks really good. some thoughts on this:

- for `EloquentIndexer` and `ModelIndexerInterface` would it be better to rename as `ScoutIndexer` and `ScoutIndexerInterface`, might need to rename `Console\Commands\ScoutIndexer`
- define `engine_map` key part of the Indexer class
- have a command to generate the indexer class
- add a indexer namespace `App\\Search` to system config. having this we might can just autoload model/indexers from custom's and lunar's. a slight problem is the extended model. this is not much of performance impact as it happens in indexing command only
-  then we can drop `models` and the new `indexers` key as well, and the search config file

alecritson:
Thanks @wychoong I'll try go over each point and mention @glennjacobs to see if he feels differently...

> this approach looks really good. some thoughts on this:
> 
> * for `EloquentIndexer` and `ModelIndexerInterface` would it be better to rename as `ScoutIndexer` and `ScoutIndexerInterface`, might need to rename `Console\Commands\ScoutIndexer`

I think for the moment not having Scout is fine, it would mean in the future (not that we have plans) if it ever changes away from Scout we can still maintain interfaces and classes and the names still make sense.

> Remove search config and auto load
> 
> * define engine part of the Indexer class to replace `engine_map`
> * have a command to generate the indexer class
> * add a indexer namespace `App\\Search` to system config. having this we might can just autoload model/indexers from custom's and lunar's. a slight problem is the extended model
> * then we can drop `models` and the new `indexers` key as well, and the search config file

I'm certainly not against this idea but seeing that we don't do anything like that currently I'd be hesitant to start introducing it here. Other area's would likely benefit from autoloading and having commands to generate classes so I would think that should be a dedicated unit of work.

alecritson:
> this approach looks really good. some thoughts on this:
> 
> * for `EloquentIndexer` and `ModelIndexerInterface` would it be better to rename as `ScoutIndexer` and `ScoutIndexerInterface`, might need to rename `Console\Commands\ScoutIndexer`
> 
> Remove search config and auto load
> 
> * define engine part of the Indexer class to replace `engine_map`
> * have a command to generate the indexer class
> * add a indexer namespace `App\\Search` to system config. having this we might can just autoload model/indexers from custom's and lunar's. a slight problem is the extended model
> * then we can drop `models` and the new `indexers` key as well, and the search config file

Had a chat with @glennjacobs and Implemented the name change to `ScoutIndexer` 😄

wychoong:
> I'm certainly not against this idea but seeing that we don't do anything like that currently I'd be hesitant to start introducing it here. Other area's would likely benefit from autoloading and having commands to generate classes so I would think that should be a dedicated unit of work.

I think can start autoloading from here, as it will avoid upgrading indexer again. 

Command generator can come later

glennjacobs (approved):

PR #1227, opened 2023-08-29
Feat - Price formatter classes

alecritson:
This PR intends to move the price formatting into a separate class which you can swap out in the configuration.

Currently formatting of prices are tied to the Lunar implementation and we should be allowing developers to provide their own methods.

---

Currently I'm not too sure on the implementation 🤔 as it stands it's just a copy of the existing methods and moved to a class reference, but I'm not sure how correct...

### DIFF
--- a/docs/core/reference/pricing.md
+++ b/docs/core/reference/pricing.md
@@ -46,16 +46,6 @@ dealing with prices, other examples include:
 
 - `amount`
 
-```php
-$productVariant = \Lunar\Models\ProductVariant::first();
-
-// Lunar\Models\Price
-$priceModel = $productVariant->basePrices->first();
-
-// Lunar\DataTypes\Price
-$priceDataType = $priceModel->price;
-```
-
 ### `DefaultPriceFormatter`
 
 The default price formatter ships with Lunar and will handle most use cases for formatting a price, lets go through
--- a/docs/core/reference/pricing.md
+++ b/docs/core/reference/pricing.md
@@ -9,7 +9,7 @@ Every storefront is different. We understand as a developer you might want to do
 requirements, so we have made price formatting easy to swap out with your own implementation, but also we provide a
 suitable default that will suit most use cases.
 
-### Price formatting
+## Price formatting
 
 The class which handles price formatting is referenced in the `config/pricing.php` file:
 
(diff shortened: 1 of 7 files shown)


glennjacobs:
I think the function parameters probably need to be flexible, as if someone isn't using PHP's NumberFormatter class it wouldn't make sense.

alecritson:
Updated to decouple the arguments from the price datatype so it's up to the formatter itself what it supports. Also added a `prices` reference which should hopefully provide enough guidance 🤞

docs/core/reference/pricing.md
```diff
@@ -0,0 +1,210 @@
...
+The class which handles price formatting is referenced in the `config/pricing.php` file:
+
+```php
+return [
+    // ...
+    'formatter' => \Lunar\Pricing\DefaultPriceFormatter::class,
+];
+```
+
+When you retrieve a `Lunar\Models\Price` model, you will have access to the `->price` attribute which will return
```
glennjacobs, line 23: Is it worth noting here that this isn't the only place a `price` type will be encountered, e.g. order lines, carts, etc.

glennjacobs:
Do we need to add something to the upgrade guide for the config value?

docs/core/reference/pricing.md
```diff
@@ -0,0 +1,210 @@
...
+
+$priceDataType->unitFormatted(
+    locale: null, 
+    formatterStyle: NumberFormatter::CURRENCY,
+    decimalPlaces: null, 
+    trimTrailingZeros: true
+);
+```
+
+### Creating a custom formatter
```
glennjacobs, line 137: Can we use H2 headers where suitable so the TOC on the right shows the sections of the page?

glennjacobs (changes requested):
Left some comments for consideration

alecritson:
> Do we need to add something to the upgrade guide for the config value?

It's not a requirement to add it when updating, Lunar would still function. It would only be relevant if the developer was to change the class and they'd get that info from the other docs.

glennjacobs:
> > Do we need to add something to the upgrade guide for the config value?
> 
> It's not a requirement to add it when updating, Lunar would still function. It would only be relevant if the developer was to change the class and they'd get that info from the other docs.

No, but it is a change. Are we only showing breaking changes in the upgrade guide?

alecritson:
> > > Do we need to add something to the upgrade guide for the config value?
> > 
> > 
> > It's not a requirement to add it when updating, Lunar would still function. It would only be relevant if the developer was to change the class and they'd get that info from the other docs.
> 
> No, but it is a change. Are we only showing breaking changes in the upgrade guide?

My take is that the upgrade guide is for steps which are required for the upgrade to be successful. For example Laravel has this statement:

> We also encourage you to view the changes in the laravel/laravel [GitHub repository](https://github.com/laravel/laravel). While many of these changes are not required, you may wish to keep these files in sync with your application. Some of these changes will be covered in this upgrade guide, but others, such as changes to configuration files or comments, will not be.

PR #1229, opened 2023-08-30
Allow enabled or visible in customer group scope.

### LINKED ISSUE #1165: Discount is not applied in the cart since Lunar version 0.3
The table lunar_customer_group_discount contains an "enabled" column and a "visible" column. As far as I can see, the "visible" column is never set by the Admin Hub. 

It seems that "visible" and "enabled" are mutually exclusive, since the adminhub::partials.availability.customer-groups view will only ever show either the "enabled" toggle, or the "status" dropdown, but never both. The "status" dropdown is tied to the "visible" column.

The applyCustomerGroupScope method in the HasCustomerGroups trait applies both the "enabled"and the "visible" scopes, resulting in it never returning any values. Maybe it should instead check if either "enabled" OR "visible" is true.

- Lunar version: 0.4.0

alecritson:
Closes #1165

### DIFF
--- a/packages/core/src/Base/Traits/HasCustomerGroups.php
+++ b/packages/core/src/Base/Traits/HasCustomerGroups.php
@@ -82,7 +82,10 @@ trait HasCustomerGroups
             })->where(function ($query) use ($endsAt) {
                 $query->whereNull('ends_at')
                     ->orWhere('ends_at', '>=', $endsAt);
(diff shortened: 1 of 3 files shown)


glennjacobs (approved):

kallefrombosnia:
Discounts weren't working on 0.4 because of this, ty @alecritson 🙂

PR #1243, opened 2023-09-08
Fix dynamically applying discounts

theimerj:
I came across a situation when discounts were not applied, because the `calculate` method was already called on cart. It is called during `CartSession::use($cart)` and also during `CartSession::current()`. This ended up with a state in `DiscountManager`, when the `$discounts` property was set to empty collection. Then the check in DiscountManager became true every time, because it only checks for `null` value and not empty collection.

After changing the check to count with empty collection as follows: `! $this->discounts || $this->discounts?->isEmpty())`, there were multiple instances of the discounts added to `$cart->discounts` property. I mitigated this behavior by only adding discounts which are not already present in that property.

I also added a test which calls the `$cart->calculate()` method multiple times to make sure that discounts are added properly.

All tests are running green now, but this implementation is ofc open for discussion. Maybe there is a better solution.

### DIFF
--- a/packages/core/src/DiscountTypes/AmountOff.php
+++ b/packages/core/src/DiscountTypes/AmountOff.php
@@ -127,7 +127,9 @@ class AmountOff extends AbstractDiscountType
             $cart->discounts = collect();
         }
 
-        $cart->discounts->push($this);
+        if (! $cart->discounts->contains($this)) {
+            $cart->discounts->push($this);
+        }
 
         $this->addDiscountBreakdown($cart, new DiscountBreakdown(
             discount: $this->discount,
--- a/packages/core/src/Managers/DiscountManager.php
+++ b/packages/core/src/Managers/DiscountManager.php
@@ -183,7 +183,7 @@ class DiscountManager implements DiscountManagerInterface
 
     public function apply(Cart $cart): Cart
     {
-        if (! $this->discounts) {
+        if (! $this->discounts || $this->discounts?->isEmpty()) {
             $this->discounts = $this->getDiscounts($cart);
         }
 
--- a/packages/core/tests/Unit/DiscountTypes/AmountOffTest.php
+++ b/packages/core/tests/Unit/DiscountTypes/AmountOffTest.php
@@ -4,6 +4,7 @@ namespace Lunar\Tests\Unit\DiscountTypes;
 
 use Illuminate\Foundation\Testing\RefreshDatabase;
 use Lunar\DiscountTypes\AmountOff;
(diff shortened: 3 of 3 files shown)


glennjacobs:
This doesn't seem correct.

`CartSession::use($cart)` does NOT run the calculate method. 

The only issue I can see, which probably needs fixing, is the Cart discounts property is not being reset when calculating, which can lead to the same discount being logged twice. I believe it probably needs adding here https://github.com/lunarphp/lunar/blob/0.5/packages/core/src/Pipelines/Cart/ApplyDiscounts.php#L18

E.g.
```php
    public function handle(Cart $cart, Closure $next)
    {
        $cart->discounts = collect([]);  // <---------- this is new
        $cart->discountBreakdown = collect([]);
```

You probably need to update your test as follows

```php
        // Calculate method called for the first time
        CartSession::use($cart)->calculate();
```

glennjacobs (changes requested):
See comments

theimerj:
Hello @glennjacobs, I updated the test.

And yes, there were 2 problems.

1) Discounts could not be applied after first `calculate` call on Cart.
2) Discounts were added to `$cart->discounts` multiple times.

Let me know if I should do some more changes. Thank you.

packages/core/src/DiscountTypes/AmountOff.php
```diff
@@ -127,7 +127,9 @@ private function applyFixedValue(array $values, Cart $cart): Cart
             $cart->discounts = collect();
         }
 
-        $cart->discounts->push($this);
+        if (! $cart->discounts->contains($this)) {
```
glennjacobs, line 130: I think rather than this, we should be resetting the `$cart->discounts` when running the cart pipelines.

Updating `Pipelines/Cart/ApplyDiscounts.php` with the following seems to work

```php
    public function handle(Cart $cart, Closure $next)
    {
        $cart->discounts = null;
```

glennjacobs (changes requested):

ryanmitchell:
Sorry for spamming the comments, but I would love if if this could be merged for 0.6. The lack of a count()/empty() check on discounts is biting us at the moment so we have to manually add it.

theimerj:
> Sorry for spamming the comments, but I would love if if this could be merged for 0.6. The lack of a count()/empty() check on discounts is biting us at the moment so we have to manually add it.

Hey @ryanmitchell, could you please develop what you mean exactly? I could have another look. You can hit me up on Discord if you want.

glennjacobs (approved):

PR #1251, opened 2023-09-15
Add serialization to Price cast

### LINKED ISSUE #1250: Scout indexer not set for Order model
- Lunar version: dev-main (upcoming 0.6)
- Laravel Version: 10
- PHP Version: 8.2

### Expected Behaviour:
Scout should be able to index data for the Order model without error when using the database_index driver.

### Actual Behaviour:
The Order model no longer has specific mappings for Scout, so it's just using toArray().
There are two issues with this...

1) The Lunar\DataTypes\Price class is not serializable

2) The expected indexed data will not be the same as previous versions

### Proposed Solution

Make the Price type serializable and implement specific search indexers for all the models that had Scout mappings previously to retain consistency.

glennjacobs:
Closes #1250 

Currently, when converting a model to an array, Price objects will be in the array. Ideally, these should also be converted to arrays so that they can be indexed etc.

- [x] Add failing test
- [ ] Implement serialization
- [ ] Determine what to do with Livewire as it appears to be using the Arrayable data now

### DIFF
--- a/packages/core/src/Base/Casts/Price.php
+++ b/packages/core/src/Base/Casts/Price.php
@@ -3,11 +3,12 @@
 namespace Lunar\Base\Casts;
 
 use Illuminate\Contracts\Database\Eloquent\CastsAttributes;
+use Illuminate\Contracts\Database\Eloquent\SerializesCastableAttributes;
 use Illuminate\Support\Facades\Validator;
 use Lunar\DataTypes\Price as PriceDataType;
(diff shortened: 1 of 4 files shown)


glennjacobs:
@alecritson I feel like this might cause too much chaos to Livewire projects. Do you think we solve the immediate issue by simply implementing more Scout indexers for the different models?

glennjacobs:
Not happy with this direction, too risky.

PR #1252, opened 2023-09-15
Additional Scout Indexers

### LINKED ISSUE #1250: Scout indexer not set for Order model

glennjacobs:
Closes #1250 

Not having a suitable Order indexer caused issues with demo seeders and also broke compatibility with any existing index formats.


alecritson (changes requested):
Looks like we still have the old indexing properties/methods in the models:

https://github.com/lunarphp/lunar/blob/feature/more-scout-indexers/packages/core/src/Models/Order.php#L56

https://github.com/lunarphp/lunar/blob/feature/more-scout-indexers/packages/core/src/Models/Order.php#L69

https://github.com/lunarphp/lunar/blob/feature/more-scout-indexers/packages/core/src/Models/Order.php#L110

https://github.com/lunarphp/lunar/blob/feature/more-scout-indexers/packages/core/src/Models/Order.php#L302

https://github.com/lunarphp/lunar/blob/feature/more-scout-indexers/packages/core/src/Models/Order.php#L302

https://github.com/lunarphp/lunar/blob/feature/more-scout-indexers/packages/core/src/Models/ProductOption.php#L45

https://github.com/lunarphp/lunar/blob/feature/more-scout-indexers/packages/core/src/Models/ProductOption.php#L65

https://github.com/lunarphp/lunar/blob/feature/more-scout-indexers/packages/core/src/Models/ProductOption.php#L117

alecritson:
Correct, should be handled by the indexer, anything else should be removed if still there.

alecritson (approved):

PR #1259, opened 2023-09-21
Set section to nullable on attributes

netzknecht:
The table column `section` should be `nullable`, because it is not a required column e.G. model property. I think, it's only relevant if the hub is used but not required, too. Currently, it has to be set to nonsense or [space] string.

https://docs.lunarphp.io/core/reference/attributes.html#attributes-1

Field | Description
-- | --
section | An **optional** name to define where an attribute should be used.

### DIFF
--- a/packages/core/database/migrations/2023_09_21_100000_set_section_to_nullable_on_attributes.php
+++ b/packages/core/database/migrations/2023_09_21_100000_set_section_to_nullable_on_attributes.php
@@ -1,0 +1,22 @@
+<?php
+
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+use Lunar\Base\Migration;
+
+class SetSectionToNullableOnAttributes extends Migration
+{
+    public function up()
+    {
+        Schema::table($this->prefix.'attributes', function (Blueprint $table) {
+            $table->string('section')->nullable()->change();
+        });
+    }
+
+    public function down()
+    {
+        Schema::table($this->prefix.'attributes', function ($table) {
+            $table->string('section')->nullable(false)->change();
+        });
+    }
+}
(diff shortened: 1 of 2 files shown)


alecritson:
We should probably update the validation as well to allow null. https://github.com/lunarphp/lunar/blob/0.5/packages/admin/src/Http/Livewire/Components/Settings/Attributes/AttributeEdit.php#L79

netzknecht:
Done. That was a good example of why testing is so important. Obviously there is no such thing here and I'm currently only working with the core package.

netzknecht:
Hi @alecritson, would it be possible to merge this pull request or should I add a test for core and admin first?

I've also discovered some more sensible but missing default values for some database field definitions like "boolean" values. If I create migrations for it, should I separate them per schema or is it okay for you, to collect them into one file? At some point it will be time to squash migrations anyway.

packages/core/database/migrations/2023_09_21_100000_set_section_to_nullable_on_attributes.php
```diff
@@ -0,0 +1,22 @@
...
+            $table->string('section')->nullable()->change();
+        });
+    }
+
+    public function down()
+    {
+        Schema::table($this->prefix.'attributes', function ($table) {
+            $table->string('section')->nullable(false)->change();
+        });
+    }
```
wychoong, line 21: this probably shouldn't be needed

netzknecht: Yes, and I can skip something like that if desired? Principle, migrations should also be reversible. Couldn't find any code relating to that column/property, is there any future plan for it? If not, we can simple drop the column.

PR #1260, opened 2023-09-21
Fix phpunit coverage path and testsuites

netzknecht:
1. Fix the path defined for code coverage inside the `phpunit.xml` file of the monorepo [lunarphp/lunar](https://github.com/lunarphp/lunar).
2. Add all testsuites of the [lunarphp/core](https://github.com/lunarphp/core) repo into its `phpunit.xml` file.
3. Collect all testsuites into the `phpunit.xml` file from the monorepo [lunarphp/lunar](https://github.com/lunarphp/lunar).

### DIFF
--- a/packages/core/phpunit.xml
+++ b/packages/core/phpunit.xml
@@ -28,6 +28,9 @@
     <testsuite name="Database">
       <directory suffix="Test.php">./tests/Database</directory>
     </testsuite>
+    <testsuite name="Utils">
+      <directory suffix="Test.php">./tests/Utils</directory>
+    </testsuite>
   </testsuites>
   <php>
     <env name="DB_CONNECTION" value="testing"/>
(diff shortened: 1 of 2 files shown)


alecritson:
Thanks @netzknecht Can you just give a brief overview of what you're trying to solve here? I only ask because I noticed some files you're referencing don't contain tests, such as stubs.

netzknecht:
You are right @alecritson - rework is done.

netzknecht:
> Thanks @netzknecht Can you just give a brief overview of what you're trying to solve here?

I've build me a [local test env with docker](https://github.com/netzknecht/lunar/tree/feature-test-in-docker) and change this to be able to run the testsuites separate, like defined inside the separate packages.

alecritson (approved):

Issue #1267, opened 2023-09-23
Asking to add USD price even when it's not the default and active


amitavroy:
- Lunar version: 0.6.0
- Laravel Version: 10.10
- PHP Version: 8.1.10
- Database Driver & Version: Mysql 8.0

### Expected Behaviour:
After creating a new current INR and enabling it, I disabled USD.
Then I try to create a product with the intention to add only INR price.

I was going through this issue #502 and the solution by Adam works however, it seems the latest version doesn't have this code. So, raising a PR

### Actual Behaviour:
I am still asked to add the USD price. The validation error doesn't allow me to save the product with just INR price.

### Steps To Reproduce:
1. Add new currency (INR in my case)
2. Enable the new current and make it default
3. Disable USD currency
4. Try to create product

![lunar_ss](https://github.com/lunarphp/lunar/assets/1438890/64b2d314-f80c-4d28-85e5-6ea63b322366)

glennjacobs:
Disabling a currency is for the frontend. If you don't want to add prices, you will need to remove the currency.

glennjacobs:
Solving this properly would require the system to delete all prices from products, orders etc. potentially a big task and highly risky, hence we don't offer a delete option currently. 

To solve the problem it would probably have to run as a background task and disable the hub whilst running. 

It's a lot of effort for a task that would rarely be used.

glennjacobs:
I get that devs will want to change the currency on first install. Maybe there's a better way. But this isn't a bug as such, so an enhancement suggestion should be discussed in GitHub Discussions.

PR #1268, opened 2023-09-23
Fixes #502 and #1267

### LINKED ISSUE #502: multiple currencies and disabled
- GetCandy version: 2.0-beta15

### Expected Behaviour:
when disable currency, it should be disabled and won't be required when filling product info

### Actual Behaviour:
validation fails when save product for disabled currency
image

### Steps To Reproduce:
- more than 1 currency
- disable 1 currency
- fill prices except disabled currency

@glennjacobs mentioned on discord this could be:
1. on intention to avoid missing prices when re-enable the currency
2. for controlling currencies on frontend

for 1. i think it's valid but also at the same time the same issue will happen for adding new currency to the store. for 2., maybe a clearer label needed

amitavroy:
This pull request tries to fix issue #502 and also the issue I have raised i.e. #1267 
The solution for #502 kind of worked for my problem as well to add a check of the active currencies when creating a new product.

### DIFF
--- a/packages/admin/src/Http/Livewire/Traits/HasPrices.php
+++ b/packages/admin/src/Http/Livewire/Traits/HasPrices.php
@@ -197,7 +197,7 @@ trait HasPrices
      */
     public function getCurrenciesProperty()
     {
-        return Currency::get();
+        return Currency::whereEnabled(true)->get();
     }
 
     /**


amitavroy:
Strange. 

I raised a properly documented issue. Went through other issues to see if there exist a solution
and based on that, raise a PR to solve the issue.

If you are closing the issue at least comment on why you feel this is not proper. I genuinely wanted to contribute and wanted to help solve that issue.

@glennjacobs

glennjacobs:
> Strange. 
> 
> 
> 
> I raised a properly documented issue. Went through other issues to see if there exist a solution
> 
> and based on that, raise a PR to solve the issue.
> 
> 
> 
> If you are closing the issue at least comment on why you feel this is not proper. I genuinely wanted to contribute and wanted to help solve that issue.
> 
> 
> 
> @glennjacobs 

I have on your referenced issue quite comprehensively.

amitavroy:
Yes, I understood your stand for my issue of #1267. So, for now I guess I will work with that solution.

However, the PR also sovles #502 

If you are solving that also with some other approach, then yes I would be fine.

glennjacobs:
I will have to check that old issue, but I don't believe this is the solution.

PR #1269, opened 2023-09-23
Check if the user is null when logging out

dewebdesigns:
Laravel Dusk passes null to the $event->user, so it throws an error. Adding a simple null check fixes this.


alecritson:
Thanks @dewebdesigns I feel like this would benefit from being in `0.6`, can you change the target branch please :)

dewebdesigns:
@alecritson I have updated the target branch to 0.6. Thank you!

alecritson (approved):

PR #1273, opened 2023-09-25
Format config files more in line with the Laravel skeleton

adevade:
- Empty line at the start and end of main array
- Empty line between all top level keys
- No leading backslash when referencing classes
- Import "one-off" classes 
- Do not import classes part of an array list, like `pipelines`
- Fix indentation on some comments
- Add example classes with suggested namespace where no defaults exist
- Change formatting of "in-array comments"

---

Nothing has been changed except formatting.
For reference: [laravel/laravel config folder](https://github.com/laravel/laravel/tree/10.x/config)


alecritson:
Thanks @adevade did you use a auto formatter to do this or was it manual? My only thought is if we did this could it slip back to the old formatting unless we can pick it up in something like pint?

adevade:
@alecritson Nope, unfortunately this was manual work. Don't know if there's a formatter available for this?

alecritson:
Hmm yeah it doesn't look like there is anything and I can't see any actions Laravel have set up for it. We'll get this change in though thanks and hopefully keep an eye on it going forward.

alecritson (approved):

PR #1276, opened 2023-09-27
Fix github action "Split Packages" on forked monorepo

netzknecht:
Adds variables for package to repository mapping inside `.github/workflows/split_packages.yml`, to allow package splitting on a forked monorepo, with deviating repository names.

Available variables:

- SPLIT_PACKAGES (type: JSON array literal as string; default: ["admin", "core", "licensing"])
- SPLIT_PACKAGES_REPOSITORIES (type: false|JSON array literal as string; default: false)
- SPLIT_UTILS  (type: JSON array literal as string; default: ["livewire-tables", "scout-database-engine"])
- SPLIT_UTILS_REPOSITORIES: (type: false|JSON array literal as string; default: false)

If the *_REPOSITORIES variable is set to false, the name of the packages/utils get used as repository name.

### DIFF
--- a/.github/workflows/split_packages.yml
+++ b/.github/workflows/split_packages.yml
@@ -1,4 +1,9 @@
 name: Split Packages
+env:
+  SPLIT_PACKAGES: ${{ vars.SPLIT_PACKAGES || '["admin", "core", "licensing"]' }}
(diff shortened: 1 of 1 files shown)


netzknecht:
New version here #1277.

PR #1285, opened 2023-10-05
Allow discounts to be limited to product variants

ryanmitchell:
This PR adds the ability to limit discounts to product variants with the corresponding UI changes required in the hub. It also adds a new `product variant search` component to the hub.


alecritson:
Thanks @ryanmitchell, I like this idea and had a play around and seems to work. After chatting with @glennjacobs we feel this should be targeting `main` for release in `0.7` since it's more of a feature and includes functionality, rather than a fix.

ryanmitchell:
No problem. Need me to change anything?

alecritson:
@ryanmitchell I don't think so from initial testing, only thing I was looking at was the nested sub query and whether it could be flattened a bit, but hadn't had a real look.

alecritson (approved):

PR #1288, opened 2023-10-06
Add the ability to exclude brands, collections, products and variants from discounts

ryanmitchell:
This PR extends the 'limitations' area of discounts to also allow items to be excluded from the discount:

Screenshot 2023-10-06 at 07 46 55

For example, this allows you to add a collection or brand, but exclude certain products from that brand.

Requires https://github.com/lunarphp/lunar/pull/1285 to be merged first


ryanmitchell:
Any thoughts on this one?

alecritson (approved):
Thanks @ryanmitchell had a play around with this and it seems pretty useful, will get it ready for the next release.

PR #1294, opened 2023-10-06
Add options to lunar:install and lunar:hub:install to allow for non-prompting usage

charlielangridge:
This updates to the command lunar:install allows options to be used to:
--retainConfig - if a config already exists, this option retains the exisiting config without prompting
--migrate - this runs the database migrations without prompting
--firstName - this option allows the input of new admin user first name
--lastName - this option allows the input of new admin user last name
--email - this option allows the input of new admin user email name
--password - this option allows the input of new admin user password name

This updates to the command lunar:hub:install allows options to be used to:
--firstName - this option allows the input of new admin user first name
--lastName - this option allows the input of new admin user last name
--email - this option allows the input of new admin user email name
--password - this option allows the input of new admin user password name


wychoong:
is there use case that you can share for using this command programmatically?

wychoong:
@glennjacobs i think the pint action is not running the check as the dirty flag only look for uncommitted files, not changed files in PR, see https://github.com/aglipanci/laravel-pint-action/issues/4#issuecomment-1475163156

charlielangridge:
@wychoong Our use case is that we have created a custom fresh setup command that migrates fresh, runs the installer and then imports a load of products - this allows us to rapidly deploy without having prompts (specifically having to type in the user details each time)

charlielangridge:
Another thing to add is the alternative in our use case would be to copy the install commands and make adjustments to our installations, but then if there's any updates to the install code in future, our custom install command  obv wouldn't be correct

glennjacobs:
> @glennjacobs i think the pint action is not running the check as the dirty flag only look for uncommitted files, not changed files in PR, see [aglipanci/laravel-pint-action#4 (comment)](https://github.com/aglipanci/laravel-pint-action/issues/4#issuecomment-1475163156)

Oh, that's a bit silly. I'll look at changing that in a different PR.

glennjacobs:
@charlielangridge can you please target `main` as this isn't a bug fix.

charlielangridge:
@glennjacobs - that's been done

glennjacobs:
I am going to close this, as we're not looking to add new functionality to v0.8 now. However, we do want to incorporate this into v1.0 so I will be raising an issue to consider this at the point we review the installer commands.

PR #1303, opened 2023-10-11
Add command for cart pruning

ryanmitchell:
This PR adds a command `lunar:prune:carts` and some new config params under `lunar.cart.prune_tables` to determine how pruning should be carried out.

I opted for using a pipeline for amending the query, as we ultimately always seem to end up there.

I've tried to default to some sensible settings (90 days, dont prune carts with orders associated). 

If pruning is enabled we automatically add a daily task rather then needing the developer to do it.

### DIFF
--- a/packages/core/config/cart.php
+++ b/packages/core/config/cart.php
@@ -162,7 +162,7 @@ return [
     */
     'prune_tables' => [
         
-        'enabled' => true,
+        'enabled' => false,
         
         'pipelines' => [
             Lunar\Pipelines\CartPrune\PruneAfter::class,
--- a/packages/core/config/cart.php
+++ b/packages/core/config/cart.php
@@ -150,4 +150,26 @@ return [
         'lines.purchasable.product',
         'lines.cart.currency',
     ],
+    
+    /*
+    |--------------------------------------------------------------------------
+    | Prune carts
+    |--------------------------------------------------------------------------
+    |
+    | Should the cart models be pruned to prevent data build up and 
+    | some settings controlling how pruning should be determined
+    |
+    */
+    'prune_tables' => [
+        
+        'enabled' => true,
+        
+        'pipelines' => [
+            Lunar\Pipelines\CartPrune\PruneAfter::class,
+            Lunar\Pipelines\CartPrune\WithoutOrders::class,
+        ],
+        
+        'prune_interval' => 90, // days
+        
+    ],
 ];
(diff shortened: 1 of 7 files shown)


packages/core/config/cart.php
```diff
@@ -150,4 +150,26 @@
...
+    | Prune carts
+    |--------------------------------------------------------------------------
+    |
+    | Should the cart models be pruned to prevent data build up and 
+    | some settings controlling how pruning should be determined
+    |
+    */
+    'prune_tables' => [
+        
+        'enabled' => true,
```
glennjacobs, line 165: I feel like this should default to `false` as someone could upgrade and not realise it's happening.

ryanmitchell: If you check the schedule you see I default to false there.

So new installs -> true
Old installs -> false

glennjacobs:
@ryanmitchell thanks for this. I think we need some documentation somewhere for this and also a test, as we don't want to make any mistakes with deleting carts.

ryanmitchell:
Yep sorry I meant to ask where you'd want the docs to go - I wasnt sure where it best fitted.

packages/core/config/cart.php
glennjacobs: I still don't think we should be pruning by default. Safer to have it `false` for all installs.

glennjacobs:
> Yep sorry I meant to ask where you'd want the docs to go - I wasnt sure where it best fitted.

It's a good point. I had started updated docs for this config area, I could do with merging what I've done so far. I'll review and get back to you.

wychoong:
Do you think create 2 commands and a documentation to add the command into app console kernal gives more flexibility to control how frequent it should be called. 

The pipelines can be use to customize (eg: query cart older than X days) and sensible defaults being provided by lunar

also the query should be optimized with cursor fetch or paginated, and wrapped in DB::transaction

ryanmitchell:
Any thoughts on this one? Im happy to add to the docs if you point me in the right direction.

glennjacobs:
> Any thoughts on this one? Im happy to add to the docs if you point me in the right direction.

I would probably suggest adding something to <https://docs.lunarphp.io/core/reference/carts.html>

ryanmitchell:
Thanks @glennjacobs. I've added some docs now.

glennjacobs (approved):

PR #1312, opened 2023-10-19
Make explicit binding for products conditional in hub context

repl6669:
While developing [lunar-api](https://github.com/dystcz/lunar-api), specifically implementing the possibility of using hashids, I came across an issue with explicit model binding for products. 

The underlying packages encodes and decodes route keys on runtime. The login can be seen [here](https://github.com/laravel-json-api/hashids/blob/develop/src/HashId.php#L128-L141). When `AdminHubServiceProvider` binds the `product` explicitly, the route key does not get decoded.

It is also quite difficult to overwrite such binding by other packages. You would have to ensure that the order of service providers is correct in order to "rebind" the routes.

These are the problematic lines:

https://github.com/lunarphp/lunar/blob/d7b6eac073d3104c1dbd462f779fcfc0d88396c5/packages/admin/src/AdminHubServiceProvider.php#L221-L223

I came up with a simple solution for now:

https://github.com/dystcz/lunar/blob/9eba7aa9e45f19d6d08c8a08f7f55656df213666/packages/admin/src/AdminHubServiceProvider.php#L221-L227

This applies the binding only for routes with hub prefix.

Another solutions would be to implement the binding via middleware or by overwriting the `resolveRouteBinding` method on the model.


alecritson (commented):
Thanks @theimerj and while I can see there is an issue here I don't think this is correct, when doing some initial testing and editing a product, `$route->getPrefix()` returns `hub/products/{product}` so the condition wouldn't match.

I think we'd need a more standard approach to determine whether it's a hub request, not just by going off the configured hub path. 🤔

ryanmitchell:
Check if its using the hub middleware?

glennjacobs:
@theimerj could you retarget the `0.7` branch, please?

repl6669:
Hey @glennjacobs, I retargeted to `0.7` and refactored it to check against Hub `Authenticate` middleware based on @ryanmitchell's idea. I also added a test which proves that soft deleted products can be accessed in the hub.

If you know about some better way how to determine that the route belongs to hub scope, let me know, i will update the code accordingly.

alecritson (approved):

PR #1314, opened 2023-10-20
Fix unit decimal calculation

alecritson:
Currently when calculating the cart lines via the pipelines, we're getting the unit cost based on the unit quantity but then also applying the same unit quantity to the cart line, even though it's already been considered. The result is if a purchasable has a unit quantity over 1 it's being doubled e.g.

```php
ProductVariant::create([
    // ...
    'unit_quantity' => 100,
]);

$price = Price::create([
   'price' => 5503
]);

$price->price->unitDecimal; // 0.55

// After $cart->calculate();

$cartLine->subTotal->unitDecimal; // 0.01
```

This PR sets the `unitQuantity` to `1` on the `$cartLine` price data types, as the unit quantity has already been factored in.

### DIFF
--- a/packages/core/src/Pipelines/Cart/CalculateLines.php
+++ b/packages/core/src/Pipelines/Cart/CalculateLines.php
@@ -26,19 +26,16 @@ class CalculateLines
 
                     return $cartLine;
                 });
-
-            $purchasable = $cartLine->purchasable;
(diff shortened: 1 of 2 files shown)


glennjacobs:
@alecritson shouldn't this target 0.6?

glennjacobs (approved):

PR #1316, opened 2023-10-27
Add ShippingOption override to cart

alecritson:
This PR looks to add the ability to carts for developers to override a shipping option for a cart in a single request, useful if you want to estimate what the total cost a cart will have with a certain shipping option selection, to present to the end user.

### Usage

You can get an estimated shipping option for the cart like so:

```php
$cart->getEstimatedShipping([
    'postcode' => '123456',
    'state' => 'Essex',
    'country' => Country::first(),
]);
```

This will fetch the cheapest shipping option from the manifest, but by default will not set the shipping override, so pipelines will function as they currently do. In order for the pipelines to take the override into account you just pass the additional `setOverride` parameter:

```php
$cart->getEstimatedShipping([
    'postcode' => '123456',
    'state' => 'Essex',
    'country' => Country::first(),
], setOverride: true);
```

Now when you call `calculate()` on a cart, the override will be taken into account, bypassing any shipping address logic you may have in place.

This is a "once" off effect so, say you have another request that calls `calculate()` without setting the meta above, the pipeline will not take it into account.

If you use the CartSession facade, this can remember the meta for you in the session, making subsequent calls easier:

```php
CartSession::estimateShippingUsing([
    'postcode' => '123456',
    'state' => 'Essex',
    'country' => Country::first(),
]);
```

You may not need to provide all arguments, but may affect how shipping options are resolved and which are added to the manifest.

Once set though, you can pass the optional `estimateShipping` parameter when getting the current cart.

This will use the shipping override in the pipelines, based on the 
```php
CartSession::current(estimateShipping: true);
```

Behind the scenes, we're just using the `getEstimatedShipping` on the cart as we fetch it and setting the override before the `calculate()` method is called.

### DIFF
--- a/docs/core/reference/carts.md
+++ b/docs/core/reference/carts.md
@@ -339,7 +339,7 @@ CartSession::setCustomer($customer);
 ```
 
 
-### Adding shipping/billing address
+## Adding shipping/billing address
 
 As outlined above, you can add shipping / billing addresses to the cart using
 the following methods:
--- a/docs/core/reference/carts.md
+++ b/docs/core/reference/carts.md
@@ -376,6 +376,50 @@ $cart->shippingAddress;
 $cart->billingAddress;
 ```
 
+### ShippingOption override
+
+In some cases you might want to present an estimated shipping cost without users having to fill out a full shipping address, this is where the `ShippingOptionOverride` comes in, if set on the cart it can be used to calculate shipping for a single request.
+
+```php
+$shippingOption = $cart->getEstimatedShipping([
+    'postcode' => '123456',
+    'state' => 'Essex',
+    'country' => Country::first(),
+]);
+````
+
+This will return an estimated (cheapest) shipping option for the cart, based on it's current totals. By default this will not be taken into account when calculating shipping in the cart pipelines, in order to enable that we need to pass an extra parameter.
+
+```php
+$shippingOption = $cart->getEstimatedShipping([
+    'postcode' => '123456',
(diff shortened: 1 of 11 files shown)


glennjacobs:
Nice. I think it just needs some docs adding.

glennjacobs:
Can we change this to an H2 in the docs so the shipping header shows on the right?

```
### Adding shipping/billing address
```

docs/core/reference/carts.md
```diff
@@ -376,6 +376,50 @@ $cart->shippingAddress;
 $cart->billingAddress;
 ```
 
+### ShippingOption override
```
glennjacobs, line 379: Should be a capital "O" on override.

glennjacobs: Actually, don't worry. The whole file needs sorting.

packages/core/src/DataTypes/ShippingOption.php
```diff
@@ -16,6 +16,7 @@ public function __construct(
         public TaxClass $taxClass,
         public $taxReference = null,
         public $option = null,
+        public bool $collection  = false,
```
glennjacobs, line 19: I just spotted this was added. I presume it's for orders that are being collected and not delivered.

Can we add the usage to the docs? Maybe also at the same time give an example of how you can manually override the shipping option directly on the cart too?

glennjacobs (dismissed):

glennjacobs (approved):

PR #1318, opened 2023-10-30
Tweak tax breakdown state update

alecritson:
There was an issue where the tax breakdown state update was only partially running or not at all when it should be. This PR looks to improve the state update by trying to fetch all relevant rows between orders and lines, not just base the update on whether the first record has been updated (which didn't help if the data had partially updated)

The other change is to move away from Eloquent and use the DB facade, mainly to help with larger datasets and memory usage.

### DIFF
--- a/packages/core/database/state/ConvertTaxbreakdown.php
+++ b/packages/core/database/state/ConvertTaxbreakdown.php
@@ -3,10 +3,7 @@
 namespace Lunar\Database\State;
 
 use Illuminate\Support\Facades\Schema;
-use Lunar\Base\ValueObjects\Cart\TaxBreakdownAmount;
-use Lunar\DataTypes\Price;
-use Lunar\Models\Order;
-use Lunar\Models\OrderLine;
+use Lunar\Facades\DB;
 
 class ConvertTaxbreakdown
 {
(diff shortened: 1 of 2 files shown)


packages/core/database/state/ConvertTaxbreakdown.php
```diff
@@ -17,74 +14,80 @@ public function prepare()
...
 
-        if (!$breakdownItem) {
-            return false;
-        }
+    protected function canRunOnTable(string $table)
+    {
+        $prefix = config('lunar.database.table_prefix');
 
-        return $breakdownItem['total'] ?? false;
+        return Schema::hasTable("{$prefix}{$table}");
```
glennjacobs, line 91: Would this not mean it will run every time you run migrations? Which on a large dataset could be an issue.

alecritson: Yeah this is a bit of a curse with the state classes, but otherwise, it's difficult to determine whether or not it should be run if there are partial orders/order lines that don't have the updated tax breakdown

When testing on a site with 2mil + order lines and 500k+ orders, once it had been run initially it takes around 2 seconds for the check.

It's something that really should be in an update command I think personally, but we don't have that luxury right now.

glennjacobs: ok, if that's true then I think it can go in. And we can look to implement an upgrade command down the line.

glennjacobs (approved):

PR #1320, opened 2023-10-31
Fixes storage disk when updating state

mattdfloyd:
Instead of using the default driver, this updates the task to always use the local disk for storing tmp json files. One of the Storage already explicitly [uses the local disk](https://github.com/lunarphp/lunar/blob/0.6/packages/core/database/state/EnsureBrandsAreUpgraded.php#L66).

My default disk is an s3 disk. Instead of getting a `FileNotFoundException`, an `UnableToReadFile` exception is thrown instead when attempting to [get the file](https://github.com/lunarphp/lunar/blob/0.6/packages/core/database/state/EnsureBrandsAreUpgraded.php#L47-L49). We could additionally catch this exception (or any exception) but I think it's more reasonable to only use the local disk for storing the temporary state - especially considering there's already an explicit call to the local disk.

### DIFF
--- a/packages/core/database/state/EnsureBrandsAreUpgraded.php
+++ b/packages/core/database/state/EnsureBrandsAreUpgraded.php
@@ -33,7 +33,7 @@ class EnsureBrandsAreUpgraded
             $brands[$brand][] = $productId;
         }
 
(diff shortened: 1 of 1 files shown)


alecritson (approved):
Thanks @mattdfloyd seems like a reasonable change, will get this in!

PR #1329, opened 2023-11-10
Avoid deprecation notices on update to lunar 0.7

netzknecht:
${var} string interpolation is deprecated since php 8.2

https://wiki.php.net/rfc/deprecate_dollar_brace_string_interpolation


alecritson:
Thanks @netzknecht this has already been done with #1323 :)

netzknecht:
> Thanks @netzknecht this has already been done with #1323 :)

But with #1325 it is redone partly!

netzknecht:
... so I created another pull request #1333.

PR #1330, opened 2023-11-10
Default values on `attribute_groups` and `attributes`

netzknecht:
Add default values on `attribute_groups` and `attributes` table.


packages/core/database/migrations/2023_09_28_100000_set_default_values_on_attributes_table.php
```diff
@@ -0,0 +1,27 @@
...
+use Lunar\Base\Migration;
+
+class SetDefaultValuesOnAttributesTable extends Migration
+{
+    public function up()
+    {
+        Schema::table($this->prefix . 'attributes', function (Blueprint $table) {
+            $table->boolean('required')->default(0)->change();
+            $table->boolean('system')->default(0)->change();
+            $table->boolean('position')->default(0)->change();
```
alecritson, line 14: I'm not sure on having `position` default to `0`, because you could end up with a bunch of attributes or groups with that position. I feel like it should be passed through. @glennjacobs What do you think?

glennjacobs: We'd never want `0` to be stored, so this could cause problems. I don't think there's value in adding it.

netzknecht:
@glennjacobs  and @alecritson please check out and review the changes of my last commits regarding your previous concerns. Thank you!

glennjacobs:
This PR seems to have lots of updates not associated to the functionality specified. Closing, but feel free to reopen if you can make a PR that is focused on a single issue.

PR #1331, opened 2023-11-10
Wrong unique constraint on attribute groups

netzknecht:
Define unique constraint on attribute groups with `attributable_type` and `handle` to allow same handle on different attributable types.

### DIFF
--- a/packages/admin/resources/views/livewire/components/settings/attributes/show.blade.php
+++ b/packages/admin/resources/views/livewire/components/settings/attributes/show.blade.php
@@ -146,7 +146,6 @@
         <x-slot name="title">{{ __('adminhub::components.attributes.show.create_title') }}</x-slot>
         <x-slot name="content">
             @livewire('hub.components.settings.attributes.attribute-group-edit', [
-                'typeHandle' => $type,
                 'attributableType' => $this->typeClass,
             ])
         </x-slot>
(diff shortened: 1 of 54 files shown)


glennjacobs:
@alecritson this could be OK. I'm trying to remember back to when we first made this. Do we need hub updates to go alongside this?

netzknecht:
@glennjacobs due to the way in which the `handle` was automatically generated from `attributable_type` and `name` when storing `AttributeGroup` models in the hub, the problem of the incorrectly defined unique index has probably not been noticed yet. 

Because I created the `AttributeGroup` models via code or seeder, the automatic generation of the `handle` inside the Livewire component was of course not effective and so I run into this issue.

With the last commit, the handle in the hub is now generated exclusively from the `name` and, due to the adjustment of the unique index, the same `handle` can be used multiple times if the `attributable_type` is different.

netzknecht:
So far there is no practical need to fetch`AttributeGroup` models based on their `handle`. I think there is no reason why to place this change in the upgrade guide.

However, the old method would currently have to be:
```
// Old
AttributeGroup::whereHandle(Str::handle(Product::class).'_foo')->first();
```

glennjacobs:
> So far there is no practical need to fetch`AttributeGroup` models based on their `handle`. I think there is no reason why to place this change in the upgrade guide.
> 
> However, the old method would currently have to be:
> 
> ```
> // Old
> AttributeGroup::whereHandle(Str::handle(Product::class).'_foo')->first();
> ```

I think the issue is that someone might be, hence we should document it in the upgrade guide.

glennjacobs:
Also, could this please target the `0.8` branch?

glennjacobs:
This PR seems to have lots of updates not associated to the functionality specified. Closing, but feel free to reopen if you can make a PR that is focused on a single issue.

PR #1335, opened 2023-11-10
Enhancement `HasDefaultRecord`-trait

netzknecht:
- All models that use a `default` field/record, now use the `HasDefaultRecord` trait.
- Add `default` field/record and `HasDefaultRecord` trait to `ProductType` and `CollectionGroup` .
- Cache with `Blink::once()` in `HasDefaultRecord::getDefault()` only if query returns a model.
- Remove some redundant code.

### DIFF
--- a/packages/core/database/migrations/2023_10_10_100000_add_default_to_collection_groups_table.php
+++ b/packages/core/database/migrations/2023_10_10_100000_add_default_to_collection_groups_table.php
@@ -1,0 +1,22 @@
+<?php
+
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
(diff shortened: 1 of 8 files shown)


netzknecht:
I realized that there are some redundant observers in `Lunar\Observers` namespace, if the trait is used but also, that there are some special cases on determining the presens of only one default. But I think its better to keep most code in the trait and implement a overwriteable method to handle such special cases on model level.

glennjacobs:
@netzknecht are you intending to finish this PR, or should it be ready for review now? Thanks.

glennjacobs:
Closing as this PR has had a status of "Draft" for a long time and our focus is now moving to 1.x.

PR #1336, opened 2023-11-11
Updating type annotations in models and traits

ILDaviz:
Hi, I refactored the model and trait section. Because I can't continue with the graphql module development if I don't fix the types in the existing models. I updated all the models present and in some traits where I needed.
Let me know if this is okay.

Best regards and congratulations on the project.

### DIFF
--- a/packages/core/src/Models/Product.php
+++ b/packages/core/src/Models/Product.php
@@ -9,6 +9,7 @@ use Illuminate\Database\Eloquent\Relations\BelongsToMany;
 use Illuminate\Database\Eloquent\Relations\HasMany;
 use Illuminate\Database\Eloquent\Relations\HasManyThrough;
 use Illuminate\Database\Eloquent\Relations\MorphMany;
+use Illuminate\Database\Eloquent\Relations\MorphToMany;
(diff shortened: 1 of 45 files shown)


packages/core/src/Models/Product.php
```diff
@@ -79,50 +82,40 @@ protected static function newFactory(): ProductFactory
 
     /**
      * Returns the attributes to be stored against this model.
-     *
-     * @return array
      */
-    public function mappedAttributes()
+    public function mappedAttributes(): array
```
alecritson, line 86: I think the docblock was incorrect here, as it's referencing a relationship I believe it should be `Collection` that is returned.

ILDaviz: Wooops I missed it.

alecritson (approved):
Thanks for this @ILDaviz ! :)

ILDaviz:
If you need more help typing the rest of the project just ask. 🚀

sanneterpstra:
When using the mappedAttributes method I now get an error:

`Lunar\Models\Product::mappedAttributes(): Return value must be of type Lunar\Models\Collection, Illuminate\Database\Eloquent\Collection returned`

PR #1337, opened 2023-11-13
Enhancement/translations

netzknecht:
Commit 77673f0f4712d8fa70658fbebe0158feb3fb5773 adds two scopes `translation` and `attributeTranslation` to constraint queries on translated model attributes and lunar attributes/fields.

Commit 438f915b73e31c3b5bf86d7b421038c26a1a5943 adds a `AsTranslated` cast, applies it to all relevant models and refactors affected code.

### DIFF
--- a/packages/admin/resources/views/partials/products/variants/basic-information.blade.php
+++ b/packages/admin/resources/views/partials/products/variants/basic-information.blade.php
@@ -9,8 +9,8 @@
       <div class="space-y-4">
 
         @foreach($variant->values as $value)
-          <x-hub::input.group :label="$value->option->name->en" for="sku">
-            {{ $value->name->en }}
+          <x-hub::input.group :label="$value->option->translate('name')" for="sku">
+            {{ $value->translate('name') }}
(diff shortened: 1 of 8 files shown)


alecritson:
@netzknecht Can you explain what this is attempting to fix and is this something that is related to a discussion?

netzknecht:
Hi @alecritson, 

this is not a fix, it's an enhancement or extension. 

Currently, lunar uses translations as json data in regular laravel database attributes/fields and in lunar attributes, too.

The new scopes in commit #77673f0 allows to query a model for a specific translation in both cases. That can be useful if you connect an external service and you must check the existence of a model by the translation.

The benefit of the second commit #438f915 is, that translated laravel database attributes are more strictly casted "AsTranslated" and all relevant accessor and mutator logic are at one place in the "AsTranslated" castable class. In addition, the accessor and mutator of the "AsTranslated" class converts everything that is no an array or collection into a collection with the current locale as locale key.

glennjacobs:
This isn't something we want to incorporate.

PR #1338, opened 2023-11-13
Implement paratest for local testing

netzknecht:
It will be nice to have "paratest" as dev dependency by default for local testing, even if it's not used with github workflows #1265.


alecritson:
I would say we want to avoid adding dependencies that Lunar isn't using. If it's something you use yourself and want to use it across projects, maybe look at installing it globally.

PR #1349, opened 2023-11-30
Update dotfiles

adevade:
References:

https://github.com/laravel/framework/blob/10.x/.gitattributes

https://github.com/laravel/laravel/blob/10.x/.gitattributes

https://github.com/livewire/livewire/blob/main/.gitattributes

https://github.com/filamentphp/filament/blob/3.x/.gitattributes


glennjacobs:
@adevade Should this PR be ready for review?

adevade:
@glennjacobs Sorry, I got busy 😅 

I was wondering if maybe the dotfiles in the separate packages should be updated as well? I'm not too familiar with how the splitting of the mono-repo works...

glennjacobs:
> @glennjacobs Sorry, I got busy 😅
> 
> I was wondering if maybe the dotfiles in the separate packages should be updated as well? I'm not too familiar with how the splitting of the mono-repo works...

Generally speaking, when developing you do so on the monorepo, so we may not even need them in each package.

adevade:
Okay, so development files like `.editorconfig` is not needed in those folder. But files like `.gitattributes` should still be updated I think. 

If someone is `composer require`-ing just the `core` package, then stuff like the `.github` folder and the tests would not have to be pulled down to their app.

glennjacobs:
> Okay, so development files like `.editorconfig` is not needed in those folder. But files like `.gitattributes` should still be updated I think.
> 
> If someone is `composer require`-ing just the `core` package, then stuff like the `.github` folder and the tests would not have to be pulled down to their app.

Yes true. It will be different in v1.0, but for now I agree.

glennjacobs:
I think this is looking good. Let me know when it's ready for review.

adevade:
I think this is ready to be reviewed! Should it still target the 0.7 branch?

glennjacobs:
> I think this is ready to be reviewed! Should it still target the 0.7 branch?

Would it be OK to target `0.8` please?

adevade:
Changed branch to `0.8` and pulled in latest changes 👍

alecritson (approved):

PR #1351, opened 2023-12-01
Enhancement `HasPosition` trait

netzknecht:
Adds a `HasPosition` trait for models that require a (manual) sort key, using a numeric `position` column in the model's table.

**Features:**
- [x] Automatic assignment of the next sort key value if not explicitly defined
- [x] Automatic checking and, if necessary, correction of invalid sort key values
- [x] Limiting the validation and correcting the sort key values ​​based on the data of other columns in the table
- [x] `Blueprint` macro to add `position` columns with the required unique constriants in migration files
- [x] Convert current position columns and values ​​to work with the trait

**To do:**
- [ ] Add tests
- [ ] If necessary, consider `position` columns in `pivot` tables.
- [ ] Documentation

### DIFF
--- a/packages/admin/src/Http/Livewire/Components/Settings/Attributes/AttributeEdit.php
+++ b/packages/admin/src/Http/Livewire/Components/Settings/Attributes/AttributeEdit.php
@@ -181,9 +181,6 @@ class AttributeEdit extends Component
         if (! $this->attribute->id) {
             $this->attribute->attribute_type = $this->group->attributable_type;
             $this->attribute->attribute_group_id = $this->group->id;
-            $this->attribute->position = Attribute::whereAttributeGroupId(
-                $this->group->id
-            )->count() + 1;
             $this->attribute->save();
             $this->notify(
                 __('adminhub::notifications.attribute-edit.created')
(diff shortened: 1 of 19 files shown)


netzknecht:
@glennjacobs and @alecritson, before I spend more time into this, what do you think about using the [spatie/eloquent-sortable](https://github.com/spatie/eloquent-sortable) package for manual sorting instead. It should cover all requirements for lunar and works well with the upcoming admin ui based on [filament](https://v2.filamentphp.com/tricks/sorting-table-with-spatie-eloquent-sortable), even if the example is for filament version 2.

glennjacobs:
> @glennjacobs and @alecritson, before I spend more time into this, what do you think about using the [spatie/eloquent-sortable](https://github.com/spatie/eloquent-sortable) package for manual sorting instead. It should cover all requirements for lunar and works well with the upcoming admin ui based on [filament](https://v2.filamentphp.com/tricks/sorting-table-with-spatie-eloquent-sortable), even if the example is for filament version 2.

Sorry, only just seen this. Yes, I think using Spatie's package could have some advantages, we just need to be careful that it doesn't conflict with anything we may have done, e.g. ordering etc.

Also, you'll want to target 0.8 for this now, thanks.

glennjacobs:
Closing as this PR has had a status of "Draft" for a long time and our focus is now moving to 1.x.

Issue #1355, opened 2023-12-06
[HUB] Error when searching via Scout in LunarPHP HUB


Globalaj:
* Lunar version: 0.7.0
* Laravel Version: 10
* PHP Version: 8.2
* Database Driver & Version: 10.4.27-MariaDB

### Expected Behaviour:

When using LunarPHP Scout, make it possible to search for email, phone number with special characters like @ and + in the area code and more.

### Actual Behaviour:

When I search via HUB for an email address with a special @ sign or a phone number with a prefix with a special + sign, the following error pops up.

![image](https://uploads.linear.app/349e9a44-b829-4968-91b5-53f939b10c03/d2e63ba8-e1ad-4920-8644-86c7bb5bab73/717da213-2a8d-4837-a42e-b76aa2e879da?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzM0OWU5YTQ0LWI4MjktNDk2OC05MWI1LTUzZjkzOWIxMGMwMy9kMmU2M2JhOC1lMWFkLTQ5MjAtODY0NC04NmM3YmI1YmFiNzMvNzE3ZGEyMTMtMmE4ZC00ODM3LWE0MmUtYjc2YWEyZTg3OWRhIiwiaWF0IjoxNzE3NDE4NzA5LCJleHAiOjMzMjg3OTc4NzA5fQ.zExEym2gNgeu10EZV0_Og-tDXVdQat_rZmqcYoNJ3tM)

lguichard:
I submit a pull request, the exception due to query SQL on scout-database-engine

wychoong:
there is a merged PR, that should closed this issue

PR #1360, opened 2023-12-14
Fix espace query on DatabaseEngine (#1355)

lguichard:
This PR fix issue #1355 

DatabaseEngine no longer espace query on SQL 

The engine execute query without quote and generate syntax error on database server

Before 

```
SELECT
  *
FROM
  `search_index`
WHERE
  `index` = "demostore_products"
  AND MATCH (`content`) against (email@* IN boolean mode)
``` 

After 

```
SELECT
  *
FROM
  `search_index`
WHERE
  `index` = "demostore_products"
  AND MATCH (`content`) against ('"email@*"' IN boolean mode)
```

### DIFF
--- a/packages/admin/src/Http/Livewire/Components/Discounts/AbstractDiscount.php
+++ b/packages/admin/src/Http/Livewire/Components/Discounts/AbstractDiscount.php
@@ -134,13 +134,6 @@ abstract class AbstractDiscount extends Component
             ->map(function ($limitation) {
                 return array_merge($this->mapProductVariantToArray($limitation->purchasable), ['type' => $limitation->type]);
             });
-            
-        $this->selectedProductVariants = $this->discount->purchasableLimitations()
-            ->wherePurchasableType(ProductVariant::class)
-            ->get()
-            ->map(function ($limitation) {
-                return $this->mapProductVariantToArray($limitation->purchasable);
-            });
 
         $this->selectedConditions = $this->discount->purchasableConditions()
             ->wherePurchasableType(Product::class)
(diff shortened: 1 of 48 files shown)


glennjacobs:
Do we need the extra inverted commas? I assumed Laravel would handle that.

E.g. would this work?

```php
fn ($query) => $query->whereFullText('content', '"'.$builder->query.'*"', ['mode' => 'boolean'])
```

glennjacobs:
Also, could you add a test to show it working?

lguichard:
> Do we need the extra inverted commas? I assumed Laravel would handle that.
> 
> E.g. would this work?
> 
> ```
> fn ($query) => $query->whereFullText('content', '"'.$builder->query.'*"', ['mode' => 'boolean'])
> ```

No, me too !  Laravel don't espace, i needed to do before inject on whereFullText.

``` php
fn ($query) => $query->whereFullText('content', '"'.$builder->query.'*"', ['mode' => 'boolean'])
```
Works fine too. More clean 

But i found a new error with "@email string => Syntax error or access violation: 1064 syntax error, unexpected FTS_TERM, expecting FTS_NUMB.  

I had make a quick fix, but the issue is also maybe a mysql injection security report when boolean mode is active on Laravel ? I'm not found any espace on Illuminate/Database/Query/Grammars/MySqlGrammar.php

The code works in any case. I think no user will use the + and - operators in their search 

``` php
fn ($query) => $query->whereFullText('content', '"'.$builder->query.'*"')
```

lguichard:
@alecritson i closed accidentally this issue, oups

PR #1379, opened 2023-12-18
Auto-fix Pint Styling

glennjacobs:
Automatically fixes Pint styling issues, rather than having to manually fix.


.github/workflows/fix-code-style.yml
```diff
@@ -0,0 +1,29 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.head_ref }}
+      - name: Set up PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: 8.3
```
alecritson, line 17: We don't use `8.3` anywhere else in actions, should this be `8.2` until we change?

glennjacobs: Not sure it matters that much, as it's only code styling. I figured it best to not start with an outdated version.

alecritson: Just thinking for the sake of consistency

glennjacobs: We need to move to 8.3, so I suggest it stays and we add a task to the roadmap to upgrade everything to 8.3.

alecritson (approved):

PR #1398, opened 2023-12-20
Fix ListOrders not use enum on default size

lguichard:


alecritson:
@lguichard Looks like this was taken care of as part of another PR :)

PR #1400, opened 2023-12-20
Move default pagination to 20, 50, 100, all

lguichard:
Currently, the default pagination use Filament default configuration. I purpose to change this 

It is more ergonomic and useful, show the products and items in a trice


glennjacobs:
This does raise the question as to whether we should have "all" as a default option.

Some tables will have thousands of records and so "all" is likely to be quite dangerous.

adevade:
> This does raise the question as to whether we should have "all" as a default option.
> 
> Some tables will have thousands of records and so "all" is likely to be quite dangerous.

I usually remove it from my Filament tables for that reason, so +1 from me.

lguichard:
@glennjacobs correct, i will remove 'all' option. During my tests, 100 lines was the maximum for having a quick load.

binaryfire:
There's an open Filament issue re: large table performance: https://github.com/filamentphp/filament/issues/9304. The consensus there was 100 max as well.

@glennjacobs I remember you saying something about wrangling with Livewire performance when creating Lunar’s current datatables package. Maybe you've got some insights you could share in the Filament issue to help the team improve this?

glennjacobs:
> There's an open Filament issue re: large table performance: https://github.com/filamentphp/filament/issues/9304. The consensus there was 100 max as well.
> 
> 
> 
> @glennjacobs I remember you saying something about wrangling with Livewire performance when creating Lunar’s current datatables package. Maybe you've got some insights you could share in the Filament issue to help the team improve this?

The performance there was mainly Eloquent related which Filament v3 seems to have addressed pretty well. I agree, 100 max.

packages/admin/src/Support/Pages/BaseListRecords.php
```diff
@@ -3,9 +3,18 @@
...
 
 abstract class BaseListRecords extends ListRecords
 {
     use Concerns\ExtendsHeaderActions;
     use \Lunar\Admin\Support\Concerns\CallsHooks;
-}
+
+    public function table(Table $table): Table
+    {
+        parent::table($table);
```
alecritson: Could this be shortened to `return parent::table($table)->paginated([20, 50, 100]);` ?

wychoong:
can this be configurable or use extension

packages/admin/src/Support/Pages/BaseListRecords.php
lguichard: I fixed.

lguichard:
@wychoong Exactly, i can create config/panel.php with configuration variables of panel and move pagination default. 
@alecritson Do you have planned separate issue with global configuration of panel ?

glennjacobs:
I'm not keen on the config approach initially.

Having looked at Filament there doesn't appear to be a way to set the options globally.

I do wonder if Filament would consider a PR to allow pagination values to be set globally on the panel or something. I'd need to look at how doable that would be. Then it would be easy for a developer to change.

adevade:
You mean [setting global pagination settings like this](https://filamentphp.com/docs/3.x/tables/advanced#global-settings)?

glennjacobs:
> You mean [setting global pagination settings like this](https://filamentphp.com/docs/3.x/tables/advanced#global-settings)?

Ahha! I was searching for that and couldn't find it... that's exactly it. 

So we could look to use that in our service provider, as long as devs can override it in their app service provider.

lguichard:
Nice thanks @sandervankasteel 

> Vous voulez dire [définir des paramètres de pagination globale comme celui-ci](https://filamentphp.com/docs/3.x/tables/advanced#global-settings)  ?


We have two solutions to implement the global settings table configuration : 

First :  we create a new options on LunarPanel Facade

``` php 
      LunarPanel::configureTable(fn(Table $table) : void {
           // ...
        });
```


Second : Add on LunarPanelManger

``` php 
        /*
        *   Default global settings for Page
        */
        Table::configureUsing(function (Table $table): void {
            $table
                ->paginationPageOptions([20, 50, 100]);
        });

```
For custom by devs : 

``` php 
    // AppServiceProvider.php 
    // ...
    public function register(): void
    {
        // Register panel
        LunarPanel::register();

        // Override panel
        LunarPanel::panel(fn($panel) =>
            $panel->path('lunar')
                ->brandName('My custom name')
        )
        ->register();

        // Override default configuration
        Table::configureUsing(function (Table $table): void {
            $table
                ->paginationPageOptions([1, 5, 2]);
        });
       // ...
```

What do we do ? @glennjacobs

glennjacobs:
I would put the following in the Lunar service provider

```php
Table::configureUsing(function (Table $table): void {
    $table
        ->paginationPageOptions([20, 50, 100])
        >defaultPaginationPageOption(20);
});
```

and then test that a developer can override this is their own app service provider.

binaryfire:
@glennjacobs This will make all Filament tables (including non-Lunar ones) use these pagination options.

glennjacobs:
> @glennjacobs This will make all Filament tables (including non-Lunar ones) to use these pagination options.

Ah yes, that's a good point.

lguichard:
> I would put the following in the Lunar service provider
> 
> ```
> Table::configureUsing(function (Table $table): void {
>     $table
>         ->paginationPageOptions([20, 50, 100])
>         >defaultPaginationPageOption(20);
> });
> ```
> 
> and then test that a developer can override this is their own app service provider.

That's it, i test override on AppServiceProvider works well

wychoong:
Use the resource extension, and a global in lunar panel manager. So that a global can be set there, and control for each resources

PR #1401, opened 2023-12-20
Add new badge count on sidebar : product, channels

lguichard:
I purpose to add new count badge on sidebar.

### DIFF
--- a/packages/admin/src/Filament/Resources/ChannelResource.php
+++ b/packages/admin/src/Filament/Resources/ChannelResource.php
@@ -41,6 +41,16 @@ class ChannelResource extends BaseResource
         return __('lunarpanel::global.sections.settings');
     }
 
+    public static function getNavigationBadge(): ?string
+    {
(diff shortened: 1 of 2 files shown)


glennjacobs:
Is this needed?

I think for things like Orders it makes sense, but not sure we need a constant reminder of how many products and channels we have. If we add lots of badges, it will distract from the more important ones.

@alecritson can I have your opinion please?

alecritson:
I agree with @glennjacobs I'm open to them being added where it makes sense, otherwise it's just noise.

glennjacobs:
Closing as I don't see a value in the additional badges.

lguichard:
Ok, as you want. Maybe channel it's not necessary. The number of products in the catalog is not indicated anywhere. For example with my catalog team the leader like follow number of new products, draft products etc. 

I purpose variation on new commit : 


Capture d’écran 2023-12-20 à 16 25 08

PR #1406, opened 2023-12-21
Feature - Global search

### LINKED ISSUE #1364: Global search

lguichard:
Closes #1364 

Implement a filament global search without GlobalSearchCustomProvider. There's a separate issue for it.

The global search indexes customer, order, collection, brand, product. Each ressource indexes the fields useful for the user.

It's a simple implementation use default Filament global search feature. In the next step, we will have write a new GlobalSearchProvider to support Laravel scout and use Model indexer in the core

Have fun the new filament panel !


lguichard:
@glennjacobs I added custom fields attributes search on TranslatedText

glennjacobs:
I've added a couple of tweaks.

There are two outstanding issues as I see it...

1) It doesn't use Laravel Scout like the table search does.
2) The speed is quite slow, this is likely due to it searching attribute JSON data. It is possible to add indexes to JSON fields in MySQL, so I guess it's possible we could tell devs to manually add them, it's not ideal though. Or maybe we only support searching on attribute data when using Laravel Scout?

glennjacobs (approved):

PR #1407, opened 2023-12-21
Feature - Add new badge total product published 

lguichard:
Reference #1401 pr

The number of products in the catalog is not indicated anywhere. For example with my catalog team the leader like follow number of new products, draft products etc. 

I purpose variation on new commit,  I think that’s a much better approach.

Capture d’écran 2023-12-21 à 13 46 50

### DIFF
--- a/packages/admin/src/Filament/Resources/ProductResource/Pages/ListProducts.php
+++ b/packages/admin/src/Filament/Resources/ProductResource/Pages/ListProducts.php
@@ -8,6 +8,7 @@ use Filament\Resources\Components\Tab;
 use Illuminate\Contracts\Pagination\Paginator;
 use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Database\Eloquent\Model;
+use Illuminate\Support\Number;
 use Lunar\Admin\Filament\Resources\ProductResource;
 use Lunar\Admin\Support\Pages\BaseListRecords;
 use Lunar\Facades\DB;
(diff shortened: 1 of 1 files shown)


alecritson:
I quite like this, although just a few points on initial testing:

- Can we format the counts? for example 

image

Would be nicer if it was `18,958`

- The counts don't take into account filters, so if you were to filter by brand the counts stay the same, which I don't think is reflective of what the table is showing.

What do you think @glennjacobs ?

glennjacobs:
> I quite like this, although just a few points on initial testing:
> 
> * Can we format the counts? for example
> 
> image
> Would be nicer if it was `18,958`
> 
> * The counts don't take into account filters, so if you were to filter by brand the counts stay the same, which I don't think is reflective of what the table is showing.
> 
> What do you think @glennjacobs ?

Number formatting will need to take into account internationalisation. Could use the new Laravel Number helpers.

lguichard:
I fixed, that's better with format, i'm forgot this point. Thanks @alecritson 

Capture d’écran 2023-12-21 à 17 34 07

glennjacobs:
I put the "Draft" count on the page, as is assists editors so they know what products need to be completed and published. However, a count of published products is more suited to a Products dashboard/report.

I want to avoid cluttering the UI too much with counts to keep things clean and focused. So I'm going to close this PR and we'll solve this instead with a future Products dashboard and report.

PR #1409, opened 2023-12-21
Model extending via interfaces

glennjacobs:
Example usage

```php
use App\Models\ProductType;
use Lunar\Facades\ModelManifest;
use Lunar\Models\Contracts\ProductType as ProductTypeContract;

public function boot(): void
{
    ModelManifest::replace(
        ProductTypeContract::class,
        ProductType::class
    );
}
```

Explicit route model bindings are set in the core, ready to be used.

```php
Route::get('/test/{productType}', function ($productType) {
    dd($productType);  // App\Models\ProductType
});
```

Get the class from the original model

```php
\Lunar\Models\ProductType::modelClass();  // App\Models\ProductType
```

So you can then use it to set Eloquent relationships

```php
public function productType(): BelongsTo
{
    return $this->belongsTo(ProductType::modelClass());
}
```

Filament Resources can use the model contract

```php
use Lunar\Models\Contracts\ProductType;

class ProductTypeResource extends BaseResource
{
    protected static ?string $model = ProductType::class;
```

## TODO

- [x] Auto-load models from the models directory
- [x] Make contracts for all models
- [ ] Add facility to allow Eloquent static calls to continue to work
- [ ] Polymorphic relation support
- [ ] Update ModelManifest facade's doc blocks
- [ ] Update all Eloquent relationships to use `::modelClass()`
- [ ] Update all Filament resource model references to use the Contract instead
- [ ] Ensure any other code in the core is using models correctly
- [ ] Tests
- [ ] Documentation


packages/admin/src/Support/Resources/BaseResource.php
```diff
@@ -40,4 +40,17 @@ protected static function hasPermission(): bool
...
+    {
+        $class = new \ReflectionClass(static::$model);
+
+        if ($class->isInterface()) {
+            return app()->get(static::$model)::class;
+        }
+
+        return static::$model ?? (string) str(class_basename(static::class))
+            ->beforeLast('Resource')
+            ->prepend('App\\Models\\');
```
wychoong:
```suggestion
        return parent::getModel();
```

wychoong:
can you add a test to ensure model events are more triggered twice when extended

example: `App\Models\Product` extends base model

in current implementation, when a product is created it will create 2 `Url` records

glennjacobs:
> can you add a test to ensure model events are more triggered twice when extended
> 
> example: `App\Models\Product` extends base model
> 
> in current implementation, when a product is created it will create 2 `Url` records

Will do, but pretty sure this implementation won't do that.

wychoong:
> Get the class from the original model
> 
> ```
> \Lunar\Models\ProductType::modelClass();  // App\Models\ProductType
> ```

do you think `::baseModel` is better?

glennjacobs:
> > Get the class from the original model
> 
> > 
> 
> > ```
> 
> > \Lunar\Models\ProductType::modelClass();  // App\Models\ProductType
> 
> > ```
> 
> 
> 
> do you think `::baseModel` is better?

Base model wouldn't be correct.

lguichard:
Just and idea, instead use Product::modelClass() in database, take advantage to add new option to  key store in database and separate relation between name of class and type. During life of project is a pain to handle complex data in database. 

``` php
// Use morphMap from Laravel
ModelManifest::morphMap([
    'product' => Product::modelClass(),
    'collection' => Collection::modelClass(),
    // ...
]);

public function productType(): BelongsTo
{
    return $this->belongsTo(ProductType::morphClass());
}
```

glennjacobs:
Replaced by #1661

PR #1410, opened 2023-12-22
add listener to hub activityLogFeed

wychoong:
using same listener name as filament panel


alecritson:
Should this be targeting `1.0` if this is specific to Filament?

wychoong:
> Should this be targeting `1.0` if this is specific to Filament?

Nope. This is for < 1.0
I’m just referring that in 1.0 there is a listener for refresh. And this is using the same name

alecritson:
Gotcha, thanks for this!

alecritson (approved):

PR #1414, opened 2023-12-22
Feature - Access control brands permission

lguichard:
I would like to propose a new permission catalog:manage-brands. 

No longer on the previous version the brand permission never implemented.


glennjacobs:
I think this is likely to open up further discussions in terms of what should be controllable and what shouldn't. 

@alecritson do we want to be concerning ourselves with this at the moment, considering the MVP goal?

alecritson:
Yeah I think, as you say, it can open the discussion quite considerably and although I don't think this PR is wrong at all, I kind of feel like it should all be looked all at once to see what area's would require what level of access etc so it's more succinct

glennjacobs:
OK, so I'm going to close this PR because we need to concentrate on the MVP and I think this is a much bigger consideration than just Brands. I'll create an issue under "Planned" to review permissions and how they can be extended.

glennjacobs:
See #1415

lguichard:
Ok, guys. I made this pr as discover a new permission bug on sidebar. Continue to discuss on new thread.

PR #1419, opened 2023-12-22
Feature - Reintroduce the installation command for the admin panel

### LINKED ISSUE #1417: Reintroduce the installation command for the admin panel
Currently, there is no option to create a Staff login when installing `v1.0`.

The install command from the previous hub needs to be implemented. See https://github.com/lunarphp/lunar/blob/0.7/packages/admin/src/Console/Commands/InstallHub.php

We don't have any custom assets to be publishing this time, but we might want to incorporate publishing Filament's assets perhaps?

lguichard:
Closes #1417


packages/core/src/Console/InstallLunar.php
```diff
@@ -237,12 +293,35 @@ public function handle(): void
...
+            $lastname = $this->ask('Whats your last name?');
+            $email = $this->ask('Whats your email address?');
+            $password = $this->secret('Enter a password');
+
+            /** @var Staff $staff */
+            $staff = Staff::create([
+                'firstname' => $firstname,
+                'lastname' => $lastname,
+                'email' => $email,
+                'password' => bcrypt($password),
```
wychoong: should be 
```php
use Illuminate\Support\Facades\Hash;

Hash::make($password)
```

packages/core/src/Console/InstallLunar.php
```diff
@@ -237,12 +293,35 @@ public function handle(): void
...
+                'admin' => true,
+            ]);
+
+            $staff->syncRoles('admin');
         }
 
+        
+        $this->comment('Publishing assets...');
+        $this->publishAssets();
```
wychoong, line 319: filament's recommendation is to add this to composer script, is adding this here fix or any reason to have this run after lunar?

packages/core/src/Console/InstallLunar.php
```diff
@@ -237,12 +293,35 @@ public function handle(): void
...
+        $this->newLine();
+
+        $this->comment('Cache clear...');
+        $this->cleaning();
```
wychoong: any reason why this need to be run after install?

glennjacobs:
You have worked on an issue assigned to me. There's quite a lot I'd do differently, so I'll take over this PR.

lguichard:
> You have worked on an issue assigned to me. There's quite a lot I'd do differently, so I'll take over this PR.

Ok i didn’t see the issue assigned.

lguichard:
Closed by another PR.

PR #1421, opened 2023-12-26
Feature - Product tag editing

### LINKED ISSUE #1373: Product tag editing
We are not converting to Spatie Tags for the moment, so we need to implement the tag editing field.

lguichard:
Closes #1373


packages/admin/resources/views/livewire/components/tags.blade.php
```diff
@@ -1,14 +1,17 @@
...
         addTag(newTag) {
             let trimTag = newTag.toUpperCase().trim();
-    
-            if (!this.tags.includes(trimTag)) {
+            
+            if (trimTag && !this.tags.includes(trimTag)) {
                 this.tags.push(trimTag);
             }            
+
+            console.log(trimTag);
```
wychoong: please cleanup the troubleshoot code

packages/admin/resources/views/livewire/components/tags.blade.php
```diff
@@ -1,14 +1,17 @@
 <div class="relative mt-1 space-y-4"
     x-data="{
         tags: $wire.entangle('tags'),
-        value: $wire.entangle('searchTerm', true),
+        newValue: $wire.entangle('searchTerm', true),
```
wychoong: don't think this change is necessary as it's not really a new value either

packages/admin/src/Filament/Resources/ProductResource.php
```diff
@@ -175,9 +176,8 @@ protected static function getStatusFormComponent(): Component
 
     protected static function getTagsFormComponent(): Component
     {
-        return Forms\Components\TextInput::make('tags')
-            ->label(__('lunarpanel::product.form.tags.label'))
-            ->placeholder('TODO: convert to Spatie Tags');
+        return Tags::make('tags')
+            ->label(__('lunarpanel::product.form.tags.label')); // TODO: convert to Spatie Tags
```
wychoong: remove the comment

wychoong (commented):
can you also provide a screenshot of the form component with label 😄

packages/admin/resources/views/livewire/components/tags.blade.php
lguichard: Oups ! Thx

packages/admin/resources/views/livewire/components/tags.blade.php
lguichard: I must be change the name of variable, there are a conflict with keyup.enter on input "$event.target.value" was empty.

That fixed two small bug on components: tag empty and key.enter event

packages/admin/src/Filament/Resources/ProductResource.php
lguichard: Why ?

lguichard:
> can you also provide a screenshot of the form component with label 😄

It's same that order tags editin
Capture d’écran 2023-12-26 à 16 24 08
g.

packages/admin/src/Filament/Resources/ProductResource.php
wychoong: because the text was a placeholder for a feature to be implemented.
the comment should be consider as outdated as mentioned in #1373 not to use spatie's for now. When the time comes to change to spatie's, that would be another issue/PR

wychoong:
> > can you also provide a screenshot of the form component with label 😄
> 
> It's same that order tags editin Capture d’écran 2023-12-26 à 16 24 08 g.

need some tweaking for the label, and ideally use [filament's wrapper/blade](https://filamentphp.com/docs/3.x/support/blade-components/input-wrapper) instead of managing the styling ourselves

maybe try fieldset
```html
<x-filament::fieldset
      :label="$getLabel()"
>
      @livewire('lunar.admin.livewire.components.tags', [
          'taggable' => $getRecord()
      ])
</x-filament::fieldset>
```

also the save button for the tags will be confusing 🤔

lguichard:
The save button is a behavior choice by core team, i'm also confusing. 

 I had tested to change getDefaultForm but I wasn't really satisfied. Move to sub navigation would be the right solution

glennjacobs:
As the tags field is part of the main form, it shouldn't have a separate "Save" button. It should save along with the other components.

lguichard:
@glennjacobs I refactored the code. I replaced and removed the custom Livewire component. Now we use the native filament tag input override. When we convert the tags to Spatie the change will be very fast.

Now the panel have only two class that manage Tag resource

``` php

// Edit form tags
use Lunar\Admin\Support\Forms\Components\Tags inherit from Filament\Forms\Components\TagsInput

// Usage
Tags::make('tags')
  ->suggestions(Tag::all()->pluck('value')->all())
  // ....
  ->label('My label');
``` 

``` php
// Display list tags
Lunar\Admin\Support\Infolists\Components\Tags inherit from Filament\Infolists\Components\TextEntry; 

// Usage
Tags::make('tags')
  ->label('My Label');
```

The form in the order edit is same that behavior Shipping Address and Billing Address.

Capture d’écran 2023-12-28 à 21 04 54

Capture d’écran 2023-12-28 à 20 52 06

glennjacobs (approved):

glennjacobs (approved):

PR #1425, opened 2023-12-28
Fix custom translated field

wychoong:
- fix displaying multiple language in form
- use default language in table
- unique validation on handle
- validate if default language is required

image

~caveats: the `required` validation for the field doesnt work. but at least can manage the product option now~


lguichard:
Oups, i also created a new branch with translated text. Why don't use legacy behavior on render ? I purpose to add "Locales" button as previous version before filament. You can discover my code here https://github.com/lguichard/lunar/commits/feature/field-translated-text/


The preview is :  
Capture d’écran 2023-12-28 à 10 36 16


Capture d’écran 2023-12-28 à 10 41 25

wychoong:
> Why don't use legacy behavior on render ?

In my pr it’s mainly focus on hydrating and dehydrating the field to ensure the basics are usable. For the UI part I only added the input field prefix to indicate the language.

I’m not sure the final design approach of core team, whether to implement the same from legacy hub, personally I’m don’t prefer clicking to enable the language. I mean if you enabled multi languagea, you should be wanting to localize?

lguichard:
@wychoong Ok. Personally, when you manage many language displaying all fields looks visually heavy. We should be open a new PR with UI part.

@glennjacobs do you have suggestions ?

glennjacobs:
> @wychoong Ok. Personally, when you manage many language displaying all fields looks visually heavy. We should be open a new PR with UI part.
> 
> @glennjacobs do you have suggestions ?

I think a toggle button on every translatable field is a bit too much, but also I don't think we should be showing ALL languages at the same time as it could become far too much.

Having thought about this problem, this is the suggested route we'd like to try.

image

So we would have an action in the section header to toggle the language that is shown. If a section doesn't have a translatable field, then it would make sense to not show it.

I'm not sure what we want to do in regards to a PR for this?

@wychoong @lguichard do either of you have a sensible starting point, and would you like to contribute it?

wychoong:
> So we would have an action in the section header to toggle the language that is shown. If a section doesn't have a translatable field, then it would make sense to not show it.

Will have try for 1.x mvp

wychoong:
@glennjacobs  this is a quick first pass

- the toggle button can be extracted for reuse
- toggle button below the text fields are just for example

let me know your feedback then I can finish it

https://github.com/lunarphp/lunar/assets/67364036/84c7621e-3970-45c3-a8a1-2a3e9c5c94bf

glennjacobs:
@wychoong Initially I was thinking it would be more like a language selector, rather than a show/hide option. 

So you'd have a dropdown to select which language you want to edit.

These considerations are more for the attribute editing, as the product option edit screen is fairly basic. But we want to ideally have a consistent UX for this.

wychoong:
> @wychoong Initially I was thinking it would be more like a language selector, rather than a show/hide option.
> 
> So you'd have a dropdown to select which language you want to edit.
> 
> These considerations are more for the attribute editing, as the product option edit screen is fairly basic. But we want to ideally have a consistent UX for this.

something like this?

https://github.com/lunarphp/lunar/assets/67364036/85d69e92-54a5-4dee-8bd5-dbd625b3aecd

adevade:
> I think a toggle button on every translatable field is a bit too much, but also I don't think we should be showing ALL languages at the same time as it could become far too much.
> 
> Having thought about this problem, this is the suggested route we'd like to try.
> 
> image
> 
> So we would have an action in the section header to toggle the language that is shown. If a section doesn't have a translatable field, then it would make sense to not show it.

Do you mean that each section should have its own locale switcher? Wouldn't it maybe be easier to have one switcher for the whole page?

PR #1441, opened 2024-01-03
Feature - Status editing updates

### LINKED ISSUE #1374: Product publishing
There is currently no validation logic for publishing products. We need to re-implement this and also probably move the status field out to an action in the header.

alecritson:
This PR looks to move the status update on product editing to a header action and also provide feedback for when a product is considered unavailable or the status is set to draft.


packages/admin/resources/lang/en/product.php
```diff
@@ -6,6 +6,16 @@
 
     'plural_label' => 'Products',
 
+    'status' => [
+        'unpublished' => [
+            'content' => 'Currently in draft status, this product is concealed across all channels and customer groups.',
```
glennjacobs: `concealed` seems a bit of a strange word to use, although technically correct. What do you think?

glennjacobs:
Apart from the "concealed" word discussion, this looks good to go.

PR #1443, opened 2024-01-04
Support attributes on customer groups

ryanmitchell:
This PR adds support for attributes on customer groups using the pre-existing conventions and approaches in the codebase.

I've not added any UI for managing these - I assume that would need done on the Filament branch later down the line?


ryanmitchell:
Any thoughts on this one?

glennjacobs:
> Any thoughts on this one?

My initial reaction was to question why it was needed. Can you explain your use case?

ryanmitchell:
We have a few needs for this, the main one is that customer orders should be pushed to a 3rd party system using different 'channel ids' based on the customer group they are in. By allowing attributes this can be managed through the UI.

Beyond that we could show different messaging to customers based on their groups, and this could be controlled from the UI too.

glennjacobs:
@ryanmitchell I am happy to add attributes to customer groups, but I think adding this to 0.8 without the UI, but having it show under settings is wrong. 

Maybe we should update this to be a 1.x PR where we can add the admin UI?

ryanmitchell:
No problem - closing in favour of https://github.com/lunarphp/lunar/pull/1726

PR #1445, opened 2024-01-04
Feature - Add scout searching to pages with searchable models.

### LINKED ISSUE #1377: Laravel Scout searching

alecritson:


glennjacobs:
Does this allow Scout to be optional?

alecritson:
> Does this allow Scout to be optional?

No. How do you want it to be optional? In the config somewhere?

alecritson:
> Does this allow Scout to be optional?

I have made some tweaks to make Scout optional via a config setting. I have also refactored things to keep it DRY and added extra checks for when Scout isn't available, even if enabled. This will then fallback to the default if not enabled or unavailable.

packages/core/src/Search/OrderIndexer.php
```diff
@@ -34,12 +34,16 @@ public function getFilterableFields(): array
 
     public function makeAllSearchableUsing(Builder $query): Builder
     {
+        // Limit to the last 5 years
```
glennjacobs: Not sure this logic should be in the core? If a site wants to restrict to 5 years, I'd suggest it's done with a custom indexer.

glennjacobs (approved):

PR #1449, opened 2024-01-05
Use Shout instead of custom alert component

wychoong:


packages/admin/src/Filament/Resources/OrderResource/Pages/ManageOrder.php
```diff
@@ -789,10 +792,11 @@ protected function getCaptureAction(): Actions\Action
                             return null;
                         }
 
-                        return view('lunarpanel::components.alert', [
-                            'content' => $message,
-                            'color' => 'danger',
-                        ]);
+                        return Shout::make('aaa')
```
glennjacobs: `aaa`?

PR #1454, opened 2024-01-08
Feature - Shipping Addon

### LINKED ISSUE #1391: Update table rate shipping package to add Filament screens

alecritson:


ryanmitchell:
Oh... so the package is being brought into core?

glennjacobs:
> Oh... so the package is being brought into core?

Yeah, for 1.x

alecritson:
> Oh... so the package is being brought into core?

It's just being added as part of the mono repo, it'll still function as an add-on. So that it's easier to keep track of changes and provide any fixes in one hit rather than being spread out and having to think what might need fixing across repos

ryanmitchell:
Nice

glennjacobs:
image


Also, the price field is very wide...

glennjacobs:
Does the Base Price need to consider currency also?

glennjacobs:
image

Can the name and description be translated?

alecritson:
> image
> 
> Also, the price field is very wide...

## Base price

Can be changed to just "Price" like it is on product pricing?

## Min Spend.

Would this be something that admins can choose? Currently, it's on the sub-total since the total will account for shipping costs already applied and could put itself over the threshold.

## Currency

Yeah, it should however, nothing in the panel currently handles currencies with pricing and it didn't seem like it should be part of this PR. It would need to be updated when this is done properly.

## Pricing

ShippingRates are purchasables so they use the same logic as other purchasables.

## Can name/desc be translated?

Are you asking if they are capable of being translated or if they can be made so they can be translated? If the former, then no, just a text column.

glennjacobs:
> ## Base price
> Can be changed to just "Price" like it is on product pricing?

OK.


> ## Min Spend.
> Would this be something that admins can choose? Currently, it's on the sub-total since the total will account for shipping costs already applied and could put itself over the threshold.

As you say, it will have to run off the sub-total. The confusing bit is whether it considers tax or not, as some will be storing prices inclusive of tax. Maybe we have helper text that specifies what it is, based upon whether `lunar.pricing.stored_inclusive_of_tax` is set to true or not. And of course, the actual shipping logic will need to be updated also.

> ## Currency
> Yeah, it should however, nothing in the panel currently handles currencies with pricing and it didn't seem like it should be part of this PR. It would need to be updated when this is done properly.

Right, so we need a proper pricing field/component to handle this. I'm happy to come back to this, as long as we make an issue to remember :-)

> ## Pricing
> ShippingRates are purchasables so they use the same logic as other purchasables.

Again, I think we might just want to add some helper text to reassure the user.

> ## Can name/desc be translated?
> Are you asking if they are capable of being translated or if they can be made so they can be translated? If the former, then no, just a text column.

I think we will need to add this, but I think we can come back to it as long as we make an issue to remind us.

packages/table-rate-shipping/lang/en/shipping-methods.php
```diff
@@ -0,0 +1,5 @@
+<?php
+
+return [
+    '',
```
glennjacobs: What is this file for? It's empty and shouldn't language files reside in the resources folder?

PR #1455, opened 2024-01-08
Add initial discount listing

### LINKED ISSUE #1371: Discount editing

alecritson:


packages/admin/src/Filament/Resources/DiscountResource.php
```diff
@@ -0,0 +1,99 @@
...
+use Filament\Support\Facades\FilamentIcon;
+use Filament\Tables;
+use Filament\Tables\Table;
+use Lunar\Admin\Filament\Resources\DiscountResource\Pages;
+use Lunar\Admin\Support\Resources\BaseResource;
+use Lunar\Models\Discount;
+
+class DiscountResource extends BaseResource
+{
+    //    protected static ?string $permission = 'catalog:manage-discounts';
```
glennjacobs: Any reason the permission is commented out?

alecritson: Permissions aren't currently set up so would be hidden from view.

PR #1461, opened 2024-01-10
Fix - Espace query on BaseListRecords

lguichard:
Minor change, i added espace query to fix error with empty results


glennjacobs:
This probably needs updating to use the new approach

```php
            $placeholders = implode(',', array_fill(0, count($ids), '?'));

            $query->whereIn(
                'id',
                $ids
            );

            $query->when(
                ! $ids->isEmpty(),
                fn ($query) => $query->orderByRaw("field(id, {$placeholders})", $ids->toArray())
            );
```

Does that approach suffer from the same issue?

alecritson:
Thanks @lguichard but it seems this has been patched in another PR along the line.

PR #1468, opened 2024-01-11
Fix date formatting in order-related charts

thomas-alrek:
This PR fixes some issues in the charts widgets, where the widgets was using `DATE_FORMAT` directly, instead of using the `db_date` helper function. This made the charts tightly coupled to MySQL.

This PR should fix this issue, ensuring the charts will work across the supported database drivers.


alecritson (approved):
Thanks for sorting this :)

PR #1475, opened 2024-01-16
Feature - Support of widget on a resource page

lguichard:
This PR add new hook to extend the ressource page with Filament header and footer widgets


packages/admin/src/Support/Pages/Concerns/ExtendsHeaderWidgets.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace Lunar\Admin\Support\Pages\Concerns;
+
+trait ExtendsHeaderWidgets
+{
+    protected function getDefaultHeaderWidgets(): array
```
alecritson, line 7: These should be static functions, check this class for reference https://github.com/lunarphp/lunar/blob/1.x/packages/admin/src/Support/Concerns/ExtendsForms.php

lguichard: Are you sure ? 

getdeFaultHeaderWidgets use to create a default widget in BaseCreateRecord, BaseEditRecord, BaseListRecord as getDefaultHeaderActions.

```php
// Example default widget in create brand 
class CreateBrand extends BaseCreateRecord
{
    protected static string $resource = BrandResource::class;
    
    public function getDefaultHeaderWidgets(): array
    {
        return [
            \Lunar\Admin\Filament\Widgets\Dashboard\Orders\OrderStatsOverview::make(),
        ];
    }
....


class MyCreateExtension extends CreatePageExtension
{
    public function headerWidgets(array $widgets): array
    {
        $widgets = [
            ...$widgets,
            //
        ];

        return $widgets;
    }
    ....
    class MyEditExtension extends EditPageExtension
    {
        public function headerWidgets(array $widgets): array
        {
            $widgets = [
                ...$widgets,
                //
            ];
    
            return $widgets;
        }

        ....
```

glennjacobs: Seems fine to me, I think Alec was getting mistaken for Resource concerns, whereas this is a Page concern.

glennjacobs (approved):

PR #1486, opened 2024-01-22
Allow discount cache to be reset

ryanmitchell:
This PR adds the ability to reset the DiscountManager query cache - this is necessary if you add a discount coupon and want to update the cart totals in the same request.

### DIFF
--- a/packages/core/src/Managers/DiscountManager.php
+++ b/packages/core/src/Managers/DiscountManager.php
@@ -209,6 +209,13 @@ class DiscountManager implements DiscountManagerInterface
         return $cart;
     }
 
+    public function resetDiscounts(): self
+    {
+        $this->discounts = null;
+
+        return $this;
+    }
+
     public function validateCoupon(string $coupon): bool
     {
         return app(
--- a/docs/core/reference/discounts.md
+++ b/docs/core/reference/discounts.md
@@ -66,6 +66,14 @@ Discount::usable()->get();
 Discount::products($productIds, $type = 'condition');
 ```
 
+### Resetting the discount cache
+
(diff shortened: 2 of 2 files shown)


packages/core/src/Managers/DiscountManager.php
```diff
@@ -209,6 +209,13 @@ public function apply(Cart $cart): Cart
         return $cart;
     }
 
+    public function resetDiscounts(): self
```
alecritson, line 215: Do you think `clearDiscounts` would be better? 🤔 @glennjacobs

glennjacobs: No, I think reset is correct.

glennjacobs:
@ryanmitchell I know the coupon stuff has tripped up devs a fair amount historically. Would it make sense for us to put this into the docs also?

ryanmitchell:
Yeah I think it would. I’ll add that to this PR at some point tomorrow

glennjacobs:
> Yeah I think it would. I’ll add that to this PR at some point tomorrow

I presume this solves the coupon issue, so we could do with a test to demonstrate that also pretty pls :-)

ryanmitchell:
@glennjacobs docs added. I tried to write a test but it's doing the same as `can_get_discount_with_coupon`, but that calls getDiscounts directly rather than bypassing the cache. If theres another test I can piggy-back on do let me know.

glennjacobs:
> @glennjacobs docs added. I tried to write a test but it's doing the same as `can_get_discount_with_coupon`, but that calls getDiscounts directly rather than bypassing the cache. If theres another test I can piggy-back on do let me know.

Thanks, I'll take a look.

glennjacobs (approved):

Issue #1485, opened 2024-01-19
Issue with discount exclusions


ryanmitchell:
We've started using discount exclusions on one of our live sites and its presenting us with an interesting issue in that any discounts with exclusions are never added to the cart due to the code here:
https://github.com/lunarphp/lunar/blob/b342878ea1fd7613f00a3a10544adfc3c9c7c582/packages/core/src/Managers/DiscountManager.php#L136-L149

Basically from what I can see, because a purchasables exists now (albeit an exclusion), the discount wont apply unless that purchasable is in the cart, which wont always be the case.

If you can direct me a little I'll open a PR to refactor this code ... my suggested approach would be to pass a type of 'limitation' to products() and productVariants() then update those scopes in the discount model so that the whereDoesntHave('purchasables') is a whereDoesntHave where the type is the same

PR #1487, opened 2024-01-22
Fix discount product and variants purchasables query

ryanmitchell:
Update the discount query check to only consider limitations, which allows exclusions to continue working. 
Maybe this needs extended to conditions too, in which case the scope would need changed to accept an array of types.

Closes https://github.com/lunarphp/lunar/issues/1485


alecritson:
Thanks for this @ryanmitchell I think if it's the exclusions that are causing the issue then the query could either exclude the exclusions or include limitations and as you say conditions.

ryanmitchell:
Do you want me to update the scopes to accept an array of types?

alecritson:
Yeah I think that would be beneficial, appreciate it

alecritson (approved):

PR #1488, opened 2024-01-22
Feat - Variant Generation

alecritson:
This PR looks to add variant generation to the filament hub. This isn't a port of the current offering in `0.7` but is instead a complete reimagining of how it should work, inspired by some other popular systems out there.


glennjacobs:
If there are no shared options, it just shows an empty dropdown list. Maybe we should show a message to say they don't have any set-up?

glennjacobs:
I don't think we need to say "Restricted" in the add option button, I think that's implied by having "shared" on the other button.

glennjacobs:
image

The input fields don't say what they are. In BigCommerce they make it clear which is the Name and which are the Values.

image

Also, does our layout allow for "option types" in the future?

glennjacobs:
image

The "add option" link feels a bit untidy next to the "add value". Maybe it needs to be presented as a typical button to define it more and add some spacing or a rule after, before the save buttons? Again, BigCommerce seem to have this UI working well, so we can take influence from it.

glennjacobs:
Once variants have been generated (and saved), you can't delete any.

glennjacobs:
image

The is too much padding on the left, and not enough on the right of each row. And too much space is given to the price and stock, which could probably be better used in the option name and SKU columns.

glennjacobs:
> No more shared options are available.

This suggests that a shared option has already been used, but I don't have any. Maybe we just remove the "more" word?

glennjacobs:
Removing all values for an option causes this error
image

I'm not sure how you are supposed to completely remove an option and all it's values?

glennjacobs:
I'm not sure if this should be included, but I can't make a shared option due to the following...

image

glennjacobs:
If I change the name of a value and save everything, then refresh the page, it hasn't actually updated and shows the old value when configuring the options.

glennjacobs:
image

I managed to remove all the values on a shared option, feels wrong as I can't put any values back.

lguichard:
@alecritson there is a little bug on dark mode 

Capture d’écran 2024-01-24 à 15 00 06

alecritson:
> @alecritson there is a little bug on dark mode
> 
> Capture d’écran 2024-01-24 à 15 00 06

Thanks for the heads up!

glennjacobs:
If I remove all the variant options I get the following error

image

glennjacobs:
Maybe for shared options we change the wording to "Remove Shared Option" so it doesn't sound so scary? This would help make it more obvious that it's a shared option also.

image

glennjacobs:
Adding an additional (normal) option gave me this error

image

glennjacobs:
If I remove all options and save the variants, I have to refresh to get the side menu to show all the sections again.

packages/admin/src/Support/Actions/Products/MapVariantsToProductOptions.php
```diff
@@ -0,0 +1,82 @@
+<?php
+
+namespace Lunar\Admin\Support\Actions\Products;
```
glennjacobs: This is a little odd, as the "Actions" folder to date has been used for Filament Actions, not Laravel-style actions. We probably need to think about giving it a more appropriate home :-)

glennjacobs (approved):

wychoong:
@alecritson the introduction of `product_product_option` table is it only for managing variants in filament? do we need to change the way we fetch variants?

alecritson:
@wychoong This is to make it A) Easier to fetch which options a product has and B) Because we now have local product options which can be specific to a product.

Before you would have to load all variants a product has, load all their product option values, fetch the option itself and then dedupe them it was quite the adventure. Now it can be simplified as such:


```php
$productOptions = $product->productOptions()->with([
    'values' => function ($query) use ($product) {
        $query->whereHas('variants', function ($relation) use ($product) {
            $relation->where('product_id', $product->id);
        });
    },
])->get();
```

wychoong:
> @wychoong This is to make it A) Easier to fetch which options a product has and B) Because we now have local product options which can be specific to a product.
> 
> Before you would have to load all variants a product has, load all their product option values, fetch the option itself and then dedupe them it was quite the adventure. Now it can be simplified as such:
> 
> ```
> $productOptions = $product->productOptions()->with([
>     'values' => function ($query) use ($product) {
>         $query->whereHas('variants', function ($relation) use ($product) {
>             $relation->where('product_id', $product->id);
>         });
>     },
> ])->get();
> ```

Noted. Mainly is because I have a process that create products outside of the provided product page. So now I need to update my code to cater insert to product_product_option, else it won’t show the variants in panel

ryanmitchell:
Just a heads up that since this PR merged you can no longer make variants on an entirely new product in the hub. You get an error that `Field 'tax_class_id' doesn't have a default value`:

CleanShot 2024-12-09 at 16 02 57@2x

If you fix that up, you then get an error on the next line as theres no $basePrice as theres no existing variant.

glennjacobs:
@alecritson Could you have a look in the morning please?

alecritson:
@ryanmitchell Are you able to provide some steps on how to reproduce this one? I've added a loom video showing what I did and it seemed to generate the variants fine 🤔 

https://www.loom.com/share/8c90d6cdf4014d25ad599d80c5048262

ryanmitchell:
@alecritson thanks for looking, you're right, it works fine when there is a base variant. The issue comes when you remove the base variant and try to add another. Maybe its not something you need to handle, but clients do strange things.

PR #1493, opened 2024-01-23
Installation process fix

webpatser:
Just did a clean install of v1.0. And this step is in the wrong order. 

First enable the filament panel. Then do the `lunar:install`. Otherwise an error occurs with the lunar:install, stating the panel is needed. But then it's unclear that you need to run the installer again to seed the database with needed permissions and users etc.


webpatser:
Since this PR was pending. I upgraded the whole install process. It now has correct return types in the sub-commands.
Also a check for a admin user was added, and if no admin is present, create one.

eugenefvdm:
Thanks so much for this.

I notice when running this pull request that `DB::transaction` still fails because of missing attributes, specifically `description`:

```mysql
SQLSTATE[HY000]: General error: 1364 Field 'description' doesn't have a default value (Connection: mysql, SQL: insert into `lunar_attributes` (`attribute_type`, `attribute_group_id`, `position`, `name`, `handle`, `section`, `type`, `required`, `default_value`, `configuration`, `system`, `updated_at`, `created_at`) values (Lunar\Models\Product, 1, 1, {"en":"Name"}, name, main, Lunar\FieldTypes\TranslatedText, 1, ?, {"richtext":false}, 1, 2024-02-11 06:24:53, 2024-02-11 06:24:53))
```

On my system the temporary fix for all attributes was to append a `description` field such as the example below:

```php
 Attribute::create([
            'attribute_type' => Product::class,
            'attribute_group_id' => $group->id,
            'position' => 1,
            'name' => [
                'en' => 'Name',
            ],
            'handle' => 'name',
            'section' => 'main',
            'type' => TranslatedText::class,
            'required' => true,
            'default_value' => null,
            'configuration' => [
                'richtext' => false,
            ],
            'system' => true,
            'description' => '', // TODO had to add this to make the seeder work
        ]);
```

I had to do this for all four attributes seeded in the installer to make it work.

From what I can see in the [1.x documentation](https://docs-v1.lunarphp.io/core/reference/attributes.html) description is maybe only partially documented. In the `MySQL` it's a JSON field that cannot be null.

I would be grateful to know what we can use as initial values for this `description`.

eugenefvdm:
Something else to consider to make a working install are tax defaults, apart from the `TaxClass`.

This is present in the Livewire starter kit:
https://github.com/lunarphp/livewire-starter-kit/blob/main/database/seeders/TaxSeeder.php

webpatser:
I will add the description. I didn't had that problem. But lots of merges last week, so maybe something changed upstream.

glennjacobs:
Thanks @webpatser for the help. We'll be working on this soon (this coming week) and this is a helpful start.

webpatser:
I added the new `description` field to the Lunar installer. This field was recently added in the migration file `2024_01_11_100000_add_description_to_attributes_table.php.`

Currently, the description is left blank; perhaps it should contain a default description. :-)

packages/core/src/Console/Commands/MakeLunarAdminCommand.php
```diff
@@ -0,0 +1,91 @@
...
+use Filament\Facades\Filament;
+use Illuminate\Console\Command;
+use Illuminate\Contracts\Auth\Authenticatable;
+use Illuminate\Support\Facades\Hash;
+use Lunar\Admin\Models\Staff;
+
+use function Laravel\Prompts\password;
+use function Laravel\Prompts\text;
+
+class MakeLunarAdminCommand extends Command
```
glennjacobs: This command should be in the admin package, as it references the Staff model.

packages/core/src/Console/InstallLunar.php
```diff
@@ -63,6 +64,12 @@ public function handle(): void
         }
 
         DB::transaction(function () {
+
+            if (! Staff::whereAdmin(true)->exists()) {
```
glennjacobs: We can't assume the admin package is installed

packages/core/src/Console/Commands/MakeLunarAdminCommand.php
```diff
@@ -0,0 +1,91 @@
...
+
+        $this->components->info('Success! '.($user->getAttribute('email') ?? 'You')." may now log in at {$loginUrl}");
+    }
+
+    public function handle()
+    {
+        $this->options = $this->options();
+
+        if (! Filament::getCurrentPanel()) {
+            $this->error('Filament has not been installed yet: php artisan filament:install --panels');
```
glennjacobs: This isn't required. The only thing required for Lunar's panel is to publish the Filament assets, which can be achieved by running

```
php artisan filament:assets
```

packages/core/src/Console/InstallLunar.php
glennjacobs: Some developers only install the core package.

glennjacobs (changes requested):

packages/core/src/Console/Commands/AddonsDiscover.php
```diff
@@ -33,5 +33,7 @@ public function handle(Manifest $manifest): void
         }
 
         $this->components->info('Addon manifest generated successfully.');
+
+        return Command::SUCCESS;
```
glennjacobs: Laravel now returns void from the handle method, see https://laravel.com/docs/10.x/artisan#command-structure

I suggest we simply omit the return statement now.

webpatser:
Feedback is processed..

glennjacobs (dismissed):

PR #1501, opened 2024-01-25
Feature - Translated column

lguichard:
This PR add new custom TranslatedTextColumn use for translated fields with some options  : 

- limitedToolip : display tooltip if the column contents exceeds the length limit.
- attributeData : Render from attribute_data field


glennjacobs (changes requested):
Could we import the class, rather than using the full namespace of `\Lunar\Admin\Support\Tables\Columns\TranslatedTextColumn` please?

lguichard:
@glennjacobs Ok !

lguichard:
@glennjacobs That's it

PR #1504, opened 2024-01-26
Feat - Variant resource editing

alecritson:
This PR adds the variant edit page


alecritson:
@glennjacobs I remember you had some idea's for how you wanted variant images to work, would be interested to hear that again before I start on that section.

packages/admin/src/Filament/Resources/ProductResource/Pages/EditProduct.php
```diff
@@ -21,6 +21,11 @@ public static function getNavigationIcon(): ?string
         return FilamentIcon::resolve('lunar::basic-information');
     }
 
+    public static function shouldRegisterNavigation(array $parameters = []): bool
+    {
+        return false;
+    }
+
```
adevade: I don't think this method takes any arguments? And it could be defined as a static property on the class, if no logic is needed.

```php
protected static bool $shouldRegisterNavigation = false;
```

https://filamentphp.com/docs/3.x/panels/navigation#disabling-resource-or-page-navigation-items

alecritson: I think their docs are wrong if anything, but I think it could potentially be moved to a static property on resources where no additional logic is required.

image

adevade: Hmm, yeah you're right. It seems to be passed sometimes... And on `EditRecord` it is indeed present.

https://github.com/search?q=repo%3Afilamentphp%2Ffilament+%22function+shouldRegisterNavigation%28%22+language%3APHP&type=code

lguichard:
@alecritson The ux of previous version was very fine with a major differentiation factor compared to others solutions. We can use the new Collapsible sidebar on Filament.

Capture d’écran 2024-01-26 à 11 31 54

glennjacobs:
> @alecritson The ux of previous version was very fine with a major differentiation factor compared to others solutions. We can use the new Collapsible sidebar on Filament.
> 
> Capture d’écran 2024-01-26 à 11 31 54

It worked for a smaller number of variants, but some stores have 100's of variants.

adevade:
There's also sub navigation since v3.1: https://filamentphp.com/community/alexandersix-filament-v3-1#sub-navigation

Same problem with 100's of variants though.

glennjacobs:
> There's also sub navigation since v3.1: https://filamentphp.com/community/alexandersix-filament-v3-1#sub-navigation
> 
> Same problem with 100's of variants though.

Yeah, we use that a fair amount. I can't think of an obvious way to paginate nav links though :-)

glennjacobs:
It could be achieved with custom pages/views, but as we're targeting an MVP currently, we just want to get something acceptable and usable at this stage.

lguichard:
Right, problem with 100's variants.

In next step, we can propose more options to edit quickly variants

glennjacobs (approved):

Issue #1447, opened 2024-01-04
New vs Returning Customer on Order listings


glennjacobs:
We want to include an indicator for whether an order is for a new or returning customer on the order listings page.

PR #1507, opened 2024-01-28
Feature - Display new / returning customer

lguichard:
This PR fix issue #1447. 

We can assume that in the future the information of Customer are grouping as one column.


alecritson (approved):
Thanks @lguichard just tweaked the column name but lets get it in :)

PR #1508, opened 2024-01-28
Fix - Limit the list of sku with line break 

lguichard:
Capture d’écran 2024-01-28 à 18 30 30


alecritson:
Maybe to save height space we can limit to just one SKU and show the "plus x more" since I don't think it's too much of a difference. What do you think?

lguichard:
@alecritson I tested this option but I had a conflict with the click on the resource.


https://github.com/lunarphp/lunar/assets/6489718/c5957d2b-f8d2-43d8-a595-f515546a5476

glennjacobs (approved):

PR #1509, opened 2024-01-28
Fix - Prevent tag format without uppercase

lguichard:


alecritson (commented):
Thanks @lguichard 

Had a quick look and it didn't change to uppercase when editing in a resource that uses tags i.e. products

Do we need some sort of unique validation on this?

I'm not sure it's the best UX having the value changed as you type, unless there was a debounce, personally I think converting to uppercase on save is fine as long as the unique validation takes into account it will be uppercase.

lguichard:
Thanks for the feedback. 

You're be right for UX. We need to change several places in panel to unified uppercase on tags. I'll review the code of tags components on this PR, give me some time.

"Do we need some sort of unique validation on this?" They must be unique

lguichard:
@alecritson I removed live() even debounce because there was some Filament bugs with unique validation.

This PR fix tag edition from sidebar menu not on tags associated to resource. In this case, i used TagInput by Filament but we don't have any option to custom render view after had created a tag. I plan to submit a PR at Filament in order to add new option.

alecritson:
@lguichard I've converted to draft whilst this is being worked on, as soon as it's ready mark it as ready and I'll have a look

glennjacobs:
I think forcing uppercase should be an optional behaviour, as it's a personal preference thing.

wychoong:
Does Laravel validation takes care of column collation?

glennjacobs:
@alecritson @lguichard I'm trying to understand what problem this is trying to fix, as we have no tests or description on this issue.

Is this a preference thing, or is there a problem? E.g. are we currently allowing both `HotProduct` and `hotproduct` for example?

alecritson:
Closing this due to inactivity. As we move towards a stable `1.0.0` if you still require this PR please ensure it is updated, with tests, where applicable, and reopen this pull request.

PR #1515, opened 2024-01-30
Feature - Pricing management

### LINKED ISSUE #1365: Product price editing improvements

alecritson:


glennjacobs:
Doesn't initially set the default currency

image

glennjacobs:
I think we default this to "1"

image

glennjacobs:
I think "Comparison Price" sounds better. Also, I think it could do with some helper text as a lot of developers don't know what to use it for.

image

glennjacobs:
I think we might need to look at a money field component, as not showing the decimal places looks a bit naff.

image

glennjacobs:
It's a bit odd these both have the "Pricing" title. Perhaps change the bottom one to "Prices"?

image

glennjacobs (changes requested):
See comments

glennjacobs:
We don't have the comparison price fields for the "Base Prices", which we'll need.

I'm unsure of the term "Base Price" as this generally means the price before any additional charges. But currently not sure of a better term. I'll do some research.

Can the hint use this text "Automatically generated based on currency exchange rates.", please?

For the Price Breaks table could we add the following description? "Reduce the price when a customer purchases in larger quantities."

PR #1522, opened 2024-01-30
sort new variants at top

wychoong:
currently when adding new product options to a product, new variants are added to the bottom of the list. Moving it up will improve UX

 image


glennjacobs:
The "new" options should go in option order, not necessarily at the bottom. We do want to improve this section by having an "unsaved changes" alert, but I'm not keen on this change.

PR #1526, opened 2024-01-31
Feat - Rename Tier to Quantity Break

alecritson:
This PR looks to change the wording of `tier` to `quantity_break` in the context of pricing


glennjacobs (changes requested):
We need an entry in the upgrade guide for this.

glennjacobs (approved):

PR #1528, opened 2024-01-31
Merge 0.8 into 1.x

glennjacobs:
- [ ] Sort the gazillion merge conflicts


alecritson:
Closing in favour of #1601 since we should be using an intermediary brach when merging between versions as fixing the conflicts would require commits directly on the `0.8` branch, which should remain intact.

PR #1539, opened 2024-02-07
Fix - Last Filament update

lguichard:
After this merge : https://github.com/filamentphp/filament/commit/e2a949bcc118239e86f22dec483a2f25234847f1 Filament triggered an exception "The panel has already been registered with the ID".

The id of panel already set in defaultPanel method.


lguichard:
Duplicate PR =)

PR #1540, opened 2024-02-07
fix panel id

wychoong:
Prior to [Filament 3.2.25](https://github.com/filamentphp/filament/releases/tag/v3.2.25) a panel can be set with a different ID when dev register in service provider, and to avoid internally referencing the wrong ID/panel, lunar overrides it again during registration.

With the latest change, the panel ID is now not allowed to be overridden, when attempted Filament will throw an exception, hence we can now(have to) remove the overrides

### DIFF
--- a/packages/admin/src/LunarPanelManager.php
+++ b/packages/admin/src/LunarPanelManager.php
@@ -86,8 +86,6 @@ class LunarPanelManager
             $panel = $fn($panel);
         }
 
-        $panel->id($this->panelId);
-
         Filament::registerPanel($panel);
 
         FilamentIcon::register([
--- a/packages/admin/composer.json
+++ b/packages/admin/composer.json
@@ -15,7 +15,7 @@
     "minimum-stability": "dev",
     "require": {
         "lunarphp/core": "self.version",
-        "filament/filament": "^3.1.23",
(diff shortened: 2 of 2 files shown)


packages/admin/src/LunarPanelManager.php
```diff
@@ -86,8 +86,6 @@ public function register(): self
             $panel = $fn($panel);
         }
 
-        $panel->id($this->panelId);
```
glennjacobs, line 89: Would we not also want to remove the $panelId property from the class now?

wychoong: we still use it in other place in the manager

https://github.com/lunarphp/lunar/blob/7ea8e820e6eab26cec27b942563ead3a364742fa/packages/admin/src/LunarPanelManager.php#L154-L157

I think it will be better than using string

glennjacobs: Ah I see, OK.

lguichard:
Oups, we have duplicate the PR https://github.com/lunarphp/lunar/pull/1539. @glennjacobs you can close my PR.

alecritson (approved):

PR #1542, opened 2024-02-08
Feature - Discount editing

alecritson:
Adds additional discount editing functionality.

- [x] Write tests


packages/admin/src/Filament/Resources/DiscountResource/Pages/EditDiscount.php
```diff
@@ -0,0 +1,64 @@
...
+        return [
+            Actions\DeleteAction::make(),
+        ];
+    }
+
+    protected function mutateFormDataBeforeSave(array $data): array
+    {
+        $minPrices = $data['data']['min_prices'] ?? [];
+        $fixedPrices = $data['data']['fixed_values'] ?? [];
+        $currencies = Currency::get();
```
glennjacobs: Should this only consider the enabled currencies?

PR #1565, opened 2024-02-16
Feature - New config on StandardMediaConversions

lguichard:
This PR remove hard coded value and create a new config


lguichard:
```sh
Run shivammathur/setup-php@v2
/usr/bin/bash /home/runner/work/_actions/shivammathur/setup-php/v2/src/scripts/run.sh

==> Setup PHP
✗ PHP Could not setup PHP [8](https://github.com/lunarphp/lunar/actions/runs/7928914847/job/21648120617?pr=1565#step:3:9).2
Error: The process '/usr/bin/bash' failed with exit code 1`
```
@glennjacobs 
The CI don't like me xD

glennjacobs:
Why do we need this? Lunar uses it for images, hence the collection name is called `images`.

lguichard:
I discover this hard coded value during the demostore conversion. The products seeder used media collection name "products". 

If they devs add Lunar package on existing project and that already use 'images' name, they need to change the default value. In big e-commerce website, we need to dispatch medias on multi storage buckets to optimize CDN etc... I think it might be necessary to split collection name to organize assets by Model.

glennjacobs:
@alecritson can I have your opinion, please?

Things to consider...

- Will this not break the admin panel, unless we introduce the config there also?
- I don't think we want to change the default 'images' collection name on other models. Sure, allow a dev to change it, but I don't believe we need to change them in our config.

lguichard:
@glennjacobs, to ensure backward compatibility probably better if we keep only 'media_collection' => 'images' allow a dev change it.

packages/core/config/media.php
```diff
@@ -13,6 +13,11 @@
         Lunar\Models\ProductOptionValue::class => StandardMediaDefinitions::class,
     ],
 
+    'collection' => [
+        'images' => 'images',
+        'videos' => 'videos',
```
glennjacobs: I don't think we need `videos` in the config, considering the system doesn't actually use it. It's only used in tests, so I think it would be misleading to include it here.

glennjacobs (changes requested):

alecritson:
@glennjacobs Doesn't look like we've hit every instance of `SpatieMediaLibraryImageColumn` when updating the `collection` method to config.

glennjacobs:
> @glennjacobs Doesn't look like we've hit every instance of `SpatieMediaLibraryImageColumn` when updating the `collection` method to config.

Should be sorted.

alecritson (approved):

glennjacobs (approved):

PR #1578, opened 2024-02-23
Fix: PHP 8.2 ${var} string interpolation deprecated

wychoong:


alecritson (approved):

packages/stripe/README.md
```diff
@@ -208,7 +208,7 @@ const buildForm = async () => {
         layout: "tabs",
         defaultValues: {
             billingDetails: {
-                name: `${billingAddress.value.first_name} ${billingAddress.value?.last_name}`,
+                name: `{$billingAddress.value.first_name} {$billingAddress.value?.last_name}`,
```
adevade, line 211: @alecritson Pretty sure this should be left as is since it's JavaScript, not PHP. Source: [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#string_interpolation)

Same goes for the other JS block further down in this file.

PR #1581, opened 2024-02-23
Feature - Register multiple extensions on LunarPanelManager

lguichard:
The PR added a new functionality to authorize multiple register extensions.


wychoong:
Would love to just have one register extension api that handle array or single extension

lguichard:
@wychoong i would be open to that.

@glennjacobs ?

packages/admin/src/LunarPanelManager.php
```diff
@@ -228,9 +228,18 @@ protected function defaultPanel(): Panel
...
-        $this->extensions[$pageClass][] = $extension;
+        $this->extensions[$pageClass][] = new $extension;
+
+        return $this;
+    }
+
+    public function registerExtensions(array $extensions, string $pageClass): self
+    {
+        foreach ($extensions as $extension) {
+            $this->extensions[$pageClass][] = new $extension;
```
alecritson: Instead of duplicating this part, should we just user the `registerExtension` method? Then if it ever changes we just update one place.

glennjacobs:
> @wychoong i would be open to that.
> 
> @glennjacobs ?

Yes, I'm open to that. It's just a case of finding a suitable function name for it.

lguichard:
@glennjacobs @alecritson 

We have two approaches

Create a LunarPanelExtension Facade

```php

LunarPanelExtension::register([
   ListActivities::class => \Lunar\Admin\Filament\Resources\ActivityResource\Pages\ListActivities::class
])
```

or

Use LunarPanel as follow : 

```php
LunarPanel::extensions([
   ListActivities::class => \Lunar\Admin\Filament\Resources\ActivityResource\Pages\ListActivities::class
])
```

```php
LunarPanel::panel(fn ($panel) => fn($panel) => $panel->path('admin'))
    ->extensions([
       ListActivities::class => \Lunar\Admin\Filament\Resources\ActivityResource\Pages\ListActivities::class
     ])
    ->register();

```

wychoong:
> @glennjacobs @alecritson
> 
> We have two approaches
> 
> Create a LunarPanelExtension Facade
> 
> ```
> LunarPanelExtension::register([
>    ListActivities::class => \Lunar\Admin\Filament\Resources\ActivityResource\Pages\ListActivities::class
> ])
> ```
> 
> or
> 
> Use LunarPanel as follow :
> 
> ```
> LunarPanel::extensions([
>    ListActivities::class => \Lunar\Admin\Filament\Resources\ActivityResource\Pages\ListActivities::class
> ])
> ```
> 
>  ```
> LunarPanel::panel(fn ($panel) => fn($panel) => $panel->path('admin'))
>     ->extensions([
>        ListActivities::class => \Lunar\Admin\Filament\Resources\ActivityResource\Pages\ListActivities::class
>      ])
>     ->register();
> ```

This is the existing way right? Introducing another facade seems unnecessary.

packages/admin/src/LunarPanelManager.php
glennjacobs: Yeah, as @alecritson said, we should call `registerExtension()` in the loop instead of using `new $extension`.

glennjacobs:
Let's go with `LunarPanel::extensions(` + `->extensions(` approach.

This shortens `registerExtension` and is in line with how Filament name things.

lguichard:
@glennjacobs Voilà 👍

PR #1582, opened 2024-02-23
Docs - add how use custom disk

StephaneBour:
Hello,

I use a particular filesystem for images (https://github.com/sausin/laravel-ovh), and I needed to change the default disk.

I spent some time figuring out how to do this, so I completed the documentation for that.

Thank you.

### DIFF
--- a/docs/core/reference/images.md
+++ b/docs/core/reference/images.md
@@ -121,3 +121,18 @@ class YourCustomModel extends Model
 ```
 
 Now your models will auto generate transforms as defined in your configuration and still use medialibrary under the hood.
+
+## Use custom disk
+
+By default, media is stored on the `public` disk. If you want to use a different disk, you need to set the environment variable `MEDIA_DISK` to the name of the disk you want to use.
+
+```shell
+MEDIA_DISK=s3
+```
+If you need more customisation, you can install publish the configuration file of `spatie/laravel-medialibrary`.
+
+```shell
+php artisan vendor:publish --provider="Spatie\MediaLibrary\MediaLibraryServiceProvider" --tag="config"
+```
+
+Then you can change the disk in the configuration file `config/media-library.php`.


docs/core/reference/images.md
```diff
@@ -121,3 +121,19 @@ class YourCustomModel extends Model
 ```
 
 Now your models will auto generate transforms as defined in your configuration and still use medialibrary under the hood.
+
+## Use custom disk
+
+By default, media is stored on the `public` disk. If you want to use a different disk, you need to set the environment variable `MEDIA_DISK` to the name of the disk you want to use.
+
+```shell
+MEDIA_DISK=ovh
```
wychoong:
```suggestion
MEDIA_DISK=s3
```

use a more common disk as example

docs/core/reference/images.md
```diff
@@ -121,3 +121,19 @@ class YourCustomModel extends Model
...
+```
+If you need more customisation, you can install `spatie/laravel-medialibrary` and publish the configuration file.
```
wychoong: the package already required by lunar, just need to publish the config file

StephaneBour: Hello, unfortunately it did not work when I launched the command. I had to include it in the project.

docs/core/reference/images.md
StephaneBour: There is a tweak in the code to handle different s3s, so if you put it "s3" it works without configuring the MEDIA_DISK variable.
https://github.com/lunarphp/lunar/blob/68398b1920627e18051bc4d6a85bd52ab8408144/packages/admin/src/Http/Livewire/Traits/HasImages.php#L251

StephaneBour:
I updated based on @wychoong 's comments, but I'm not supportive of these changes at now.

glennjacobs (approved):

glennjacobs:
Thanks @StephaneBour

PR #1592, opened 2024-02-29
Properly compute shipping order line total

olivierguerriat:
The `total` column of the order line didn't consider the tax for shipping lines.

I added a few assertions and updated `CalculateTax` to fill every amounts.

I also simplified `ApplyShipping` to avoid recomputing the same amounts and only fill the breakdown. Please check that it makes sense, I'm out of my comfort zone there.


alecritson:
Appreciate this has been hanging around a bit. I'm not sure the `shippingSubTotal` should be in the calculate tax action as we need it before we calculate tax , the sub total needs to be available to any actions that may come between and it's a bit out the scope of what the CalculateTax action is attempting to do.

olivierguerriat:
Thanks for reviewing. Splitting concerns makes sense, I reverted `ApplyShipping` and adapted `CalculateTax` accordingly.

glennjacobs:
Having reviewed the code these are my thoughts...

- We need documentation to reflect these changes
- If there is no shipping, should the shipping values be `0` rather than `null`?
- Maybe we should have a test to cover the no-shipping scenario?

glennjacobs:
@alecritson I think this is probably needed, however, do we need to consider any updates to the Orders side of things?

glennjacobs:
@alecritson do you think we should have all the dynamic totals in the cached properties?

alecritson:
> @alecritson do you think we should have all the dynamic totals in the cached properties?

If it seems like some properties are missing I don't see why we couldn't revisit it in a dedicated PR.

glennjacobs (approved):

PR #1594, opened 2024-02-29
Feature - Refactor order status action

alecritson:
The action to update an orders status and preview the related emails is quite convoluted and the result is what should be a fairly quick action to perform, ends up being a bit of a chore.

For this reason, this PR removes the wizard and preview approach in favour of a more streamlined form, until this can be revisited.


glennjacobs:
@alecritson I think the Additional Content and Email should only show once at least one mailer has been selected.

alecritson:
> @alecritson I think the Additional Content and Email should only show once at least one mailer has been selected.

Pushed some tweaks for this 🤞

glennjacobs:
I feel like we should default this to checked, as otherwise it's an additional click for 99% of orders.

image

alecritson:
> I feel like we should default this to checked, as otherwise it's an additional click for 99% of orders.
> 
> image

This should hopefully be sorted.

glennjacobs (approved):

PR #1600, opened 2024-03-01
Fix - Method remove into ShippingModifiers

lguichard:
The ShippingModifiers store modifiers in collection, the method remove tried to remove the key that didn't exist.


alecritson (commented):
Do we need a new test or to tweak an existing one?

lguichard:
> Do we need a new test or to tweak an existing one?

That's it ;-)

glennjacobs (approved):

PR #1607, opened 2024-03-04
Hotfix - Update ShippingModifier to return closure

alecritson:
We use Pipelines for shipping modifiers, but the one provided by the add on was not using the `$next($cart)` closure, so any additional modifiers where not being triggered.


glennjacobs:
@alecritson do we need to update the docs to reflect this also?

alecritson:
> @alecritson do we need to update the docs to reflect this also?

Docs have been updated

glennjacobs (approved):

PR #1615, opened 2024-03-07
Update OrderResource.php

glennjacobs:
This adds the ability to copy email addresses from the table column, and also polls for new content every 60 seconds. Good for if a new order has come in.

### DIFF
--- a/packages/admin/src/Filament/Resources/OrderResource.php
+++ b/packages/admin/src/Filament/Resources/OrderResource.php
@@ -116,7 +116,7 @@ class OrderResource extends BaseResource
                 ->label(__('lunarpanel::order.table.email.label'))
                 ->toggleable()
                 ->copyable()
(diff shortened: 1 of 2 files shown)


packages/admin/src/Filament/Resources/OrderResource.php
```diff
@@ -113,7 +114,10 @@ public static function getTableColumns(): array
                 ->toggleable(),
             Tables\Columns\TextColumn::make('shippingAddress.contact_email')
                 ->label(__('lunarpanel::order.table.email.label'))
-                ->toggleable(),
+                ->toggleable()
+                ->copyable()
+                ->copyMessage('Email address copied')
```
alecritson: Needs translation :)

glennjacobs: Done!

PR #1616, opened 2024-03-07
Add customer group pricing RM

### LINKED ISSUE #1614: Customer Group Pricing not displayed for minimum quantity is 1
- Lunar version: 1.x-dev
- Laravel Version: 10.46.0
- PHP Version: 8.3
- Database Driver & Version: mySQL8

Hello,

On the admin panel, I have set up Customer groups and pricing for those groups with no minimum quantity.

### Expected Behaviour:
Customer group pricing should be shown even if the quantity is set to 1.

### Actual Behaviour:
The pricing is not displayed when the min_quantity is set to 1.

I have identified that the "Price breaks" table only appears if customer group pricing has tier pricing.

You can find more details in the code at [#Support/RelationManagers/PriceRelationManager.php](https://github.com/lunarphp/lunar/blob/2298d629119b22e8a4332c0005efb522f5201320/packages/admin/src/Support/RelationManagers/PriceRelationManager.php#L99)

### Steps To Reproduce:
Create a random customer group.
Add pricing for that group and set the minimum quantity to 1.

Thanks

wychoong:
fixes #1614

### DIFF
--- a/packages/admin/resources/lang/en/relationmanagers.php
+++ b/packages/admin/resources/lang/en/relationmanagers.php
@@ -162,6 +162,9 @@ return [
             'min_quantity' => [
                 'label' => 'Minimum Quantity',
                 'helper_text' => 'Select the minimum quantity this price will be available for.',
+                'validation' => [
+                    'unique' => 'Customer Group and Minimum Quantity must be unique.',
+                ],
             ],
             'currency_id' => [
                 'label' => 'Currency',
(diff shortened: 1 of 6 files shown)


glennjacobs (changes requested):
This doesn't appear to work for Product Variants. Also, I think the tab name should be "Price Breaks" not "Pricing".

wychoong:
> This doesn't appear to work for Product Variants. Also, I think the tab name should be "Price Breaks" not "Pricing".

whoops

PR #1617, opened 2024-03-07
Use normal search if scout disabled

wychoong:
there are few places in the panel is using scout search for searchable Select field.

this PR is to cater for disabled scout search


packages/admin/src/helpers.php
```diff
@@ -66,3 +72,58 @@ function db_date($column, $format, $alias = null)
         return $select;
     }
 }
+
+if (! function_exists('get_search_builder')) {
+
+    function get_search_builder($model, $search): Laravel\Scout\Builder|Builder
+    {
+        $scoutEnabled = config('lunar.search.scout_enabled', false);
```
glennjacobs: This config value looks wrong; it should be `lunar.panel.scout_enabled`, I think.

See https://github.com/lunarphp/lunar/blob/1.x/packages/admin/config/panel.php#L37

glennjacobs (changes requested):

packages/admin/src/helpers.php
wychoong: Yes, because another PR was merged after this

glennjacobs (approved):

PR #1618, opened 2024-03-07
Remove panel search config

wychoong:
admin and core package both contain config/search.php, and since both are published to config/lunar/*, admin's search.php is override by core's

this pr merged the admin/search into admin/panel config


wychoong:
@glennjacobs as discussed, however I feel it should be in core/search, as all indexers are there, and if we want to disable scout completely (eg: no setup, no indexing), its more natural to use `lunar.search.scout_enabled` than `lunar.panel.scout_enabled`

let me know if you want to change it

glennjacobs:
> @glennjacobs as discussed, however I feel it should be in core/search, as all indexers are there, and if we want to disable scout completely (eg: no setup, no indexing), its more natural to use `lunar.search.scout_enabled` than `lunar.panel.scout_enabled`
> 
> let me know if you want to change it

I see what you're saying, but I don't agree as this setting only affects the admin panel. I don't think we want some sort of master switch to control the core Scout settings.

PR #1622, opened 2024-03-11
While Using simple pagination, it solves string * int issue

Summonshr:
Currently, in ListOrders and ListProducts, when we select 'all' option in pagination, it tends to error stating, string * int 
image

This solves the above issue. Also, snippet is found in the documentaion of source of filamentphp version 3.


glennjacobs:
@Summonshr I've not come across this error before. Can you provide more information on how to replicate it?

glennjacobs:
Leaving open, because I want to understand the problem. But we don't want to use SimplePaginate.

glennjacobs:
Oh, I came across it. For some reason I thought you were referring to the checkbox selection, but you mean the pagination dropdown. 

I'm going to close this, as we have another PR to solve this.

PR #1631, opened 2024-03-13
Laravel 11 Support for Lunar v1 (Clean Install)

adam-code-labx:
This PR gets Lunar v1 up and running with Laravel 11 🚀

I've forked and updated the following repositories:

- marvinosswald/filament-input-select-affix
- leandrocfe/filament-apex-charts
- cartalyst/converter
- kalnoy/nestedset

For Lunar v1 users to benefit from these updates, they will need to add these repositories to their project's composer.json

```json
"repositories": [
    {
        "type": "vcs",
        "url": "https://github.com/adam-code-labx/filament-input-select-affix"
    },
    {
        "type": "vcs",
        "url": "https://github.com/adam-code-labx/filament-apex-charts"
    },
    {
        "type": "vcs",
        "url": "https://github.com/adam-code-labx/converter"
    },
    {
        "type": "vcs",
        "url": "https://github.com/adam-code-labx/laravel-nestedset"
    }
]
```

My next step will be to submit PRs to these repositories, ensuring that everyone can benefit from Laravel 11 compatibility.
Then of course the above repositories will no longer be required.

I've put it through tests on both Laravel 10 and 11 everything passes, updated spatie-media and its enum fill for conversions, tweaked the docs for early opt-in support, and fixed up the spatie permissions migration. 

Also adjusted the GitHub workflow to support Laravel 11 testing.


adam-code-labx:
Not quite sure how to configure your test matrix workflows so will leave this to you guys.
everything is passing locally

CleanShot 2024-03-13 at 02 28 59@2x

Sent minimal PR's to the 2 filament packages.

These 2 however already have Laravel 11 Shift PR's pending
- cartalyst/converter
- kalnoy/nestedset

adevade:
@adam-code-labx Nice work, man! 👍

packages/admin/database/migrations/2023_05_31_100000_create_permission_tables.php
```diff
@@ -25,111 +24,94 @@ public function up()
             throw new \Exception('Error: team_foreign_key on config/permission.php not loaded. Run [php artisan config:clear] and try again.');
         }
 
-        if (! Schema::hasTable($tableNames['permissions'])) {
```
wychoong, line 28: the checking on if tables already exist is required

packages/core/src/Base/StandardMediaDefinitions.php
```diff
@@ -13,7 +13,7 @@ public function registerMediaConversions(HasMedia $model, Media $media = null):
     {
         // Add a conversion for the admin panel to use
         $model->addMediaConversion('small')
-            ->fit(Manipulations::FIT_FILL, 300, 300)
+            ->fit(Fit::Fill, 300, 300)
```
wychoong, line 16: this is new in spatie/image:v3 which is used by media-library:v11
should update composer to only keep v11

wychoong:
one of the annoying upgrade to v11 is in the migration where `->change` doesn't retain existing attributes https://laravel.com/docs/11.x/upgrade#modifying-columns

was this being considered in this PR?

packages/core/src/Base/StandardMediaDefinitions.php
glennjacobs: I was thinking that too, we can't support both versions. We'd have to move to v11

glennjacobs:
Should we not be considering Laravel Shift for this? I'm concerned we may have missed important updates here.

packages/admin/database/migrations/2023_05_31_100000_create_permission_tables.php
adam-code-labx: Note this is just copied over from their updated migration

wychoong: yes, and the check was added during implementation https://github.com/lunarphp/lunar/pull/1092/commits/79bad227446315f28d192aff52bbbb957a5e379c

composer.json
```diff
@@ -11,29 +11,29 @@
...
         "lukascivil/treewalker": "0.9.1",
-        "marvinosswald/filament-input-select-affix": "^0.1.0",
+        "marvinosswald/filament-input-select-affix": "dev-main",
         "php": "^8.1",
         "spatie/laravel-activitylog": "^4.4",
         "spatie/laravel-blink": "^1.6",
-        "spatie/laravel-medialibrary": "^10.0.0",
-        "spatie/laravel-permission": "^5.10",
+        "spatie/laravel-medialibrary": "^10.0|^11.0",
+        "spatie/laravel-permission": "^5.10|^6.4",
```
wychoong: probably can keep just ^6.4

packages/admin/database/migrations/2023_05_31_100000_create_permission_tables.php
adam-code-labx: tbh with this being a migration to create the tables during the initial install I don't see why we need this.
The upgrade path for example should be a new command `lunar:hub:upgrade` this command could then be responsible for upgrading to v1 either by running update migrations on the existing tables or adding additional Database State classes such as `EnsurePermissionsAreUpgraded`

Initially this PR is targeted for a clean install.
I think a new command for the hub upgrade should be considered once we come out of the early alpha stage 👀

packages/core/src/Base/StandardMediaDefinitions.php
adam-code-labx: No problem will update this draft PR today

composer.json
adam-code-labx: As long as this is working still with Laravel 10 then I agree.
Will test first then remove the legacy version constraints 👍

packages/admin/database/migrations/2023_05_31_100000_create_permission_tables.php
glennjacobs: Yep, see #1382

adam-code-labx:
> one of the annoying upgrade to v11 is in the migration where `->change` doesn't retain existing attributes https://laravel.com/docs/11.x/upgrade#modifying-columns
> 
> was this being considered in this PR?

This is currently a draft PR. I will be working further on this today so will make sure I go through the upgrade guide carefully.

@glennjacobs Sure we could try to run Laravel Shift to upgrade to Laravel 11 but I feel this is targeted more for simpler projects, this would certainly suggest using multiple version constraints for different packages. The manual approach imo seems the best option.

I am not sure if Shift will even modify the github test workflows?

adam-code-labx:
Reverted test workflows for now.
Dan always makes it look easy: https://github.com/filamentphp/filament/blob/3.x/.github/workflows/tests.yml
Perhaps this is just an issue with `orchestra/testbench` 🤔

adam-code-labx:
Just updated deps
`Error: Could not authenticate against github.com`

I assume this means shifts must be private 😲
Strange it looks public to me! https://github.com/laravel-shift/converter/tree/l11-compatibility 😂

adam-code-labx:
Can I suggest as v1 is currently in the early alpha stage migrations are simplified merging all the update/add/remove migrations into the create migrations. Then the new lunar upgrade command would perhaps run upgrade migrations from it's own directory. Then we don't even need to worry at this stage about L11 retain existing attributes.

This seems like the best approach for a clean install.
What do you guys think? @wychoong @glennjacobs

packages/admin/database/migrations/2023_05_31_100000_create_permission_tables.php
wychoong: regardless the upgrade path is by a command or migration, the [initial concern](https://github.com/lunarphp/lunar/pull/1092#discussion_r1217609298) was "what if user already have spatie-permission installed before lunar" so either way we need a check. 

after taking a look at spatie's migration migration publish, it seems handled if user already published it before, so now it leave us with 2 option:
- remove the migration file and let the user publish it manually
- the upgrade command takes care of it

problem with first option is the state `EnsurePermissionsAreUpgraded` will run before user can publish spatie's migration

so maybe before we have the new upgrade command, the check for table existence in this migration should remain?

adam-code-labx: Is there any modifications specific to lunar with this migration?

If not then I agree it would be better to remove this and opt-in for the user to publish the latest migrations.
The same would apply to the media table which I can not see any modifications from their default migration stub.

wychoong:
> Can I suggest as v1 is currently in the early alpha stage migrations are simplified merging all the update/add/remove migrations into the create migrations. Then the new lunar upgrade command would perhaps run upgrade migrations from it's own directory. Then we don't even need to worry at this stage about L11 retain existing attributes.
> 
> This seems like the best approach for a clean install. What do you guys think? @wychoong @glennjacobs

I understand the pain to review the columns modifications and to update the migration, but I have to say no to this.
Doing this will make all users upgrading from v0.x -> v1 a big issue, missing all the update/add/remove migrations in between.
A quick search of `->change` in the codebase only have 9 files, and that is the only keyword we need to take care iinm

glennjacobs:
> Can I suggest as v1 is currently in the early alpha stage migrations are simplified merging all the update/add/remove migrations into the create migrations. Then the new lunar upgrade command would perhaps run upgrade migrations from it's own directory. Then we don't even need to worry at this stage about L11 retain existing attributes.
> 
> This seems like the best approach for a clean install. What do you guys think? @wychoong @glennjacobs

Certainly something @alecritson and I are keen to do. The list of migrations is getting quite big and it's the only time we could probably do this.

glennjacobs:
> > Can I suggest as v1 is currently in the early alpha stage migrations are simplified merging all the update/add/remove migrations into the create migrations. Then the new lunar upgrade command would perhaps run upgrade migrations from it's own directory. Then we don't even need to worry at this stage about L11 retain existing attributes.
> > This seems like the best approach for a clean install. What do you guys think? @wychoong @glennjacobs
> 
> I understand the pain to review the columns modifications and to update the migration, but I have to say no to this. Doing this will make all users upgrading from v0.x -> v1 a big issue, missing all the update/add/remove migrations in between. A quick search of `->change` in the codebase only have 9 files, and that is the only keyword we need to take care iinm

Yeah, we would have to solve this somehow programmatically. If it's not possible, then we can't really do it.

Laravel does have it's migration squashing which resolves any performance issues.

packages/admin/database/migrations/2023_05_31_100000_create_permission_tables.php
wychoong: > Is there any modifications specific to lunar with this migration?

no modification apart from the check

adam-code-labx:
@wychoong It's not a problem to update these migrations, the point I was making is this should be part of the new upgrade process. The new upgrade command which would obviously be a separate PR at some point would run only the upgrade migrations and database state classes.

The focus for this early Laravel 11 upgrade should be targeted for a clean install at least until we reach the beta stage.

Of course there is a lot more to consider in terms of upgrading from v0.x. 
This way we can separate the upgrade process.

adam-code-labx:
> Certainly something @alecritson and I are keen to do. The list of migrations is getting quite big and it's the only time we could probably do this.

Happy to work on this and update this PR unless you prefer to work internally on this?

glennjacobs:
> > Certainly something @alecritson and I are keen to do. The list of migrations is getting quite big and it's the only time we could probably do this.
> 
> Happy to work on this and update this PR unless you prefer to work internally on this?

I'm happy for you to make a start on this. The issue doesn't have any description currently, but the main idea is to remove the "state" stuff and instead have an upgrade command that handles that sort of thing and anything else required between versions. We could even look to Filament for inspiration, as they add their upgrade command to the app's composer.json.

adam-code-labx:
> > > Certainly something @alecritson and I are keen to do. The list of migrations is getting quite big and it's the only time we could probably do this.
> > 
> > 
> > Happy to work on this and update this PR unless you prefer to work internally on this?
> 
> I'm happy for you to make a start on this. The issue doesn't have any description currently, but the main idea is to remove the "state" stuff and instead have an upgrade command that handles that sort of thing and anything else required between versions. We could even look to Filament for inspiration, as they add their upgrade command to the app's composer.json.

I have focussed on the clean install 1st.
Latest commit merges all migrations. Also left a copy of the upgrade/merged which the upgrade command will be able to run. (hopefully will get some time this weekend to start on the PR)

The review directory migrations:
- UpdateProductOptionRelations (Perhaps this would be better as state class)
- RemovePositionFromProductOptionsTable (Any reason for this? positioning options is important)
- UpdateProductOptionHandleFk (Surely handles should be unique?)

Only other change is I had to rename the dates on a few tables to work with the constrained.
The only problem I had was with `CreateCartsTable` so for this one I dropped the orders constraint

```php line 18
// @todo Note removed constraint as the orders table comes after the carts table
$table->foreignId('order_id')->nullable();
```

adam-code-labx:
@glennjacobs @alecritson 
Please can you take a look into the github workflows can't seem to get them to work correctly
Perhaps permission issue not sure 🤔

adam-code-labx:
I am going to revert back to my forked repos as laravel-shift causes the authentication issues

adam-code-labx:
@glennjacobs I have solved most of the issues with the workflows (laravel-shift requires tokens so reverted back to my forked repositories for now until the official packages are updated)

The only issue to solve is with laravel-medialibrary. We currently need to support both versions as v11 is required for Laravel 11 however this version also requires php 8.2.

Perhaps the best option here instead of these painful workarounds is to change the requirements for lunar v1 to v8.2? considering 8.1 is only supported until 25 Nov 2024 and the final version I believe will be around September.

glennjacobs:
@adam-code-labx yep, let's move to PHP 8.2+

The workflows can test against 8.2 & 8.3

packages/admin/database/migrations/2023_05_31_100000_create_permission_tables.php
adam-code-labx: Further work will be done for this during the upgrade PR

adam-code-labx:
I will start the upgrade PR next weekend. This should give you guys plenty of time to review this PR 1st.
@glennjacobs @alecritson Let me know if anything else needs to be done on this one.

Thanks have a great weekend 👍

adam-code-labx:
Both my filament package PR's have now been accepted so we only have 2 dependencies pending L11 update.

Here are my suggestions for these 2 remaining packages:
`cartalyst/converter` can be replaced with are own internal converters for each unit of measurement
which we could place for example here `packages/core/src/Base/UnitConverters`

For the exchange rates we should use [laravel-exchange-rates](https://github.com/ash-jc-allen/laravel-exchange-rates) 
Ash has already upgraded this to Laravel 11

`kalnoy/nestedset` I believe this one is used with the Collection model.
With this package under MIT licence and relatively small package we could simply add the NodeTrait and other classes for example here `packages/admin/src/Support/Nestedset`

Let me know your thoughts on this @glennjacobs, if your ok with my suggestions I will add this to next weekend todos 😉

adam-code-labx:
Hi @glennjacobs 
Can we please add this one to the roadmap?
https://github.com/orgs/lunarphp/projects/8/views/8

Any idea when you can review this?
My plan was to work on the upgrade PR this weekend once this one was completed.
Thanks.

glennjacobs:
> Hi @glennjacobs Can we please add this one to the roadmap? https://github.com/orgs/lunarphp/projects/8/views/8
> 
> Any idea when you can review this? My plan was to work on the upgrade PR this weekend once this one was completed. Thanks.

When all the packages are updated, then we can review this. I think there's one left from what I can see?

adam-code-labx:
> > Hi @glennjacobs Can we please add this one to the roadmap? https://github.com/orgs/lunarphp/projects/8/views/8
> > Any idea when you can review this? My plan was to work on the upgrade PR this weekend once this one was completed. Thanks.
> 
> When all the packages are updated, then we can review this. I think there's one left from what I can see?

Yep one left but tbh I think we can replace this with our own internal measurement converters. (Cartalyst doesn’t seem to be actively maintained 1 year since there last commit)

“cartalyst/converter can be replaced with are own internal converters for each unit of measurement
which we could place for example here packages/core/src/Base/UnitConverters”

glennjacobs:
> > > Hi @glennjacobs Can we please add this one to the roadmap? https://github.com/orgs/lunarphp/projects/8/views/8
> > > Any idea when you can review this? My plan was to work on the upgrade PR this weekend once this one was completed. Thanks.
> > 
> > 
> > When all the packages are updated, then we can review this. I think there's one left from what I can see?
> 
> Yep one left but tbh I think we can replace this with our own internal measurement converters. (Cartalyst doesn’t seem to be actively maintained 1 year since there last commit)
> 
> “cartalyst/converter can be replaced with are own internal converters for each unit of measurement which we could place for example here packages/core/src/Base/UnitConverters”

That could work, as UOM conversions are never going to change :)

adam-code-labx:
I will try to find some time this week.
@glennjacobs would you prefer the converters to be a separate PR or just push the changes here?

glennjacobs:
> I will try to find some time this week. @glennjacobs would you prefer the converters to be a separate PR or just push the changes here?

Might be easier to check a separate PR, thanks.

adam-code-labx:
Thanks @wychoong cartalyst updated their package now to support L11
https://github.com/cartalyst/converter/pull/32#issuecomment-2017658190

I still believe it will be better to create are own implementation, however right now we can at least move forward with this PR.

composer.json
```diff
@@ -11,35 +11,35 @@
...
         "ext-intl": "*",
         "filament/filament": "^3.2.25",
-        "filament/spatie-laravel-media-library-plugin": "^3.0-stable",
+        "filament/spatie-laravel-media-library-plugin": "^3.2",
         "guzzlehttp/guzzle": "^7.3",
-        "illuminate/http": "^10.0",
-        "kalnoy/nestedset": "^6.0",
-        "laravel/framework": "^10.0",
+        "illuminate/http": "^10.0|^11.0",
+        "kalnoy/nestedset": "^v6.x-dev",
```
glennjacobs, line 23: This isn't a true release. We will need either `6.0.4` or `7.0.0` tagged.

adam-code-labx: Unfortunately this was never tagged just updated in their dev branch https://github.com/lazychaser/laravel-nestedset/issues/598#issuecomment-2011851119

glennjacobs: They will tag it eventually.

adam-code-labx: At least we are not using a forked branch anymore. Also as lunar v1 is still in alpha, I would not delay this based on using their dev branch when we can simply provide a patch update once they finally get round to tagging their release 👀

glennjacobs: I don't mind doing that as long as we create an issue so we don't forget :)

adam-code-labx:
@glennjacobs How about we trim down this PR and just focus only on the Laravel 11 upgrade so this can be released perhaps with 1.0.0-alpha.9 this week?

We can save all migration-related modifications for the Lunar install which I can then go ahead and start a draft PR if this works for you?

glennjacobs:
> @glennjacobs How about we trim down this PR and just focus only on the Laravel 11 upgrade so this can be released perhaps with 1.0.0-alpha.9 this week?
> 
> 
> 
> We can save all migration-related modifications for the Lunar install which I can then go ahead and start a draft PR if this works for you?

Yes I agree.

adam-code-labx:
@glennjacobs Sure will get on to this tonight and start the PR for the Lunar install.

Thinking we can use Laravel prompts something similar to my new RESTPresenter package 😜
https://github.com/xtend-packages/rest-presenter/blob/main/src/Commands/RESTPresenterSetupCommand.php

adam-code-labx:
@glennjacobs Restoring the migrations resulted in all tests failing.

Let's keep this PR targeted only for clean installs, focused on updating workflows and test cases, while also ensuring we have cleaned up the migrations to leave us only with create tables, previous updated migrations have now been merged.

The next PR for the lunar-install will handle the upgrade process. However this one might not be straight forward and will certainly take time to get right, this really shouldn't delay L11 support. 

We just to need to make sure everyone is aware that this is currently for clean installs and updating from previous v1 alpha and not currently supported to be used when upgrading from 0.x. 

Perhaps as previously suggested this could be released and tagged to another branch.
for example `"lunarphp/lunar": "feature/laravel-11"`

glennjacobs:
@adam-code-labx many thanks for all your efforts. I reviewed the PR and decided it would be quicker for me to start on a clean branch to get this done, taking influence from your updates.

The new PR can be found here #1685 

The migrations and upgrade command can be discussed on #1382 to determine how we want to move forward.

PR #1633, opened 2024-03-13
Re render price tables when base prices change

ryanmitchell:
Totally get that this is specific to us, but it would be lovely to force a re-render the customer group / tier pricing tables when the base prices changes (as we do some dependent calculations in the background).

Pretty please?


glennjacobs:
@ryanmitchell no ;-)

This is too specific to your requirements. Maybe we could add a generic listener to all relation managers for them to refresh, and then you use a page extension to add a lifecycle event on the field that fires the livewire event?

Just an idea, no idea if it would work.

ryanmitchell:
Ok - let me unpack that and I'll update this when I get it working.

ryanmitchell:
The problem with the suggested approach is that the trigger is coming from the pricing relation manager form, not the resource form, so adding a lifecycle event doesn't seem to be possible.

What about making the event more generic ('refresh-relation-tables') but still dispatching it from the same place?

glennjacobs:
> The problem with the suggested approach is that the trigger is coming from the pricing relation manager form, not the resource form, so adding a lifecycle event doesn't seem to be possible.
> 
> What about making the event more generic ('refresh-relation-tables') but still dispatching it from the same place?

I don't see what the Filament resource has to do with it? Do you have code to show what you've tried so far?

ryanmitchell:
Sure, so I have a extended product variant resource:

```php
<?php

namespace App\Lunar\Extensions;

use Filament\Forms;
use Lunar\Admin\Support\Extending\ResourceExtension;

class ProductVariantResourceExtension extends ResourceExtension
{
    public function extendForm(Forms\Form $form): Forms\Form
    {
      // form schema stuff here thats not relevant but all works fine on the default form
    }

    public function getRelations(array $managers) : array
    {
        dd($managers); // <--- never fires
    }
}
```

I presume what I'd need to do is make my own pricing manager that extends the default one, and add a hydration somewhere, but I cant access it through this extension method. 

I'm trying to trigger an update when the compare_at price of the variant changes.

ryanmitchell:
Ok, got it working as follows:

```
<?php

namespace App\Lunar\Extensions\Pages;

use Filament\Forms;
use Illuminate\Database\Eloquent\Model;
use Lunar\Admin\Support\Extending\EditPageExtension;

class ManageVariantPricing extends EditPageExtension
{
    private $form;

    public function extendForm(Forms\Form $form): Forms\Form
    {
        $this->form = $form;

        return $form;
    }

    public function afterUpdate(Model $record, array $data): Model {
        $this->form->getLivewire()->dispatch('base-prices-updated');

        return $record;
    }
}
```

Part of the issue was that `ManagesProductPricing` wasnt calling the beforeSave/afterSave hooks so I'll PR that seperately.

What do you want to call this event? 'refresh-relation-tables' or something?

packages/admin/src/Support/RelationManagers/BaseRelationManager copy.php
```diff
@@ -0,0 +1,15 @@
...
+use Filament\Resources\RelationManagers\RelationManager;
+use Livewire\Attributes\On;
+
+class BaseRelationManager extends RelationManager
+{
+    #[On('refresh-relation-manager')]
+    public function refreshRelationManager()
+    {
+        // force a re-render
+    }
```
wychoong: this can be removed by having the #[On] attribute on component level

ref: https://github.com/livewire/livewire/pull/7667

ryanmitchell: Nice. Fairly sure @glennjacobs is going to reject this PR either way :)

wychoong: i think if you solved the beforeSave/afterSave hooks issue it can be a nice addition 😜

glennjacobs: @ryanmitchell I think this could be quite handy. Does anything need updating on this PR now?

ryanmitchell: It’s works as is. It’s just whether you want the change @wychoong suggested .

glennjacobs: @wychoong can you give me an example of how the code would change with your suggestion above? I don't have experience with it at all.

alecritson: Just my two cents, I think since we hardly use PHP attributes anywhere else (just in the order manager that I can see) maybe we should avoid for now to keep things consistent.

packages/admin/src/Support/RelationManagers/BaseRelationManager.php
```diff
@@ -0,0 +1,15 @@
...
+use Livewire\Attributes\On;
+
+class BaseRelationManager extends RelationManager
+{
+    #[On('refresh-relation-manager')]
+    public function refreshRelationManager()
+    {
+        // force a re-render
+    }
+}
```
wychoong:
```suggestion
#[On('refresh-relation-manager')]
class BaseRelationManager extends RelationManager
{
    //
}
```

@glennjacobs this should work

packages/admin/src/Support/Concerns/Products/ManagesProductPricing.php
```diff
@@ -70,6 +70,8 @@ protected function handleRecordUpdate(Model $record, array $data): Model
 
         $this->basePrices = $this->getBasePrices($variant);
 
+        $this->dispatch('base-prices-updated');
```
wychoong: @ryanmitchell is this intended or it should be dispatch `refresh-relation-manager` for RelationManager?

ryanmitchell: Yep it should, thanks!

packages/admin/src/Support/RelationManagers/BaseRelationManager.php
glennjacobs: @ryanmitchell does that work for you?

ryanmitchell: I've updated it - I've no way of testing at the moment so I'm taking @wychoong's word for it!

glennjacobs:
@ryanmitchell this is causing an error for me, that doesn't exist on 1.x. This is when saving the prices.

image

ryanmitchell:
As I said I can’t test at the moment. Was it working before the last change?

glennjacobs:
> As I said I can’t test at the moment. Was it working before the last change?

Same issue, unfortunately.

ryanmitchell:
Can you give me a full stack trace to work with?

ryanmitchell:
@glennjacobs just tested it and I can't recreate your error on the product pricing page, the updating works fine for me.

glennjacobs (approved):

PR #1637, opened 2024-03-13
Move default to same name column, and remove deprecated BooleanColumn…

Summonshr:
The column default can me merged with name like this as the column only has valid value for one row. Also replaced BooleanColumn with IconColumn.

image


glennjacobs:
How would someone know that the icon represents the default option?

Summonshr:
May be need some other icon or add tooltip on it. Not sure what would be appropriate.

badge may be. stating, it's default.

glennjacobs:
Updated to use a badge instead

image

alecritson (approved):

PR #1647, opened 2024-03-14
Only load lunar filament styles inside filament

kylekanderson:
This PR prevents the package's `lunar-panel.css` file from being included outside of Filament.

Currently, the `LunarPanelProvider` has a `registerPanelAssets` method that registers this file as a Filament Asset. By doing so, it's included anywhere the `@filamentStyles` blade directive is used.

For any sites that implement Filament on their storefront (for example, by using Filament's Form Builder or Notifications features), this `@filamentStyles` blade directive is likely added to their various blade layout files (as instructed by the Filament documentation). However, this will result in the style rules defined in the `lunar-panel.css` file being included on the storefront pages. And in that case, those style rules may inadvertently overwrite existing style rules set on the user's storefront.

To prevent this, this PR wraps the registration of the `lunar-panel.css` file inside of the `Lunar::isServing()` conditional. By doing so, the style is only included on Filament pages and not on the storefront.

### DIFF
--- a/packages/admin/src/LunarPanelProvider.php
+++ b/packages/admin/src/LunarPanelProvider.php
@@ -2,6 +2,7 @@
 
 namespace Lunar\Admin;
 
+use Filament\Facades\Filament;
 use Filament\Support\Assets\Css;
 use Filament\Support\Events\FilamentUpgraded;
 use Filament\Support\Facades\FilamentAsset;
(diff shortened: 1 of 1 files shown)


glennjacobs:
I tested this PR locally, but the styles no longer seem to load in the Lunar panel.

image

kylekanderson:
Okay, this should be resolved now. I just switched from using the `Filament::isServing()` method to the `Filament::serving()` method, which allows you to pass a callback function that is called when inside Filament.

glennjacobs (approved):

PR #1648, opened 2024-03-14
Display shared product options

lguichard:
The shared product options list didn't display shared options. Now the user can translate and manage all product options.


alecritson:
Hi @lguichard thanks for this, if we want non shared options to be editable and shown they will need to have restricted editing, since they are managed via the product. I think if they are shared we shouldn't be able to change anything other than the translatable fields.

lguichard:
> Hi @lguichard thanks for this, if we want non shared options to be editable and shown they will need to have restricted editing, since they are managed via the product. I think if they are shared we shouldn't be able to change anything other than the translatable fields.

That's it !

glennjacobs (approved):

Issue #1657, opened 2024-03-19
Error when attaching `discount` to a `customerGroup` in the Admin Hub


ha-zeek:
- Lunar version: 1.0.0-alpha.7
- Laravel Version: 10.48.3
- PHP Version: 8.1
- Database Driver & Version: MySQL 8.2

### Expected Behaviour:
The `Discount` should be attached to a `CustomerGroup` .

### Actual Behaviour:
an error is prompted (refer image). 

![image](https://github.com/lunarphp/lunar/assets/108339521/5c4dd0a1-d556-4979-8040-e8a487ecf93d)

There is the `customerGroups` relationship in the `Lunar\Models\Discount`, however it is calling the relationship from `Lunar\Models\CustomerGroup` instead.

### Steps To Reproduce:
In Admin Hub, create a discount and attach it to a customerGroup.

ryanmitchell:
A similar error happens with scheduling channel availability

ryanmitchell:
Can be closed now https://github.com/lunarphp/lunar/pull/1658 has merged

PR #1658, opened 2024-03-19
Define inverse discount relationships

ryanmitchell:
Define the inverse customerGroups and channels relationships on discounts so the Filament relation managers work.

Closes https://github.com/lunarphp/lunar/issues/1657

### DIFF
--- a/packages/core/src/Models/Channel.php
+++ b/packages/core/src/Models/Channel.php
@@ -68,6 +68,20 @@ class Channel extends BaseModel
     }
 
     /**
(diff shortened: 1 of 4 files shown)


alecritson:
Thanks @ryanmitchell I've added some tests for this quickly, if they all greenlight and you're happy with the approach we can get this in :)

ryanmitchell:
looks good to me

alecritson (approved):

alecritson (approved):

PR #1661, opened 2024-03-20
Model extending

alecritson:
This PR looks to build on the current work done for model extending and includes but not limited to the following changes:

- Filament resources use the model contract for DI
- Change to use Laravel's `morphMap` instead of having concrete classes in the database for polymorphic relationships
- There was an issue with custom models when they extended Lunar ones, the `newFactory` method didn't like having the return type.
- Added a `addDir`
- Tweaked `guessContractClass` to take namespace into account as not all Lunar models (from addons) will have the same namespace
- Added `addDirectory` method to prevent excessive calls to `replace` or `add` methods.
- Updated relationships to use `modelClass()`
- Static functions should forward to the custom class when called.
- If a custom class name is different to the one it extends i.e. `App\Models\MyProduct` instead of `App\Models\Product` it will still use the correct table name.
- Tweaked `observe` method on models that use `HasModelExtending` to forward to the custom class even if observe is called on the Lunar model.
- Removed redundant tests and updated docs.

Set as draft for the moment as there is a fair amount of changes so just want to double check before hitting the button.


packages/core/database/migrations/2024_03_15_100000_remap_polymorphic_relations.php
```diff
@@ -0,0 +1,75 @@
...
+                }
+            }
+        }
+    }
+
+    public function down()
+    {
+        Schema::table($this->prefix.'prices', function (Blueprint $table) {
+            $table->renameColumn('min_quantity', 'tier');
+        });
```
olivierguerriat: This appears to be leftover from a copy/paste.

packages/core/src/Base/ModelManifest.php
```diff
@@ -88,10 +112,68 @@ protected function bindingName(string $modelClass): string
         return Str::camel($shortName);
     }
 
-    public function guessContractClass(string $modelClass)
+    public function guessContractClass(string $modelClass): string
     {
-        $shortName = (new \ReflectionClass($modelClass))->getShortName();
+        $class = (new \ReflectionClass($modelClass));
```
mattdfloyd:
```suggestion
        $class = new \ReflectionClass($modelClass);
```

packages/core/src/Base/ModelManifest.php
```diff
@@ -88,10 +112,68 @@ protected function bindingName(string $modelClass): string
...
+    {
+        $modelClasses = collect(
+            Discover::in(__DIR__.'/../Models')
+                ->classes()
+                ->extending(BaseModel::class)
+                ->get()
+        )->mapWithKeys(
+            fn ($class) => [
+                \Illuminate\Support\Str::snake(class_basename($class)) => $class::modelClass(),
+            ]
```
mattdfloyd: It would be convenient to be able to prefix all morphMap names through a config setting. Didn't look to see how this would affect the rest of the changes but I assume it's fairly similar to how table names can be prefixed.

mattdfloyd (commented):
Looking forward to this! Look like it will resolve the paper cuts I've run into just recently with Model extension. Thanks!

docs/core/reference/carts.md
```diff
@@ -47,18 +47,19 @@ Lunar\Models\CartLine
...
 | updated_at       |                                              |
 | meta             | JSON data for saving any custom information. |
 
 ```php
+$purchasable = \Lunar\Models\ProductVariant::create([/** ... */]);
 $cartLine = new \Lunar\Models\CartLine([
     'cart_id' => 1,
-    'purchasable_type' => ProductVariant::class,
-    'purchasable_id' => 123,
+    'purchasable_type' => $purchasable->getMorhClass(),
```
glennjacobs: Typo? Should it be `getMorphClass()`?

docs/core/upgrading.md
```diff
@@ -18,6 +18,47 @@ php artisan migrate
...
 
+## [Unreleased]
+
+### High Impact
+
+#### Model Extending
+
+Model extending has been completely rewritten and will require changes to your Laravel app if you have previously extended Lunar models.
+
+The biggest difference is now Lunar models implement a contract (interface) and support Dependency injection across your storefront and the Lunar panel.
```
glennjacobs: `Dependency` should be lower case.

packages/core/src/Base/ModelManifest.php
alecritson: Thanks for the suggestion, I think this is a good option to have and I've tweaked the PR to hopefully accommodate this :)

glennjacobs:
@alecritson I've updated the parent branch and we can now see we have some conflicts. And we know we probably need to update Brands from your other PR that was merged.

alecritson:
> @alecritson I've updated the parent branch and we can now see we have some conflicts. And we know we probably need to update Brands from your other PR that was merged.

The conflicts should all be sorted 🤞

lguichard:
@alecritson, you have missing to change Filament model in the ressources by contract :)

AidasK:
InstallLunar class should be also updated right? It currently uses ` 'attributable_type' => Product::class,` and instead it should pick the morphed name.

packages/core/database/factories/BaseFactory.php
```diff
@@ -0,0 +1,13 @@
...
+
+namespace Lunar\Database\Factories;
+
+use Illuminate\Database\Eloquent\Factories\Factory;
+
+abstract class BaseFactory extends Factory
+{
+    public function modelName()
+    {
+        return (new $this->model)::modelClass();
```
glennjacobs, line 11: From what I can see, this is never used. Do we need it?

packages/core/src/LunarServiceProvider.php
```diff
@@ -175,6 +175,8 @@ public function register(): void
         $this->app->singleton(DiscountManagerInterface::class, function ($app) {
             return $app->make(DiscountManager::class);
         });
+
+        \Lunar\Facades\ModelManifest::register(); // Facade is the same name as the class, change??
```
glennjacobs: Looks like we need to address the comment.

packages/core/src/Models/Address.php
```diff
@@ -34,14 +34,14 @@
  * @property ?\Illuminate\Support\Carbon $created_at
  * @property ?\Illuminate\Support\Carbon $updated_at
  */
-class Address extends BaseModel implements Addressable
+class Address extends BaseModel implements \Lunar\Models\Contracts\Address, Addressable
```
glennjacobs: Considering we're in the models directory, this could be changed simply to `Contracts\Address`. Same for the other models. It saves on having to import the interface whilst slimming it down.

packages/table-rate-shipping/src/Models/ShippingExclusion.php
```diff
@@ -3,11 +3,12 @@
...
 
 use Illuminate\Database\Eloquent\Factories\HasFactory;
+use Illuminate\Database\Eloquent\Relations\BelongsTo;
+use Illuminate\Database\Eloquent\Relations\MorphTo;
 use Lunar\Base\BaseModel;
-use Lunar\Shipping\Facades\Shipping;
 use Lunar\Shipping\Factories\ShippingExclusionFactory;
 
-class ShippingExclusion extends BaseModel
+class ShippingExclusion extends BaseModel implements \Lunar\Shipping\Models\Contracts\ShippingExclusion
```
glennjacobs: Same here, this could be `Contracts\ShippingExclusion`

docs/core/extending/models.md
```diff
@@ -8,233 +8,125 @@ Lunar provides a number of Eloquent Models and quite often in custom application
...
 :::
 
-## Extendable Models
-All Lunar models are now extendable.
-This means you can now add your own functionality or change out existing core model behaviour using your own model implementations.
+## Replaceable Models
+All Lunar models are replaceable, this means you can instruct Lunar to use your own custom model, throughout the ecosystem, using dependency injection.
+
 
 ### Registration:
```
glennjacobs: Not your doing, but we need to remove the `:` from the heading.

docs/core/extending/models.md
```diff
@@ -8,233 +8,125 @@ Lunar provides a number of Eloquent Models and quite often in custom application
...
+### Static call forwarding
 
-class Product extends \Lunar\Models\Product
-{
-    use SearchableTrait;
-}
-```
-Note: shouldBeSearchable could also be overridden by adding directly to the Product class above.
-In this example we are showing you how the core model can be made aware of your own model and trait methods.
+If you have custom methods in your on model, you can call those functions directly from the Lunar model instance.
```
glennjacobs: Should be `own model`?

docs/core/extending/models.md
```diff
@@ -8,233 +8,125 @@ Lunar provides a number of Eloquent Models and quite often in custom application
...
-    protected function isSameAddress(): bool
-    {
-        return $this->billingAddress->first()->id === $this->shippingAddress->first()->id;
-    }
-}
+```php
+\Lunar\Models\Product::observe(/** .. */);
 ```
 
 ## Dynamic Eloquent Relationships
```
glennjacobs, line 132: I feel like we need to explain that dynamic relationships are for when you don't want/need to create your own model. They are separate from model extending.

alecritson: Tweaked the wording

packages/core/database/factories/BaseFactory.php
alecritson: It's used by Laravel

packages/core/src/LunarServiceProvider.php
alecritson: I've removed the comment, not sure what needs changing as it's a common pattern in Laravel?

AidasK:
Kudos for all the work <3. But eta for this? Looks like the most needed and essential feature of v1

glennjacobs:
> Kudos for all the work <3. But eta for this? Looks like the most needed and essential feature of v1

It'd essentially ready and on a PR. We are just trying to get some more PRs merged before that one goes in.

RRosalia:
Oeh we need this one

glennjacobs:
@alecritson going to need to fix the conflicts before this can be merged.

PR #1667, opened 2024-03-21
Use Lunar Migration class for media table

FinnPaes:
Currently this migration extends the default migration class of Laravel. The normal migration class does not set the connection from the Lunar connection of course.

This will cause errors in the upcoming migration `2022_08_09_100000_create_media_variant_table`, because this table references to the normal `media` table.
This only happens if using another connection for Lunar tables.

### DIFF
--- a/packages/core/database/migrations/2021_08_10_101547_create_media_table.php
+++ b/packages/core/database/migrations/2021_08_10_101547_create_media_table.php
@@ -1,6 +1,6 @@
 <?php
 
-use Illuminate\Database\Migrations\Migration;
+use Lunar\Base\Migration;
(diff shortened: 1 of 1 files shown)


FinnPaes:
I would also like to see this change in 1.x (which I actually discovered this issue in). Should I open a seperate PR for 1.x as well?

glennjacobs:
> I would also like to see this change in 1.x (which I actually discovered this issue in). Should I open a seperate PR for 1.x as well?

Assuming this is merged, it will then be merged into 1.x after.

glennjacobs (approved):

glennjacobs:
Thanks @FinnPaes

PR #1670, opened 2024-03-25
Fix - Product variants widget saving

lguichard:
This PR fixes when product variants saving avoid to erase data of product options shared.

### DIFF
--- a/packages/admin/src/Filament/Resources/ProductResource/Widgets/ProductOptionsWidget.php
+++ b/packages/admin/src/Filament/Resources/ProductResource/Widgets/ProductOptionsWidget.php
@@ -283,7 +283,7 @@ class ProductOptionsWidget extends BaseWidget implements HasActions, HasForms
 
             $optionValue = $option['value'];
 
-            if (! $optionModel->exists) {
+            if (! $optionModel->shared) {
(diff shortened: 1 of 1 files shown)


alecritson:
@lguichard Not too sure on this one as we want to allow users to edit the product option values which are linked  to the product (not shared) without having to go through the settings screen.

lguichard:
In this case, there are the conflict with data save in settings screen and product screen. We should fix production variants edition.

alecritson:
> In this case, there are the conflict with data save in settings screen and product screen. We should fix production variants edition.

I don't think it's so much of a problem if, when editing through the settings screen, it's restricted, since you should mainly be editing non shared product options via the product.

lguichard:
@alecritson Humm, i think w’re talking about the same thing. My initial commit fix the bevahiour below :

- if someone has save the shared product option with handle = color and name fr : couleur / en : color. If you generate new variants in edit product the handle and localize fields are erase with new data.

I agree with all anothers points.

lguichard:
@alecritson My bad, i changed the condition, I just noticed that write a bad condition ^^

alecritson (approved):

PR #1676, opened 2024-03-29
Fix - Update MediaRelationManager

lguichard:
This PR cleaning the code, add translation and limit the name of file that fix edit view.

I moved into the new class ManageMediasRelatedRecords to avoid duplicate code.


packages/admin/src/Filament/Resources/CollectionResource/Pages/ManageCollectionMedia.php
```diff
@@ -2,56 +2,10 @@
...
     protected static string $resource = CollectionResource::class;
-
-    protected static string $relationship = 'media';
-
-    public function getTitle(): string
-    {
-        return __('lunarpanel::collection.pages.media.label');
-    }
-
-    public function getBreadcrumbs(): array
```
alecritson, line 22: Looks like the breadcrumbs got caught up in the purge, collections have slightly different logic for theirs so would need to be kept in tact :)

lguichard: @alecritson Fixed

glennjacobs (approved):

PR #1685, opened 2024-04-05
Laravel 11 Compatibility

glennjacobs:
Support Laravel 11

- [x] CHORE - Check all `->change()` migration entries.
- [ ] Create an issue to update nestedset package once a release has been tagged.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,7 +41,7 @@
         "laravel/pint": "1.13.1",
         "mockery/mockery": "^1.4.4",
         "nunomaduro/larastan": "^2.0",
-        "orchestra/testbench": "^8.0",
+        "orchestra/testbench": "^8.0|^9.0",
         "pestphp/pest": "^2.0",
         "pestphp/pest-plugin-laravel": "^2.0",
         "symplify/monorepo-builder": "^10.0"
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-    "name": "lunarphp/monorepo",
+    "name": "lunarphp/lunar",
     "description": "Lunar Monorepo",
     "license": "MIT",
     "authors": [
--- a/composer.json
+++ b/composer.json
@@ -58,18 +58,9 @@
             "Lunar\\Database\\Factories\\": "packages/core/database/factories",
             "Lunar\\Database\\Seeders\\": "packages/core/database/seeders",
             "Lunar\\Database\\State\\": "packages/core/database/state",
-            "Lunar\\Meilisearch\\": [
-                "packages/meilisearch/src",
-                "packages/meilisearch/src/"
-            ],
-            "Lunar\\Opayo\\": [
-                "packages/opayo/src",
-                "packages/opayo/src/"
-            ],
-            "Lunar\\Paypal\\": [
-                "packages/paypal/src",
-                "packages/paypal/src/"
-            ],
+            "Lunar\\Meilisearch\\": "packages/meilisearch/src/",
+            "Lunar\\Opayo\\": "packages/opayo/src/",
+            "Lunar\\Paypal\\": "packages/paypal/src/",
             "Lunar\\Shipping\\": "packages/table-rate-shipping/src",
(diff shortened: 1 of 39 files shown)


composer.json
```diff
@@ -57,6 +58,10 @@
             "Lunar\\Database\\Factories\\": "packages/core/database/factories",
             "Lunar\\Database\\Seeders\\": "packages/core/database/seeders",
             "Lunar\\Database\\State\\": "packages/core/database/state",
+            "Lunar\\Meilisearch\\": [
```
alecritson: Some weird duplication here.

packages/admin/database/migrations/2023_05_31_100000_create_permission_tables.php
```diff
@@ -25,111 +24,94 @@ public function up()
...
         }
 
-        if (! Schema::hasTable($tableNames['permissions'])) {
-            Schema::create($tableNames['permissions'], function (Blueprint $table) {
-                $table->bigIncrements('id'); // permission id
-                $table->string('name');       // For MySQL 8.0 use string('name', 125);
-                $table->string('guard_name'); // For MySQL 8.0 use string('guard_name', 125);
-                $table->timestamps();
-
+        Schema::create($tableNames['permissions'], function (Blueprint $table) {
```
alecritson: Is there a reason for removing the `hasTable` check?

packages/core/database/migrations/2022_01_12_100000_add_columns_to_attributes_table.php
```diff
@@ -18,6 +18,8 @@ public function up()
     public function down()
     {
         Schema::table($this->prefix.'attributes', function (Blueprint $table) {
+            $table->dropIndex($this->prefix.'attributes_searchable_index');
```
alecritson: Rather than having to specify the full index name, would it not be nicer to pass as an array.

```php
$table->dropIndex(['searchable']);
```

https://laravel.com/docs/11.x/migrations#dropping-indexes

packages/admin/database/migrations/2023_05_31_100000_create_permission_tables.php
glennjacobs: This is a copy/paste job from the Spatie package, it's not my code.

packages/core/database/migrations/2022_01_12_100000_add_columns_to_attributes_table.php
glennjacobs: I tried that, but it didn't like it.

composer.json
glennjacobs: Weird!

glennjacobs: That should be sorted now.

packages/admin/database/migrations/2023_05_31_100000_create_permission_tables.php
glennjacobs: Do we need that check? Was it something we added?

packages/core/database/migrations/2022_01_12_100000_add_columns_to_attributes_table.php
glennjacobs: Actually, maybe I didn't use the array syntax. I'll try it again,

packages/admin/database/migrations/2023_05_31_100000_create_permission_tables.php
alecritson: Looks like it was there from the start, I think it's in the event the developer also publishes the media library migrations there's a potential conflict there.

packages/core/database/migrations/2022_01_12_100000_add_columns_to_attributes_table.php
glennjacobs: Should be sorted now.

packages/admin/database/migrations/2023_05_31_100000_create_permission_tables.php
glennjacobs: Added it back in.

alecritson (dismissed):

PR #1688, opened 2024-04-10
[1.x] Remove `nunomaduro/larastan`

whoami15:
This PR removes `nunomaduro/larastan` from dependencies. The package `larastan/larastan` is now being used instead.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -39,7 +39,6 @@
         "larastan/larastan": "^2.9",
         "laravel/pint": "1.13.1",
         "mockery/mockery": "^1.4.4",
-        "nunomaduro/larastan": "^2.0",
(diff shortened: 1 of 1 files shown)


alecritson (approved):

adevade:
@alecritson This gets added automatically by the `monorepo-builder.php` script. What's the difference between the packages defined in there and the ones defined directly in `composer.json`?

glennjacobs:
> @alecritson This gets added automatically by the `monorepo-builder.php` script. What's the difference between the packages defined in there and the ones defined directly in `composer.json`?

@alecritson this looks like a mistake. It should have been updated here https://github.com/lunarphp/lunar/blob/1.x/monorepo-builder.php#L44

PR #1691, opened 2024-04-11
Disable certain attribute fields when editing

### LINKED ISSUE #1655: Error when saving a product with a `Lunar\FrieldTypes\TranslatedText` attribute
- Lunar version: 1.0.0-alpha.7
- Laravel Version: 10.48.3
- PHP Version: 8.1.27
- Database Driver & Version: MySql 8

### Expected Behaviour:
Save your product and the data is saved

### Actual Behaviour:
When you save the product you get an error

### Steps To Reproduce:
Add an Attribute Group with an attribute of type `Lunar\FrieldTypes\TranslatedText` to a product type.

Try to save the product.

You get the error

`Lunar\FieldTypes\TranslatedText value must be a collection.`

### Additional info:

I think the error occurs since this fix but I'm not sure: https://github.com/lunarphp/lunar/pull/1639

Looks like I'm not able to save the description field either, but there's no error.

image

alecritson:
The hub doesn't really support the changing of certain attribute columns once hey have been created. This is because it's hard to keep the `attribute_data` in sync with what the new fieldtype might require, leading to quite an overhead for something seemingly simple.

This PR will disable the `handle` and `type` columns when editing an attribute to preserve and prevent data issues. If a developer needs to change the fieldtype then it should be done in a migration or similar so that appropriate steps can be taken to ensure the data is migrated properly.

As a side effect this should close #1655


wychoong:
this might not solve the issue of changed attribute type.
for example, if the user is using the panel purely, and created a TranslatedText attribute `summary`, and later on deleted it, and created new TextField with the same handle `summary`. It essentially is the same of changing the attribute type, and the form is generated as TextField input but the data is cast/hydrated as TranslatedText

probably should add a checker/validate in `formatStateUsing` to nullify it if the type is different
https://github.com/lunarphp/lunar/blob/25bf71f927460a8329a45082bc6f5bd5690706d9/packages/admin/src/Support/Forms/AttributeData.php#L54

if adding a new method to FieldTypes, maybe we can have a sort of transformer to handle a basic conversion like "text field value" -> ['en' => "text field value"] and vice versa

glennjacobs:
> this might not solve the issue of changed attribute type. for example, if the user is using the panel purely, and created a TranslatedText attribute `summary`, and later on deleted it, and created new TextField with the same handle `summary`. It essentially is the same of changing the attribute type, and the form is generated as TextField input but the data is cast/hydrated as TranslatedText
> 
> probably should add a checker/validate in `formatStateUsing` to nullify it if the type is different
> 
> https://github.com/lunarphp/lunar/blob/25bf71f927460a8329a45082bc6f5bd5690706d9/packages/admin/src/Support/Forms/AttributeData.php#L54
> 
> if adding a new method to FieldTypes, maybe we can have a sort of transformer to handle a basic conversion like "text field value" -> ['en' => "text field value"] and vice versa

This is very true. @alecritson I think we need to consider this.

glennjacobs (changes requested):
See comments made.

alecritson:
@wychoong I've pushed up a couple of tweaks based on what I think you mean. As for changing the format of the data from one to the other I think that's out the scope of what this will be trying to do. We have spoke about it in the past and I think migrating data based on whether a fieldtype changes should be it's own unit of work with it's own considerations. 🤔

wychoong:
> @wychoong I've pushed up a couple of tweaks based on what I think you mean. As for changing the format of the data from one to the other I think that's out the scope of what this will be trying to do. We have spoke about it in the past and I think migrating data based on whether a fieldtype changes should be it's own unit of work with it's own considerations. 🤔

Yup, looks good. Maybe handling for Text and Translated can be included, seems common enough scenario

glennjacobs:
@alecritson I'm not sure if this is my install or the PR, but the below is what I'm seeing. I don't see values for handle or type and I can barely see the toggles.

image

alecritson:
> @alecritson I'm not sure if this is my install or the PR, but the below is what I'm seeing. I don't see values for handle or type and I can barely see the toggles.
> 
> image

This is what I see, can you check your install.
image

glennjacobs (approved):

PR #1695, opened 2024-04-15
Lock DOMPDF to `^2.0.7`

alecritson:
There is a current issue on DOMPDF 2.0.5 https://github.com/dompdf/dompdf/issues/3434

This PR Locks the version to one patch version behind the latest until it's fixed.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
     "require": {
         "awcodes/shout": "^2.0.2",
         "barryvdh/laravel-dompdf": "^2.0",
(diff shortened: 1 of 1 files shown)


alecritson:
Ok recommendation is to lock to `2.0.4` currently or upgrade, since we don't really have control over this too much lets lock.

glennjacobs:
@alecritson changing to `^2.0.7` should fix it I believe

glennjacobs (approved):

PR #1699, opened 2024-04-16
Trigger searchable when records are updated.

alecritson:
There are numerous places in the panel, like relationship managers, which don't trigger it's owner record's Observers and as a consequence  it's not always clear if a searchable record has been updated on the index.

Whilst the use of observers will still work this PR looks to also hook into various `after` triggers in the panel to manually call `searchable` on the affected model.


glennjacobs:
@alecritson I think we should be putting events into the admin panel and then listening to them for functionality like this.

glennjacobs (approved):

PR #1704, opened 2024-04-20
Fix - Prevent bulk delete action on resources with childs

### LINKED ISSUE #1559: Deleting a collection with child collections causes an error
We need to update the admin panel to disable deleting of collections that have child collections.

lguichard:


alecritson:
Thanks @lguichard the only issue I can potentially see with this is calling the relationship like `->attributes->count()` is going to lazy load them for each row which is going to result in quite a lot of heavy queries potentially.

Would it be more efficient to modify the query and load the count like `$query->withCount(['attributes'])` and then change the check to `->attributes_count` instead?

lguichard:
@alecritson Thanks for the feedback. It was so obvious that I forgot it 😵‍💫

alecritson:
> @alecritson Thanks for the feedback. It was so obvious that I forgot it 😵‍💫

Thanks @lguichard I've made one little tweak to the statement logic, hopefully that's alright and makes sense :)

glennjacobs:
@lguichard @alecritson I do agree we have an issue here that needs to be resolved, but the fundamental issue I can see here is that by turning off the bulk update checkbox, we'll not be able to do other bulk updates in the future. I think we need to look at a different approach and we also need to look at the delete buttons on the individual edit pages.

glennjacobs (changes requested):
See comments made.

alecritson:
Thanks for this @lguichard. I'm going to close this PR for now as after some discussion, we are going to take a wider look at deleting across Lunar and address some other concerns that have come up.

PR #1709, opened 2024-04-25
Fix - Cart bottleneck

lguichard:
The CartSessionManager must be cache the cart instance. 

Currently, the CartSession::current generate n+1 query.


alecritson:
Thanks @lguichard Do you think we could simplify this and remove some a duplicated `if` by doing

```php
$this->cart = $this->cart ?: Cart::with(
)->find($cartId);
```

lguichard:
@alecritson Oups ! In the future I will pay more attention to the style code =)

alecritson (approved):

PR #1718, opened 2024-04-30
Handle cart line update validation

alecritson:
Currently when validating the quantity when updating cart lines, we don't pass the purchasable so most of the validation actually gets skipped. This PR tries to handle if the purchasable is missing and act accordingly.


glennjacobs:
I feel like we should have a test for this, as it's the cart logic.

glennjacobs (approved):

PR #1727, opened 2024-05-07
Fixed hardcoded navigation group

ThorNissen:
Fixed hardcoded navigation group


alecritson:
@glennjacobs Do you think this should be looked at on a wider scope, I feel like #1760 is part of this?

glennjacobs:
> @glennjacobs Do you think this should be looked at on a wider scope, I feel like #1760 is part of this?

I would just get this one in, then handle the other issue with a separate PR.

alecritson (approved):

PR #1729, opened 2024-05-08
Fix observer namespace for shipping observer

alecritson:
Currently the observer registered in the shipping service provider has the wrong namespace.


glennjacobs:
@alecritson I guess that happened pre-Larastan.

Do we need to be fixing old data?

glennjacobs (approved):

PR #1385, opened 2023-12-19
Provide an option to automatically add BuyXGetY discount rewards

ryanmitchell:
3rd times a charm?

This PR provides an option to automatically add BuyXGetY discount rewards.

It adds a toggle to the discount create/edit screen to automatically add.

When selected we work out if any rewards are available that haven't yet been added to the cart and if we so we add them. Along the way we keep track of how many we added, so we can decrement/increment if cart values are changed, and if necessary remove the item from the cart when it no longer is eligible.

Devs can check for whether an item was added by discount by checking the meta->added_by_discount array values ... if the sum is the same as the cart line quantity then it was added by a discount and can be displayed differently (if required).

Replaces https://github.com/lunarphp/lunar/pull/881
Closes https://github.com/lunarphp/lunar/issues/876

PR #1733, opened 2024-05-09
Provide an option to automatically add BuyXGetY discount rewards

### LINKED ISSUE #876: Discount BuyXGetY automatically add
It would be great if there was the option to add Y items automatically to the cart when you have matching X items in your cart. This is really useful for say a free ebook alongside a paperback purchase.

I'd be happy to work on this if it was something you'd consider acceptable?

ryanmitchell:
This PR provides an option to automatically add BuyXGetY discount rewards.

It adds a toggle to the discount create/edit screen to automatically add.

When selected we work out if any rewards are available that haven't yet been added to the cart and if we so we add them. Along the way we keep track of how many we added, so we can decrement/increment if cart values are changed, and if necessary remove the item from the cart when it no longer is eligible.

Devs can check for whether an item was added by discount by checking the meta->added_by_discount array values ... if the sum is the same as the cart line quantity then it was added by a discount and can be displayed differently (if required).

Replaces https://github.com/lunarphp/lunar/pull/1385
Closes https://github.com/lunarphp/lunar/issues/876


glennjacobs:
@ryanmitchell the logic doesn't seem quite right for me. Please see below a Loom video demonstrating what I'm seeing.

https://www.loom.com/share/a03f218e2482432abb5e4f5b91346294?sid=2f6843a6-0b63-4f77-9201-3515cf06253c

ryanmitchell:
Ok so to clarify - you would expect it to remove any that have been added? thats a little more opinionated than I had been aiming for (which is fine) - I was leaving them in the cart, but just not marking them as having been added by the discount any more.

glennjacobs:
> Ok so to clarify - you would expect it to remove any that have been added? thats a little more opinionated than I had been aiming for (which is fine) - I was leaving them in the cart, but just not marking them as having been added by the discount any more.

I'm not really concerned about it removing (that's just what I had taken from your description).

>  and if necessary remove the item from the cart when it no longer is eligible

However, it shouldn't apply the additional product when the qty is 23.

glennjacobs:
@ryanmitchell from looking at your code, doesn't it try to automatically remove the item?

glennjacobs:
@ryanmitchell I think the issue with the 23 is that it's probably loading the cart first, which sees it's got 24, so applies the discount and then we update the qty to 23, making it seem like 23 caused the item to be added.

If it auto-removed the item, then I guess this wouldn't happen.

ryanmitchell:
> @ryanmitchell from looking at your code, doesn't it try to automatically remove the item?

Yeah it does... it's been too long since I looked at this properly.

> @ryanmitchell I think the issue with the 23 is that it's probably loading the cart first, which sees it's got 24, so applies the discount and then we update the qty to 23, making it seem like 23 caused the item to be added.

Yep thats whats happening. I'll try and get another look over the next day or two.

glennjacobs:
@ryanmitchell I'll try take a look now to see if I can get my head around it

glennjacobs:
@ryanmitchell having looked at the code further, it's obvious the "removing" part of the code won't be hit as the discount bails on line 69 if it doesn't qualify.

So I propose we don't worry about removing the item, only automatically adding it.

Do you agree? If so, I can look to finish up the PR.

ryanmitchell:
Yep that works for me - keeps it simpler

glennjacobs:
The outstanding issue on this is the fact that the discounts apply before a qty change is considered, meaning the automatically applied product can come back if the previous qty was above the qualifying amount.

The CartSession Manager calculates the cart on load which causes this. Changing this could have been quite achievable if it wasn't for the estimate shipping logic. @alecritson I'll probably need to discuss this one with you.

glennjacobs (approved):

PR #1736, opened 2024-05-10
Add attribute data to `ProductOption` and `ProductOptionValue`

adevade:
This PR adds attribute data to `ProductOption` and `ProductOptionValue`.

It will be useful for storing meta data related to a **specific option** (_such as a description, or a link to a sizing chart_) and to a **specific option value** (_such as a hex value for a color, or a short description_).

It hasn't been tested yet, and no UI has been added, since I didn't know if this would get accepted. It has previously been requested/discussed here: https://github.com/lunarphp/lunar/discussions/1296

If you agree i can clean it up, test it and add a UI to the panel. Or feel free to push changes as necessary. 👍

---

### Related
I also agree with the linked discussion that it would be useful to have a handle on `ProductOptionValue` for usage in query strings for example. What are your thoughts on this?


Saifallak:
`such as a hex value for a color` great addition, but anyway in the admin panel, can we choose field type of attributes?

adevade:
Marking this as ready for review, even though there has to be some UI to be able to setup the extra fields in the admin panel.

glennjacobs:
@adevade I do think this would be a welcome addition. As you say, it needs the admin UI applying to this PR first before it could be considered for review.

The handle concept could be trickier to implement, but doable. It could be auto-generated based upon the value initially, with logic to avoid duplicates, e.g. "red" becomes "red-2" or something. And then under settings we could allow the handles to be manually edited.

glennjacobs:
I had a long think about this and my preferred route is to go with meta, rather than attribute data. So closing in favor of #2346 

This will then allow add-on packages to manage the meta in their preferred way.

PR #1737, opened 2024-05-10
[Spring cleaning] Refactor migrations

adevade:
**Note:** This PR is probably easiest reviewed commit by commit instead of all changes at once.

Just some spring cleaning and standardization of the migration classes. I converted all to anonymous classes, like Laravel did during the 10.x releases.

Removed PHPDoc blocks, since they were not always there and less clutter.

All tests still passing locally.

Feel free to request changes or revert single commits if you don't like some of the changes.


alecritson:
Thanks @adevade 

@glennjacobs This looks like something we should get in (if wanted) before too much else gets merged to give us a good baseline.

glennjacobs (approved):

glennjacobs:
> @glennjacobs This looks like something we should get in (if wanted) before too much else gets merged to give us a good baseline.

Maybe run this on an existing site to ensure nothing weird happens before merging?

adevade:
@glennjacobs Yeah, I have not tested for that. I don't believe there should be any problems though.

alecritson:
Just ran this on the starter kit and didn't come against any issues so if everyone is happy @glennjacobs I guess it can go in.

PR #1742, opened 2024-05-13
Consider authenticated user when fetching cart from session

alecritson:
Currently the cart session manager doesn't try to fetch a users most recent and active cart if they don't already have a cart id in the session.

This PR adds a check to fallback on a users active cart before trying to create anything new.

### DIFF
--- a/packages/core/src/Base/Traits/LunarUser.php
+++ b/packages/core/src/Base/Traits/LunarUser.php
@@ -4,6 +4,7 @@ namespace Lunar\Base\Traits;
 
 use Illuminate\Database\Eloquent\Relations\BelongsToMany;
 use Illuminate\Database\Eloquent\Relations\HasMany;
+use Lunar\Models\Cart;
 use Lunar\Models\Customer;
 use Lunar\Models\Order;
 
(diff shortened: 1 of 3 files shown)


glennjacobs (changes requested):
This should have a test.

glennjacobs (approved):

wychoong:
this breaks the `CartSession::forget()` functionality where after forgetting the cart, it fetches by user

my testing
image

my logs
```
[2024-05-23 11:34:35] local.INFO: delete  << calling CartSession::forget()
[2024-05-23 11:34:35] local.INFO: fetch from session  
[2024-05-23 11:34:35] local.INFO: fetch from user  
```

glennjacobs:
> this breaks the `CartSession::forget()` functionality where after forgetting the cart, it fetches by user
> 
> my testing image
> 
> my logs
> 
> ```
> [2024-05-23 11:34:35] local.INFO: delete  << calling CartSession::forget()
> [2024-05-23 11:34:35] local.INFO: fetch from session  
> [2024-05-23 11:34:35] local.INFO: fetch from user  
> ```

I get the problem. But could you explain your use case for `CartSession::forget()`?

wychoong:
My use case is to delete a cart belongs to the user. 
And also sort of one click to empty the cart. 

Although as workaround we can loop to remove the cart lines, there are more things to clear, eg: addresses, shipping option, coupon, meta, cache?, etc. 

A simple forget and delete will be more performant

alecritson:
@wychoong Myself and Glenn discussed this and propose the following solution: https://github.com/lunarphp/lunar/pull/1775

PR #1743, opened 2024-05-13
Merge 0.8

glennjacobs:
Merges changes from the `0.8` branch into `1.x`.


glennjacobs:
@alecritson do you think we should have the `.github` folders on each package?

packages/admin/src/Models/Staff.php
```diff
@@ -117,6 +124,16 @@ public function getFullNameAttribute(): string
...
     }
 
+    /**
+     * Get staff member's Gravatar URLs.
+     */
+    public function getGravatarAttribute(): string
+    {
+        $hash = md5(strtolower(trim($this->attributes['email'])));
+
+        return "https://www.gravatar.com/avatar/$hash?d=mp";
```
alecritson: Through Lunar we usually do `"{$hash}"` when using string interpolation

alecritson (changes requested):

packages/admin/src/Models/Staff.php
glennjacobs: This was just a merge, but i can update it.

alecritson:
> @alecritson do you think we should have the `.github` folders on each package?

I assume we need them because they have specific workflows which auto close PRs?

glennjacobs:
> > @alecritson do you think we should have the `.github` folders on each package?
> 
> I assume we need them because they have specific workflows which auto close PRs?

Yeah, they're to auto-close PRs on the read-only repos. It could be argued that all the packages should have them.

alecritson (dismissed):

tests/core/Unit/Search/ProductOptionIndexerTest.php
```diff
@@ -11,6 +11,6 @@
 
     $data = app(ProductOptionIndexer::class)->toSearchableArray($productOption);
 
-    expect($data['name_en'])->toEqual($productOption->name['en']);
-    expect($data['label_en'])->toEqual($productOption->label['en']);
+    $this->assertEquals($productOption->name->en, $data['name_en']);
```
alecritson: Just realised after I approved that this has been reverted to PHPUnit syntax.

alecritson (approved):

PR #1745, opened 2024-05-14
Allow order creation without a shipping address when a collection shipping option has been selected

tvlokven:
Currently, a shipping address is always required if a Cart is "shippable", i.e. when it contains physical products. However, when a customer chooses to pick up an order, rather than having it delivered, a shipping address should not be required.


alecritson:
Thanks @tvlokven Are you able to provide a test to prove this is working as intended? :)

tvlokven:
> Thanks @tvlokven Are you able to provide a test to prove this is working as intended? :)

@alecritson Yes of course, I added tests for the different scenarios.

glennjacobs (approved):

PR #1752, opened 2024-05-16
Fix: Unknown Match in CustomerGroupRelationManager

tuvshinjargal:
Not sure this is a correct way to fix or no. But, happened an exception/issue during installation and testing when attach a customer group to product in availability section:

Here is the exception;

Screenshot 2024-05-15 at 9 26 02 PM

After this, it will look like this.

Screenshot 2024-05-15 at 9 42 55 PM


alecritson:
Thanks @tuvshinjargal Do you know what the state was if it wasn't `0` or `1` ?

tuvshinjargal:
@alecritson  This was `0` (`false`) on the table

Screenshot 2024-05-16 at 5 27 41 AM

alecritson:
Thanks @tuvshinjargal I don't think this solution is correct, I'm going to close but if possible could you raise as an issue instead with more details on how to reproduce so we can approach this a different way :)

PR #1757, opened 2024-05-17
Allow Stripe to be faked at a facade level

alecritson:
Currently Stripe is faked within it's own tests, which is great but doesn't allow other apps or packages that use it to do the same in their own tests.

This PR moves the fake method into the Facade class (much like how Laravel do) so that it is it available for other developers to utilise.

Usage

```
\Lunar\Stripe\Facades\Stripe::fake();
```

---

This PR will now also rename `StripeFacade` to `Stripe`


packages/stripe/src/Facades/Stripe.php
```diff
@@ -0,0 +1,32 @@
...
+ * @method static buildIntent(int $value, string $currencyCode, \Lunar\Models\CartAddress $shipping): \Stripe\PaymentIntent
+ */
+class Stripe extends Facade
+{
+    /**
+     * {@inheritdoc}
+     */
+    protected static function getFacadeAccessor(): string
+    {
+        return 'gc:stripe';
```
glennjacobs: `gc:` I'm guess this stands for "GetCandy", update to "lunar"?

glennjacobs (approved):

PR #1767, opened 2024-05-21
Refactor Lunar transactions to use payment checks DTO

alecritson:
Currently there isn't a standard way for Lunar to show what payment checks were successful, such as Address checks or Security code checks. Although it does make somewhat of an attempt, due to all payment providers being different it is leading to a lot of false positives.

This PR looks to introduce `PaymentChecks` and `PaymentCheck` Data Transfer Objects which allow each payment driver to return what payment checks have taken place, what the message was and whether it was successful.

Usage is via the `Transaction` model like so:

```php
foreach($transaction->paymentChecks() as $check) {
    $check->successful; // true/false
    $check->message; // Matched
    $check->label; // PostalCode
}
```

For developers who have implemented their own payment type they will need to implement the `getPaymentChecks` method and interpret the meta data on the transaction accordingly. If a payment driver doesn't store this information they don't need to do anything since a base method is implements in the `AbstractPaymentDriver`.

An implementation may look like the following (snippet from Stripe payment driver):

```php
use Lunar\Base\DataTransferObjects\PaymentCheck;
use Lunar\Base\DataTransferObjects\PaymentChecks;

// ...

public function getPaymentChecks(Transaction $transaction): PaymentChecks
{
    $meta = $transaction->meta;

    $checks = new PaymentChecks;

    if (isset($meta['address_line1_check'])) {
        $checks->addCheck(
            new PaymentCheck(
                successful: $meta['address_line1_check'] == 'pass',
                label: 'Address Line 1',
                message: $meta['address_line1_check'],
            )
        );
    }

    // ...

   return $checks;
}

```


glennjacobs:
@alecritson I think we need an entry in the upgrade docs, as I presume this will be breaking if someone has made their own payment driver?

glennjacobs:
> @alecritson I think we need an entry in the upgrade docs, as I presume this will be breaking if someone has made their own payment driver?

Scrap that, I can see it's on the abstract class. So not needed.

glennjacobs:
An old test order, but this has a cross but is matched on the security code.

image

alecritson:
> An old test order, but this has a cross but is matched on the security code.
> 
> image

This should be patched

glennjacobs (approved):

PR #1769, opened 2024-05-21
Add textarea fieldtype

ryanmitchell:
This PR adds a textarea fieldtype which maps to a Filament textarea. 
This allows you have multiline text attributes without requiring them to be rich text.


alecritson:
Thanks @ryanmitchell looking good, just one small issue. If you have the textarea set to autoheight and click save, it reverts to a scrollable field.

Before save

image

After save

image


After Refresh


image

ryanmitchell:
Interesting - any thoughts on why thats happening? I copied the synth approach from the other field types.

alecritson:
Not too sure actually, but I added a Textarea component to the form manually and tried and it did the same, maybe it's a problem on filament.

alecritson:
Looks like it was reported last year and apparently fixed 🧐 https://github.com/filamentphp/filament/issues/7620

glennjacobs:
It feels a little odd to have a textarea field, but then have a text field which can be rich text.

Should we consider moving the rich text option to this textarea field?

glennjacobs:
Hi @ryanmitchell 

I've considered this, and I think it's wrong. We should enhance the current Text and TranslatedText lunar field types to allow for "multi-line".

So I'm going to close this, but please let me know if you'd like to work on the "multi-line" option.

PR #1771, opened 2024-05-22
Update transaction.blade.php

eugenefvdm:
Fix date format

The date showed the month in the place of he minutes and seconds were missing.

### DIFF
--- a/packages/admin/resources/views/infolists/components/transaction.blade.php
+++ b/packages/admin/resources/views/infolists/components/transaction.blade.php
@@ -76,7 +76,7 @@
                     />
                 </div>
                 
-                <span>{{ $transaction->created_at->format('jS F Y h:ma') }}</span>
+                <span>{{ $transaction->created_at->format('jS F Y h:i:s') }}</span>
             </div>
 
             @if($threeD = $transaction->meta['threedSecure'] ?? null)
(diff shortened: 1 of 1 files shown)


alecritson:
Thanks @eugenefvdm although I'm not sure we need to show seconds I think something like `2nd January 2024 03:30pm" should be sufficient. @glennjacobs ?

glennjacobs:
> Thanks @eugenefvdm although I'm not sure we need to show seconds I think something like `2nd January 2024 03:30pm" should be sufficient. @glennjacobs ?

Seconds is unlikely to be useful, so yeah, let's remove.

eugenefvdm:
I added minutes because if you're doing many transactions per minute you can use the seconds for troubleshooting.

But of course, I can change it to the below as per @alecritson 

```php
$today = now()->format("jS F Y h:ia");
echo $today;
22nd May 2024 01:42pm
```

alecritson (approved):

alecritson:
> I added minutes because if you're doing many transactions per minute you can use the seconds for troubleshooting.
> 
> But of course, I can change it to the below as per @alecritson
> 
> ```
> $today = now()->format("jS F Y h:ia");
> echo $today;
> 22nd May 2024 01:42pm
> ```

Sure, although I don't think this is a concern for Lunar right now as we currently don't have a way to view all transactions in a single place. Might be something for the future

PR #1772, opened 2024-05-22
Fix null value errors in attribute casting

ryanmitchell:
When you add a new attribute and edit a item but dont set a value for it, you receive an error due to the value being null when the cast tries to work out its type.

This PR works around that issue by ignoring null values.


alecritson:
@ryanmitchell Are you able to either create an issue with steps or provide steps on how to reproduce this the issue this is fixing as currently I'm not getting any errors

ryanmitchell:
So it happens for me when I create a new product variant attribute then edit an existing variant but dont blur that field (so leave it blank and never touch it).

alecritson:
I just tried these steps and everything seemed ok 🤔 here's a quick video of what I did.

https://www.loom.com/share/25293d74d5ef47ce9b0fe5080e816114

ryanmitchell:
Here's my error trace: https://flareapp.io/share/17WvaBnP
Clean install - with a lot of attribute fields and groups.

alecritson:
> Here's my error trace: https://flareapp.io/share/17WvaBnP Clean install - with a lot of attribute fields and groups.

What does your attribute data look like for the offending product variant?

ryanmitchell:
```
{"publication-status": {"value": "In Print", "field_type": "Lunar\\FieldTypes\\Dropdown"}, "onix-country-of-manufacture": {"value": "asdsdasdas", "field_type": "Lunar\\FieldTypes\\Text"}}```

Tried nulling it but no difference there.

alecritson:
> ```
> {"publication-status": {"value": "In Print", "field_type": "Lunar\\FieldTypes\\Dropdown"}, "onix-country-of-manufacture": {"value": "asdsdasdas", "field_type": "Lunar\\FieldTypes\\Text"}}```
> 
> Tried nulling it but no difference there.
> ```

Strange mine looks basically the same yet unable to reproduce this one, things seem to work as expected.

ryanmitchell:
Not sure why. I'm on a clean install and it happens consistently on every product.

alecritson:
> Not sure why. I'm on a clean install and it happens consistently on every product.

I just tried on a new install of the starter kit and I still wasn't able to reproduce this. Are you able to `dd` the `$value` to see what's being sent up whilst it's being saved? Bit of a head scratcher for sure. 

@glennjacobs Might need to ask if you can reproduce this

ryanmitchell:
Sure, so its:

CleanShot 2024-05-23 at 08 07 59@2x

alecritson:
> Sure, so its:
> 
> CleanShot 2024-05-23 at 08 07 59@2x

Very strange, how are your attributes set up? Specifically the ones which are showing as null?

ryanmitchell:
Nothing abnormal - its a pretty vanilla build (just starting it):

CleanShot 2024-05-23 at 08 29 49@2x

alecritson:
> Nothing abnormal - its a pretty vanilla build (just starting it):
> 
> CleanShot 2024-05-23 at 08 29 49@2x

Very strange, so if I save a variant which has a new attribute and `dd` on the `handleRecordUpdate` in `BaseEditRecord` then my attribute data looks like this

image

if I so the same on a product, it looks like this:

image


Everything seems to have an appropriate field type set with a value 🤔 you haven't got any custom code that uses hooks or anything? Seems like the issue is the panel isn't hydrating the form attribute data for some reason

ryanmitchell:
Nope it's a pretty clean install at this point. Oh well, must be something on our side, but I've no idea what.

That said, it's a pretty simple PR - if the value is null you wouldn't want it being saved anyway?

alecritson:
Weirdly this just came up on one of our sites 😅 I've assigned @glennjacobs to see if he is able to reproduce. My only concern would be that the attributes should have the correct mapping on save and I'd be worried this would just be masking another issue altogether.

ryanmitchell:
Oh good that you can see it now, and that I'm not losing my mind.

lguichard:
I have the bug in the Brand edit page but i can't reproduce the steps

wychoong:
I’m having same issue and replicable with these steps https://github.com/lunarphp/lunar/pull/1772#issuecomment-2124263199

My issue is with Toggle field and so far I traced to this line
https://github.com/lunarphp/lunar/blob/fe6108a4bfb4b0fc35a35ee190c2e8e8afcafc61/packages/core/src/FieldTypes/Toggle.php#L43
By making sure it’s returning 0 or 1 only

It also makes me think is there a change in livewire that the issue only happens now 😂

alecritson:
@ryanmitchell Finally managed to consistently reproduce this, for me it was to add a new attribute to a different product type and then try switching that product to that type, which caused the error.

I switched to the branch on @wychoong 's PR and the error went away

ryanmitchell:
Great!

PR #1775, opened 2024-05-23
Add soft deletes to carts

alecritson:
This PR adds soft deletes to carts. The idea behind this is when you want to `forget` a cart, it is likely you want to remove the cart from a users current and future session. 

The problem is in order to fully achieve this you would need to fully remove the cart from the database, which means there is a loss of important data as a store owner.

There are also situations you may want to forget the cart session, such as logging out, but you may want to retain that cart for when the user logs back in and can pick up from where they left off.

This PR looks to solve these issues by adding soft deletes to carts. The default behaviour is the cart will be soft deleted when you call `CartSession::forget`, but you can opt to keep the cart "active" by passing the `delete: false` parameter.

### Existing Storefronts who want to soft delete carts
No further action is needed, call `CartSession::forget()` and the carts should be deleted.

### Existing Storefronts who **do not** want to soft delete carts

```
// Old
CartSession::forget();

// New
CartSession::forget(delete: false);
```


wychoong:
I have tested with this, it does solve my problem.

I would suggest to change the method to `::forget()` and `::delete()` instead of the parameter 
make a breaking change now is better for a clearer api

glennjacobs (dismissed):

glennjacobs:
> I have tested with this, it does solve my problem.
> 
> I would suggest to change the method to `::forget()` and `::delete()` instead of the parameter make a breaking change now is better for a clearer api

@alecritson and I considered your suggestion, but we both prefer the originally proposed naming.

wychoong:
The cart prune command should be updated as well?

glennjacobs:
> The cart prune command should be updated as well?

Good point, we might want to add `->withTrashed()`.

alecritson:
I've adjusted the prune command to use withTrashed, hopefully that'll do it.

glennjacobs:
@alecritson although we have it in the upgrade guide, we should also probably add to the main docs.

PR #1776, opened 2024-05-24
Fix/empty attributes

wychoong:
Fix error `get_class(): Argument #1 ($object) must be of type object, null given`

this happens when added a new attribute to a resource.
the fix https://github.com/lunarphp/lunar/pull/1776/commits/7b19a0e4c213c065f6520209074c4375d06bcfc0 is to mutate the state during field dehydration to return new field type instance when state is null
`formatStateUsing` doesnt set the instance during field hydration as it only apply when the state is not null

other commits are some inconsistency that I found while troubleshooting, it should still works without those commits

### DIFF
--- a/packages/admin/src/Support/FieldTypes/Toggle.php
+++ b/packages/admin/src/Support/FieldTypes/Toggle.php
@@ -18,7 +18,6 @@ class Toggle extends BaseFieldType
                 $attribute->translate('description')
             )
             ->default(false)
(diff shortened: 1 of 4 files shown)


wychoong:
related to https://github.com/lunarphp/lunar/pull/1772

i think ultimately its down to whether the attribute_data field should keep the field types when the value is empty

glennjacobs (approved):

PR #1778, opened 2024-05-25
fix language factory possible languageCode collision 

wychoong:
when running test, when have enough iteration (eg: ->repeat(50)) there is a possibility of hitting UniqueConstraintViolationException

changing from `->unique()->languageCode` to `->slug()` fixes this as its not a fixed set of language codes


alecritson:
Thanks @wychoong 

Not sure this is entirely needed, In your loop could you not just override the `code` yourself?

wychoong:
> Thanks @wychoong
> 
> Not sure this is entirely needed, In your loop could you not just override the `code` yourself?

The problem is when create a product with the factory, if no language id is provided to url factory, it will trigger the language factory. So when create multiple products, it might hit the constraint. 

And in my test suite setup I configured a seeder to create language for convenience with getDefault(). But now chances of hitting the constraint is even higher. 

However to always pass a language id when creating products is unnecessary when language is not a concern in the test case. 

```php
Product::factory()
        ->hasUrls(1, [
            'default' => true,
            // 'language_id' => Language::getDefault(),
        ])
```

alecritson (approved):

PR #1779, opened 2024-05-25
Add unique index to collectionGroup handle

wychoong:
currently it does not have unique check on db level and only the admin form have unique validation.

was writing test suite and not aware of created duplicate group handles and page unable to fetch the correct group attached to products...


glennjacobs:
@alecritson I guess this is potentially a breaking change, so should be mentioned in the upgrade guide.

wychoong:
added changelog

PR #1783, opened 2024-05-26
cleanup ProductOption accessor/mutator and fix fields casting

wychoong:
with the accessor/mutator defined for the `name` and `label` fields, the cast is never used
this PR removed the redundant accessor/mutator and changed to `AsArrayObject` cast instead of `As Collection`

using `AsArrayObject` allows following usage
image

as compared to previous accessor/mutator json_encode
image

or `AsCollection`
image


alecritson (dismissed):

glennjacobs:
@alecritson @wychoong is this still needed? I can see it's draft, but at the same time Alec approved it.

wychoong:
Not sure why it’s changed to draft

wychoong:
@glennjacobs I took a fresh look on it, this should be the proper way. updated my findings in the first comment.

It also makes me think should all the usage of `AsCollection` cast in the models to be reviewed and changed to `AsArrayObject`

PR #1791, opened 2024-05-28
Add stock validation helpers and classes

alecritson:


docs/core/upgrading.md
```diff
@@ -18,6 +18,27 @@ php artisan migrate
 
 Lunar currently provides bug fixes and security updates for only the latest minor release, e.g. `0.7`.
 
+## [Unreleased]
+
+### Medium Impact
+
+If you are using your own classes that implement the `Purchasable` interface, you will need to add the following additional method:
```
glennjacobs: `methods`

packages/core/src/Base/Purchasable.php
```diff
@@ -72,4 +72,8 @@ public function isShippable();
      * @return string
      */
     public function getThumbnail();
+
+    public function canBeFulfilledAtQuantity(int $quantity): bool;
+
+    public function getTrueStockValue(): int;
```
glennjacobs: I'm not sure on the name "True Stock", I'll give it some thought.

glennjacobs: How about "Total Inventory"?

packages/core/src/Models/Cart.php
```diff
@@ -612,6 +614,19 @@ public function canCreateOrder(): bool
         return $passes;
     }
 
+    /**
+     * @throws ValidationException
+     */
+    public function validateStock(): void
+    {
+        $this->lines->each(
+            fn ($line) => app(CartLineStock::class)->using(
```
glennjacobs, line 623: Would we not need to consider whether the item is a physical item or not? E.g. what if someone is selling digital goods?

alecritson: Since we have the `isShippable()` method on the the purchasable interface, I added a check into the validator. I assume this should cater enough for people that use it until we support digital goods fully.

glennjacobs: I don't think we should be mixing shipping with stock, it's not the same thing. I think the answer is we need to look at digital goods later. The system allows for items to "always" be purchasable which is enough for now.

alecritson: Yeah that's fair. I've reverted.

docs/core/upgrading.md
```diff
@@ -33,6 +33,27 @@ To allow for recalculation we have now introduced `$cart->recalculate()` to forc
...
+public function getTotalInventory(): int;
+```
+
+If you are checking the `ProductVariant` `purchasable` attribute in your code, you should update the following check:
+
+```php
+// Old
+$variant->purchasable == 'backorder';
+// New
+$variant->purchasable == 'in_stock_on_backorder';
```
glennjacobs: Needs updating

packages/admin/resources/lang/en/productvariant.php
```diff
@@ -56,7 +56,7 @@
             'options' => [
                 'always' => 'Always',
                 'in_stock' => 'In Stock',
-                'backorder' => 'Backorder Only',
+                'in_stock_or_on_backorder' => 'In Stock or on Backorder',
```
glennjacobs: Should be "In Stock or On Backorder"

glennjacobs (approved):

PR #1793, opened 2024-05-28
Refactor Staff model

adevade:
**Note:** This PR is easiest to review commit by commit. Feel free to pick and choose what you like and revert the ones you don't.

The reason for removing the Docblocks is because Filament barely uses any. With clear property/method names, type hints and return types they often feel superfluous. 

I _think_ that we can remove the custom route binding after [this PR was merged in Livewire](https://github.com/livewire/livewire/pull/7913)(?) 

I usually use my accessors as the rest of my attributes (snake cased), but don't know if you have a strong preference? Both still works. Especially in `$appends` I think it should be snake cased, since otherwise it would be mixing case with the other fields from the DB.

Marking as draft for now, to let the tests run and get some feedback.


alecritson:
@adevade Sorry this one took a while, looks pretty okay to me. I've added a small test (which didn't exist before) lmk your thoughts and we can get this merged.

adevade:
@alecritson Nice work with the test! 👍

One thing I noticed is that `firstname` and `lastname` is defined without underscores. Should `fullname` be that as well, for consistency? Easier than migrating both first and last name columns I guess?

alecritson:
> @alecritson Nice work with the test! 👍
> 
> One thing I noticed is that `firstname` and `lastname` is defined without underscores. Should `fullname` be that as well, for consistency? Easier than migrating both first and last name columns I guess?

I think if anything we'd be better to migrate `firstname` to `first_name` it's something that's bothered us internally since elsewhere we use underscores. Not sure if it's something we need to worry about right now though @glennjacobs ?

glennjacobs:
> > @alecritson Nice work with the test! 👍
> > One thing I noticed is that `firstname` and `lastname` is defined without underscores. Should `fullname` be that as well, for consistency? Easier than migrating both first and last name columns I guess?
> 
> I think if anything we'd be better to migrate `firstname` to `first_name` it's something that's bothered us internally since elsewhere we use underscores. Not sure if it's something we need to worry about right now though @glennjacobs ?

I guess if we don't do it now, then it may never get another look. So maybe we get it done in this PR?

adevade:
I've now updated the code base to use `first_name` and `last_name`.

Since users might be referencing `$staff->firstname` in their own code, I've added fallback accessor and mutators for both `$firstname` and `$lastname` on the Staff model.

I did not update [the CLI argument keys on `MakeLunarAdminCommand`](https://github.com/lunarphp/lunar/blob/736d14c86f96be08d90a49643bac97f77aa2231f/packages/admin/src/Console/Commands/MakeLunarAdminCommand.php#L19-L20), since it feels better for them to not have underscores.

I have not yet touched any documentation.

alecritson:
Hi @adevade I realise this is an old PR, apologies.

Is this something that still needs active work? Looks like there is a checklist of items in the readme?

adevade:
@alecritson Should still be good to merge 👍 The tests are passing. Haven't looked at it in a good while though...

alecritson (approved):

PR #1795, opened 2024-05-30
Add attribute check on name attribute on base resource

alecritson:
Currently if the name attribute is just `Text` instead of `TranslatedText` this will fail as the panel is rendering a translated text input, causing the name value to be an array of locales.

This PR adds an additional check on the method that fetches the name input so the form data is correct on save.


glennjacobs:
Considering we want to move to a dedicated `name` field for models like Product, I don't think we want to be adding further "code smell". So closing in favour of moving to the dedicated name field ASAP.

glennjacobs:
@alecritson as we've decided not to make the `name` change in v1, I've reopened this PR.

glennjacobs:
@alecritson will we also need this for Collection editing etc, or does it only affect Products?

alecritson:
> @alecritson will we also need this for Collection editing etc, or does it only affect Products?

It's already handled on the collection page.

glennjacobs (approved):

PR #1796, opened 2024-05-30
Fix - Global search order

lguichard:
This PR fix null attribute contact_email and return fullName of the customer.


packages/admin/src/Filament/Resources/OrderResource.php
```diff
@@ -237,10 +237,10 @@ public static function getGlobalSearchResultDetails(Model $record): array
         $details = [
             __('lunarpanel::order.table.status.label') => $record->getStatusLabelAttribute(),
             __('lunarpanel::order.table.total.label') => $record->total?->formatted,
-            __('lunarpanel::order.table.customer.label') => $record->shippingAddress->fullName,
+            __('lunarpanel::order.table.customer.label') => $record->customer->fullName,
```
alecritson: We won't always have a customer record associated, should this be the `billingAddress`?

packages/admin/src/Filament/Resources/OrderResource.php
```diff
@@ -237,10 +237,10 @@ public static function getGlobalSearchResultDetails(Model $record): array
         $details = [
             __('lunarpanel::order.table.status.label') => $record->getStatusLabelAttribute(),
             __('lunarpanel::order.table.total.label') => $record->total?->formatted,
-            __('lunarpanel::order.table.customer.label') => $record->shippingAddress->fullName,
+            __('lunarpanel::order.table.customer.label') => $record->shippingAddress?->fullName,
```
alecritson: Certainly still more inclined to use the `billingAddress` here as we will usually have it and I think it better reflects the customer. Since orders could be sent to an address not associated to the customer such as drop shipping. So I don't think the Shipping address is always accurate in that sense.

glennjacobs (approved):

PR #1788, opened 2024-05-27
TextInput FormComponent

sugandhgandhi:
replace "TextInput FormComponent" with  "FileUpload FormComponent" in File FieldTypes

PR #1798, opened 2024-05-31
Fix/attribute validation

### LINKED ISSUE #1712: Attribute validation rules
* Lunar version: `^1.0.0-alpha`
* Laravel Version: `^11.0`
* PHP Version: `^8.2`
* Database Driver & Version: Using Laravel Sail image `mysql/mysql-server:8.0`

### Expected Behaviour:

In `0.8` it was possible to add a validation rule to an attribute.
The same should be possbile in `1.0.0-alpha`

### Actual Behaviour:

Right now it doesn't use the rule when using the attribute.

### Steps To Reproduce:

* A fresh Lunar `1.0.0-alpha` installation.
* Create an attribute group eg. "SEO".
* Create an attribute on the group eg. "Meta description" with the rule `max:50`
* Use the attribute on a product and save the product. (Should validate using `max:50` but it doesn't)

Example of `1.0.0-alpha` rule setup

1 0-rules

Example of `0.8` rule setup

0 8-rules

Example of `0.8` validation

0 8-validation

wychoong:
fixes #1712


alecritson:
@wychoong Looks like there are a number of failing tests, will set to draft for the time being and when it's done mark it as ready and I'll take a look :)

wychoong:
i added in fixes from #1788 as well

alecritson:
Thanks @wychoong 

This seems to work well, my only thinking now is that the validation rules input doesn't feel very intuitive, it's very easy to type something in wrong and will break the edit screens see screenshots:

image

image

Do we need validation on that field or do something like Statamic where it will suggest rules and add them like tags. 🤔 

@glennjacobs What do you think?

wychoong:
I think a validation on the rules can be done, but the rule builder/tags could be too much of effort

wychoong:
https://github.com/lunarphp/lunar/assets/67364036/907b9c48-785a-4e12-a454-c3d553ff9dd5




```php
->debounce()
                    ->rules([
                        fn (): Closure => function (string $attribute, $value, Closure $fail) {
                            if (blank($value)) {
                                return;
                            }

                            $fails = [];
                            foreach (explode('|', $value) as $rule) {
                                try {
                                    Validator::make(['test' => 'a'], ['test' => $rule])->validate();
                                } catch (\BadMethodCallException $e) {
                                    $fails[] = $rule;
                                } catch (Throwable $e) {
                                    //
                                } catch (Exception $e) {
                                    //
                                }
                            }

                            if (filled($fails)) {
                                $fail('Invalid rules: ' . implode(', ', $fails));
                            }
                        },
                    ])
                    ->afterStateUpdated(function ($livewire, Forms\Components\TextInput $component) {
                        $livewire->validateOnly($component->getStatePath());
                    })

glennjacobs (approved):

PR #1800, opened 2024-05-31
Fix - Add support key/value in order additional information 

lguichard:
This PR fix error if meta data of order contain an array.

Capture d’écran 2024-05-31 à 22 37 31


alecritson:
Thanks @lguichard I think the logic can be simplified a bit by first checking for the array instance and returning `KeyValueEntry` and then have the existing logic underneath, this will remove the `else` statement and also make it look a bit tidier :)

lguichard:
That's it

alecritson (approved):

alecritson (approved):

PR #1803, opened 2024-06-03
Fix - Upload media to external disk

lguichard:
This PR fix error Livewire temporaly file doesn't exist. 

If the dev uses several front instances, we need to use S3 or another temporary directory to upload temporary files on the external disk.

Currently, Livewire generate an exception with temporary file in S3.

Filament SpatieMediaLibraryFileUpload uses the same method :  
https://github.com/filamentphp/filament/blob/85594f3a37b37219d7b7f76a3f7a1a85fdb9cfad/packages/spatie-laravel-media-library-plugin/src/Forms/Components/SpatieMediaLibraryFileUpload.php#L143

### DIFF
--- a/packages/admin/src/Support/RelationManagers/MediaRelationManager.php
+++ b/packages/admin/src/Support/RelationManagers/MediaRelationManager.php
@@ -40,6 +40,7 @@ class MediaRelationManager extends BaseRelationManager
                     ->hiddenOn('edit')
                     ->storeFiles(false)
                     ->imageEditor()
+                    ->required()
                     ->imageEditorAspectRatios([
                         null,
                         '16:9',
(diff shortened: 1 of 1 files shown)


alecritson:
Thanks @lguichard What's a good way to test this out?

lguichard:
@alecritson You try to configure ngnix or traefik as reverse proxy with load balancer and multiple instances php-fpm. and configure LIVEWIRE_TEMPORARY_DISK=s3 That's little complicated for a test. 

The exception is 
Capture d’écran 2024-06-04 à 12 34 21

alecritson:
Thanks @lguichard I see what you mean. I had a little play around locally and made a couple of (small) tweaks.

I noticed media wasn't required on the form, this should now be sorted.

I also saw that using `addMediaFromString` will store the filename as `text.txt` (see below)

image

So I have added the following 

```php
->usingFileName(
    $data['media']->getClientOriginalName()
)
```

Which should allow us to keep the original name of the uploaded file. If you're happy, I'm happy and we can get this in.

lguichard:
@alecritson you are happy, i'm happy =)

alecritson (approved):

sulfo:
Hey @lguichard.
Did you run into issues where `$data['media']->get()` is null?

PR #1804, opened 2024-06-04
Fix - Price non-numeric value encountered

lguichard:
This PR fix error occurs when the use try to save decimal price


alecritson:
Thanks @lguichard We'll probably need to make sure it's covered on all places we require a price.

glennjacobs (approved):

PR #1805, opened 2024-06-06
Cart performance improvements

glennjacobs:
Previously calling `->calculate()` would force the cart pipelines to always run. This update will only re-run the cart pipelines if the cart hasn't been calculated previously, or we're forcing it to.

### DIFF
--- a/packages/core/src/Managers/CartSessionManager.php
+++ b/packages/core/src/Managers/CartSessionManager.php
@@ -18,9 +18,9 @@ class CartSessionManager implements CartSessionInterface
     public function __construct(
         protected SessionManager $sessionManager,
         protected AuthManager $authManager,
-        protected $channel = null,
-        protected $currency = null,
-        public $cart = null
+        protected ?Channel $channel = null,
+        protected ?Currency $currency = null,
+        public ?Cart $cart = null
     ) {
         //
     }
@@ -72,7 +72,7 @@ class CartSessionManager implements CartSessionInterface
      */
     public function manager(): ?Cart
     {
-        if (! $this->cart) {
+        if (! $this->cart?->exists) {
             $this->fetchOrCreate(create: true);
         }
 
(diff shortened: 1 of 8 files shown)


packages/core/src/Managers/CartSessionManager.php
```diff
@@ -122,48 +119,59 @@ private function fetchOrCreate($create = false, bool $estimateShipping = false)
             return $create ? $this->cart = $this->createNewCart() : null;
         }
 
-        $this->cart = $this->cart ?: Cart::with(
+        $this->cart = $this->cart?->exists ? $this->cart : Cart::with(
             config('lunar.cart.eager_load', [])
         )->find($cartId);
 
-        if (! $this->cart) {
+        if (! $this->cart?->exists) {
```
alecritson: Not sure we need the `exists` check here since it's already done above?

packages/core/src/Managers/CartSessionManager.php
```diff
@@ -122,48 +119,59 @@ private function fetchOrCreate($create = false, bool $estimateShipping = false)
...
             if (! $create) {
                 return null;
             }
 
             return $this->createNewCart();
```
alecritson: Is there potential to remove the nested `if` statement by doing:

```php
return $create ? $this->createNewCart() : null;
```

packages/core/src/Managers/CartSessionManager.php
glennjacobs: It's still possible for the cart not to exist, if the find fails.

alecritson: But in that scenario it will just return `null`, not an empty `Cart` model right?

glennjacobs: Oh, I see what you mean.

packages/core/src/Managers/CartSessionManager.php
```diff
@@ -122,48 +119,55 @@ private function fetchOrCreate($create = false, bool $estimateShipping = false)
...
-                $this->getShippingEstimateMeta(),
-                setOverride: true
-            );
+            $this->estimateShipping();
         }
 
         $this->use($this->cart);
 
-        return $this->cart->calculate();
+        return $this->cart;
```
alecritson: Since `use` returns the cart, can we just return the method call?

alecritson (approved):

PR #1810, opened 2024-06-10
Set min value based on currency factor on order management

alecritson:


packages/admin/src/Filament/Resources/OrderResource/Pages/ManageOrder.php
```diff
@@ -852,7 +852,9 @@ protected function getCaptureAction(): Actions\Action
                     ->default(fn ($record) => number_format($record->total->decimal, $record->currency->decimal_places, '.', ''))
                     ->live()
                     ->autocomplete(false)
-                    ->minValue(1)
+                    ->minValue(
+                        1 / $this->getRecord()->currency->factor
```
glennjacobs: Should we not inject the record, e.g. `->minValue(fn ($record)`?

alecritson: I was just keeping it the same as the other instance, if you'd prefer the inject route I can update both.

glennjacobs: Injecting I think is the preferred way to go about it and it's what is used directly above this in suffix + default, so I think it makes sense.

glennjacobs (approved):

PR #1811, opened 2024-06-11
Moving Asset registration of lunar-panel css to Panel definition

abishekrsrikaanth:
Currently the registration of the lunar-panel.css is being within `Filament:serving` method which is triggered only on a Filament Panel request. So, this causes an issue of the assets registered within it not being published when `php artisan filament:upgrade` is being called. Since `lunar-panel.css` needs to only be served for the Lunar panel requests, it is ideal that the asset registration can be moved where the Lunar Panel is being setup.


glennjacobs (approved):

glennjacobs:
Thanks @abishekrsrikaanth

PR #1812, opened 2024-06-13
Limit auth payload parameters based on Opayo limitations

alecritson:


glennjacobs:
@alecritson it looks like you've implemented the address char limits, but I'm not sure this is comprehensive.

packages/opayo/src/Opayo.php
```diff
@@ -87,18 +88,18 @@ public function getAuthPayload(AuthPayloadParameters $parameters): array
...
-                'city' => $parameters->billingAddressCity,
-                'postalCode' => $parameters->billingAddressPostcode,
+                'address1' => Str::limit($parameters->billingAddressLineOne, 50, ''),
+                'address2' => Str::limit($parameters->billingAddressLineTwo, 50, ''),
+                'address3' => Str::limit($parameters->billingAddressLineThree, 50, ''),
+                'city' => Str::limit($parameters->billingAddressCity, 40, ''),
+                'postalCode' => Str::limit($parameters->billingAddressPostcode, 10, ''),
                 'country' => $parameters->billingAddressCountryIso,
             ],
             'strongCustomerAuthentication' => [
```
glennjacobs, line 101: We don't know what browsers will give us, so we should probably put limits on these fields to match Opayo.

packages/opayo/src/Opayo.php
```diff
@@ -118,13 +119,13 @@ public function getAuthPayload(AuthPayloadParameters $parameters): array
...
-                'shippingCity' => $parameters->shippingAddressCity,
-                'shippingPostalCode' => $parameters->shippingAddressPostcode,
+                'recipientFirstName' => Str::limit($parameters->recipientFirstName, 20, ''),
+                'recipientLastName' => Str::limit($parameters->recipientLastName, 20, ''),
+                'shippingAddress1' => Str::limit($parameters->shippingAddressLineOne, 50, ''),
+                'shippingAddress2' => Str::limit($parameters->shippingAddressLineTwo, 50, ''),
+                'shippingAddress3' => Str::limit($parameters->shippingAddressLineThree, 50, ''),
+                'shippingCity' => Str::limit($parameters->shippingAddressCity, 40, ''),
+                'shippingPostalCode' => Str::limit($parameters->shippingAddressPostcode, 10, ''),
                 'shippingCountry' => $parameters->shippingAddressCountryIso,
```
glennjacobs: Do we put a limit of 2 on the country fields, just for completeness?

alecritson: Yeah you said you wanted all the max lengths so just started going down the list in the docs...

packages/opayo/src/Opayo.php
```diff
@@ -84,47 +85,47 @@ public function getAuthPayload(AuthPayloadParameters $parameters): array
...
```
glennjacobs: Missing a limit here on country

adevade:
You could use `Str::take($value, $limit)` instead, so you don't have to define empty string instead of trailing ellipsis for every call.

```php
/**
 * Take the first or last {$limit} characters of a string.
 *
 * @param  string  $string
 * @param  int  $limit
 * @return string
 */
public static function take($string, int $limit): string
```

[laravel/framework/src/Illuminate/Support/Str.php](https://github.com/laravel/framework/blob/c2776c320d96776b5d56c38659d0c55d76e0cded/src/Illuminate/Support/Str.php#L1586-L1600)

alecritson:
> You could use `Str::take($value, $limit)` instead, so you don't have to define empty string instead of trailing ellipsis for every call.
> 
> ```
> /**
>  * Take the first or last {$limit} characters of a string.
>  *
>  * @param  string  $string
>  * @param  int  $limit
>  * @return string
>  */
> public static function take($string, int $limit): string
> ```
> 
> [laravel/framework/src/Illuminate/Support/Str.php](https://github.com/laravel/framework/blob/c2776c320d96776b5d56c38659d0c55d76e0cded/src/Illuminate/Support/Str.php#L1586-L1600)

Thanks for the suggestion @adevade I think that looks nicer for sure.

glennjacobs (approved):

PR #1814, opened 2024-06-17
Product bundles

shepard153:
This pull request introduces the concept of Product Bundles. It allows you to create a Bundle and attach purchasable items to it. This PR is a work in progress, and I am seeking your feedback to ensure it aligns with Lunar's requirements before proceeding further. I would appreciate your input on whether my approach is correct.

Not sure about bundle SKU, EAN, etc., but I've copied that stuff for now to make it work. It's basically a copy of ProductVariant with little changes.

Also there is a problem with Discounts so I had to hack my way around but I was able to add bundle to cart in my shop.

### DIFF
--- a/packages/core/database/factories/BundleFactory.php
+++ b/packages/core/database/factories/BundleFactory.php
@@ -1,0 +1,30 @@
+<?php
+
+namespace Lunar\Database\Factories;
+
+use Illuminate\Database\Eloquent\Factories\Factory;
+use Illuminate\Support\Str;
+use Lunar\Models\Product;
+use Lunar\Models\Bundle;
+use Lunar\Models\TaxClass;
+use Lunar\Models\TaxRateAmount;
+
+class BundleFactory extends Factory
+{
+    protected $model = Bundle::class;
+
+    public function definition(): array
+    {
+        return [
+            'tax_class_id' => TaxClass::factory()->hasTaxRateAmounts(
+                TaxRateAmount::factory()
+            ),
+            'sku' => Str::random(12),
+            'unit_quantity' => 1,
+            'gtin' => $this->faker->unique()->isbn13,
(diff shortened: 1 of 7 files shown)


glennjacobs:
@shepard153 thanks for starting this. I'm going to set to draft for the moment, as it's not complete.

I will take a bit of time to think this one through, as we need to get it right. I'll come back later in the week.

glennjacobs:
I'm going to close this as it's not progressing and it's something the core team need to have a proper think about. We do intend to add bundle functionality at some point in the future.

vitalijalbu:
@shepard153 hi, can it be done separately by extening product resource ? seems no now wants to include extra functions here... 😒

Navi2016:
Still boggles my mind, a webshop without product bundles....how!? It's a core feature of any shop.

shepard153:
@vitalijalbu I've created workaround few days after this PR was set to draft. My client has product sample bundles and I've created separate entity called featured_samples along with Filament resource. User creates new entity and he can set name, description, image, price, active state and attach any product from the catalog. When ordered, the bundle is displayed as one item. Fortunately, there are no bundle discounts or items spliting, because it would require more work.

image

vitalijalbu:
> @vitalijalbu I've created workaround few days after this PR was set to draft. My client has product sample bundles and I've created separate entity called featured_samples along with Filament resource. User creates new entity and he can set name, description, image, price, active state and attach any product from the catalog. When ordered, the bundle is displayed as one item. Fortunately, there are no bundle discounts or items spliting, because it would require more work.
> 
> image

ok thanks, looks great but did you extended the product lunar resource? meaning is the re a sidebar nav that allows user to navigate and associate products?

can u share pls a fullscreen page of that? thanks a lot...

PR #1817, opened 2024-06-21
Force Media RM to always have 1 primary record

wychoong:
to always ensure 1 record is primary
- if current record is primary, unset others
- if current is not primary, set other as primary
- if no other record is primary, current will be primary
- after delete, if no primary record, set other as primary


alecritson:
Just thinking, would this sort of thing be better off in an observer/listener, since media can be changed without the panel this code will never fire.

wychoong:
> Just thinking, would this sort of thing be better off in an observer/listener, since media can be changed without the panel this code will never fire.

That is very true, and I have considered that. But why I didn’t implement as observer is because of bulk delete, it will trigger for all. 

I can implement as observer to be inline with `default` flag as other model.

alecritson:
Yeah we have something similar here https://github.com/lunarphp/lunar/blob/1.x/packages/core/src/Observers/UrlObserver.php if that was what you was referring to?

wychoong:
> Yeah we have something similar here https://github.com/lunarphp/lunar/blob/1.x/packages/core/src/Observers/UrlObserver.php if that was what you was referring to?

Yup

wychoong:
@alecritson I've updated the implementation to use observer

alecritson:
Thanks @wychoong it would be good to have a few test assertions for this as it's quite complex

wychoong:
> Thanks @wychoong it would be good to have a few test assertions for this as it's quite complex

Added the test.

should the observer apply to:
a. all collections
b. or only the one configured in `config('lunar.media.collection')`
c. or all collections only if the media is with `custom_properties.primary = true`

alecritson:
> > Thanks @wychoong it would be good to have a few test assertions for this as it's quite complex
> 
> Added the test.
> 
> should the observer apply to: a. all collections b. or only the one configured in `config('lunar.media.collection')` c. or all collections only if the media is with `custom_properties.primary = true`

Appreciate this @wychoong . I made a few tweaks, just to clean up tests a little and remove the need for an actual image in the filesystem. I also got rid of the nested if statements on the observer, hopefully it's all good. Well move to @glennjacobs for last review.

glennjacobs (approved):

PR #1820, opened 2024-06-22
Fix shipping options not retrieving unrestricted zones

wychoong:
for reference: https://discord.com/channels/497326121441034240/1248563023296073759


wychoong:
@glennjacobs would like a quick release on this fix. I did a quick manual test, it seems working fine and the change is just eloquent query fix, it shouldnt break anything

alecritson (approved):

PR #1828, opened 2024-06-28
Fixes to shipping methods

alecritson:
- Allow customer groups to associate to shipping methods
- Handle shipping method deletion, currently this would cause a integrity exception and when editing a shipping zone, if you had removed the shipping method an exception would be thrown
- Allow shipping rates to be removed


glennjacobs:
@alecritson if I try to delete a Shipping Zone I get an error. It's probably best we look to fix that at the same time.

glennjacobs:
@alecritson here's a Loom to demonstrate

https://www.loom.com/share/9f9a7de91f1e4367a60993a0ec5301c4?sid=799c1229-71da-4b16-8665-0b6c4803e7d1

glennjacobs (changes requested):
Need to ensure all delete methods are handled correctly.

alecritson:
> Need to ensure all delete methods are handled correctly.

Updates pushed

glennjacobs:
@alecritson customer group associations didn't exist previously. How will this work for guest customers that don't belong to a customer group?

alecritson:
> @alecritson customer group associations didn't exist previously. How will this work for guest customers that don't belong to a customer group?

https://github.com/lunarphp/lunar/blob/5e0515b1d7d86c959547015483c16dbf24994b10/packages/table-rate-shipping/src/Resolvers/ShippingRateResolver.php#L68

glennjacobs:
> > @alecritson customer group associations didn't exist previously. How will this work for guest customers that don't belong to a customer group?
> 
> https://github.com/lunarphp/lunar/blob/5e0515b1d7d86c959547015483c16dbf24994b10/packages/table-rate-shipping/src/Resolvers/ShippingRateResolver.php#L68

What I mean is, how do we control whether an option is (or is not) available to guests? I don't feel like the Availability page is very self-explanatory currently.

alecritson:
> > > @alecritson customer group associations didn't exist previously. How will this work for guest customers that don't belong to a customer group?
> > 
> > 
> > https://github.com/lunarphp/lunar/blob/5e0515b1d7d86c959547015483c16dbf24994b10/packages/table-rate-shipping/src/Resolvers/ShippingRateResolver.php#L68
> 
> What I mean is, how do we control whether an option is (or is not) available to guests? I don't feel like the Availability page is very self-explanatory currently.

Added a notice like products if a shipping method isn't available and also tweaked the creation action so it will set up the customer groups automatically.

maurice-ellis:
@glennjacobs @alecritson, I mentioned this in the Discord, but believe the same question has to be asked for Discounts for guests as well. Currently, to set a discount, the customer group has to be attached to a discount or else it won't apply because of this line.

https://github.com/lunarphp/lunar/blob/5e0515b1d7d86c959547015483c16dbf24994b10/packages/core/src/Managers/DiscountManager.php#L126

alecritson:
> @glennjacobs @alecritson, I mentioned this in the Discord, but believe the same question has to be asked for Discounts for guests as well. Currently, to set a discount, the customer group has to be attached to a discount or else it won't apply because of this line.
> 
> https://github.com/lunarphp/lunar/blob/5e0515b1d7d86c959547015483c16dbf24994b10/packages/core/src/Managers/DiscountManager.php#L126

Thanks for reporting, will look to get a separate PR done for this :)

glennjacobs:
@alecritson it looks like we're re-using the customer group UI, but we don't change the wording. See below...

image

glennjacobs:
@alecritson as this is a breaking change (because shipping methods will no longer be enabled without an associated customer group), I guess we should add something to the upgrade guide?

glennjacobs (changes requested):
Associations IU needs text changing and also need upgrade guide entry.

glennjacobs:
@alecritson, although not specific to this PR, it is now related. Do we need to change the text for the customer groups UI in other places, such as collections, discounts, etc.?

alecritson:
> @alecritson, although not specific to this PR, it is now related. Do we need to change the text for the customer groups UI in other places, such as collections, discounts, etc.?

It's likely, but yeah would be a separate PR

glennjacobs (approved):

maurice-ellis:
> @alecritson it looks like we're re-using the customer group UI, but we don't change the wording. See below...
> 
> image

Just wanted to chime in. Like product, can the default group already be added by default?

PR #1829, opened 2024-06-28
fix: staff resource roles test

wychoong:
because roles and permissions are hidden when `admin` is `true`, the state is not hydrated and is not `roles`'s `saveRelationshipsUsing` is not called...

https://github.com/lunarphp/lunar/blob/a941bafdff8a9c5dc543a60566a54cac5ce343ab/packages/admin/src/Filament/Resources/StaffResource.php#L149-L157

tested by repeating the test `->repeat(50)`
before
image

after
image


glennjacobs:
@wychoong I'm so confused as to why this has only run two (Vercel) checks and none of the others. Any ideas?

glennjacobs:
Ahhhh 
image

I'll wait until they're working again before I merge.

glennjacobs (approved):

PR #1835, opened 2024-07-01
Update `set-output` commands

adevade:
This PR updates the use of `set-output`command in shipping package. Now uses the new recommended way: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

Removes these warnings on the checks:
image

### DIFF
--- a/.github/workflows/admin_tests.yml
+++ b/.github/workflows/admin_tests.yml
@@ -24,7 +24,7 @@ jobs:
 
       - name: Get composer cache directory
         id: composer-cache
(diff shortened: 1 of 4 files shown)


adevade:
I don't know if the cached is even used. The output shows the following message both before and after this PR:

image

alecritson:
Thanks @adevade If we get this in we can always revisit, I think this is something to do with the scope in which the cache is available so might just need some tweaks on the repo itself :)

alecritson (approved):

PR #1837, opened 2024-07-02
Move to unified test runner for GitHub Workflow

adevade:
This PR moves all tests into one runner, while still running each suite separately with the help of the matrix. It runs all tests even if one matrix fails.

I have looked at and borrowed from Filaments test runner: [filamentphp/filament/.github/workflows/tests.yml](https://github.com/filamentphp/filament/blob/a51b7daa3f1249b534daae34516a9022e605f8ee/.github/workflows/tests.yml)


glennjacobs (approved):

glennjacobs:
Thanks @adevade, a good little tidy-up.

PR #1838, opened 2024-07-02
Bump dependencies in `monorepo-builder.php`

adevade:
This PR updates the dependencies listed in `monorepo-builder.php`, and runs the `monorepo-builder merge` command.

**Question:** What is the benefit of having these dev dependencies defined in `monorepo-builder.php`, instead of just in `composer.json`?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,16 +34,16 @@
         "spatie/laravel-blink": "^1.7",
         "spatie/laravel-medialibrary": "^11.0.0",
         "spatie/laravel-permission": "^6.4",
-        "stripe/stripe-php": "^7.114",
+        "stripe/stripe-php": "^14.4",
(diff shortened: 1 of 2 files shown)


glennjacobs:
Thanks. Sometimes this seems to get missed as it's a manual process. I'm not quite sure on the best approach to avoid it happening again.

> **Question:** What is the benefit of having these dev dependencies defined in `monorepo-builder.php`, instead of just in `composer.json`?

The individual packages don't have tests inside them, so therefore don't have the dev dependencies.

glennjacobs (approved):

adevade:
> The individual packages don't have tests inside them, so therefore don't have the dev dependencies.

If they're defined in the root `composer.json` dev-deps, would they be stripped out after a `monorepo-builder merge` command, if not defined through the php file?

PR #1846, opened 2024-07-05
Fix: Forget Cart in retrieving in same request

wychoong:
Fix an issue where using `CartSession::forget()` -> `CartSession::current()` in the same request, the `$this->cart` still holding the old cart. 

also fix an issue when `Cart::find()` returns `null` unable to assign to `public Cart $cart`

### DIFF
--- a/packages/core/src/Managers/CartSessionManager.php
+++ b/packages/core/src/Managers/CartSessionManager.php
@@ -70,6 +70,8 @@ class CartSessionManager implements CartSessionInterface
             );
         }
 
+        unset($this->cart);
+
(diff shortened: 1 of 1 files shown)


packages/core/src/Managers/CartSessionManager.php
```diff
@@ -70,6 +70,8 @@ public function forget(bool $delete = true): void
             );
         }
 
+        unset($this->cart);
```
wychoong, line 73: i couldn't replicate this anymore, but i think its no harm to add it

glennjacobs (approved):

PR #1847, opened 2024-07-05
Add tax rate editing to the panel

alecritson:


glennjacobs (changes requested):
@alecritson As we improve tax editing in the admin panel, let's also switch to using a Filament cluster to consolidate the three tax resources: Tax Classes, Tax Zones, and Tax Rates.
See https://filamentphp.com/docs/3.x/panels/clusters

glennjacobs:
@alecritson trying to edit a tax rate gives this incorrect error

image

glennjacobs (changes requested):
@alecritson When adding "Tax Rate Amounts" it allows us to add duplicate entries for the same "Tax Class".

Also, there is no facility to delete a Tax Rate Amount, only edit.

glennjacobs (approved):

PR #1851, opened 2024-07-05
Fixes attribute description migration in postgresql

### LINKED ISSUE #1850: Migration failure when upgrading to 1.0 - Postgresql
- Lunar version: 1.0.0-alpha29
- Laravel Version: 10.48.15
- PHP Version: 8.3.8
- Database Driver & Version: Postgresql 15.6

### Expected Behaviour:

- Running migrations during an upgrade should not fail.

### Actual Behaviour:

- core/database/migrations/2024_01_11_100000_add_description_to_attributes_table.php fails with a QueryException

<details>
<summary>QueryException</summary>

```
 Illuminate\Database\QueryException 

  SQLSTATE[23502]: Not null violation: 7 ERROR:  column "description" of relation "lunar_attributes" contains null values (Connection: pgsql, SQL: alter table "lunar_attributes" add column "description" json not null)

  at vendor/laravel/framework/src/Illuminate/Database/Connection.php:829
    825▕                     $this->getName(), $query, $this->prepareBindings($bindings), $e
    826▕                 );
    827▕             }
    828▕ 
  ➜ 829▕             throw new QueryException(
    830▕                 $this->getName(), $query, $this->prepareBindings($bindings), $e
    831▕             );
    832▕         }
    833▕     }

      +10 vendor frames 

  11  database/migrations/2024_01_11_100000_add_description_to_attributes_table.php:12
      Illuminate\Support\Facades\Facade::__callStatic()
      +25 vendor frames 

  37  artisan:35
      Illuminate\Foundation\Console\Kernel::handle()
```

</details>

### Steps To Reproduce:

- Use Postgresql or other DB that doesn't default columns to null.
- Have records in the attributes table before running migration.
- Execute migrations

jstoks:
fixes #1850 

Tested manually against postgresql 15

- Makes the column explicitly nullable to prevent migration failure for existing projects using Postgresql.

### DIFF
--- a/packages/core/database/migrations/2024_01_11_100000_add_description_to_attributes_table.php
+++ b/packages/core/database/migrations/2024_01_11_100000_add_description_to_attributes_table.php
@@ -9,7 +9,7 @@ return new class extends Migration
     public function up(): void
     {
         Schema::table($this->prefix.'attributes', function (Blueprint $table) {
-            $table->json('description')->after('name')->nullable(true);
+            $table->json('description')->after('name')->nullable();
         });
     }
 
--- a/packages/core/database/migrations/2024_01_11_100000_add_description_to_attributes_table.php
+++ b/packages/core/database/migrations/2024_01_11_100000_add_description_to_attributes_table.php
@@ -9,7 +9,7 @@ return new class extends Migration
     public function up(): void
     {
         Schema::table($this->prefix.'attributes', function (Blueprint $table) {
-            $table->json('description')->after('name');
+            $table->json('description')->after('name')->nullable(true);
         });
     }
 


packages/core/database/migrations/2024_01_11_100000_add_description_to_attributes_table.php
```diff
@@ -9,7 +9,7 @@
     public function up(): void
     {
         Schema::table($this->prefix.'attributes', function (Blueprint $table) {
-            $table->json('description')->after('name');
+            $table->json('description')->after('name')->nullable(true);
```
alecritson: I think this can just be `nullable()` since the default parameter is `true` anyway.

jstoks: 👍 You're right. I removed the argument.

alecritson (approved):

PR #1854, opened 2024-07-08
Fix cart id hanging around in the session when not cleared

alecritson:
Currently if a cart has a completed order but wasn't cleared from the session it will seize up and cause problems when users try to interact with it. This also causes data issues as the cart no longer maps to the completed order correctly.

### DIFF
--- a/packages/core/config/cart.php
+++ b/packages/core/config/cart.php
@@ -3,17 +3,6 @@
 use Lunar\Actions\Carts\GenerateFingerprint;
 
 return [
-
-    /*
-    |--------------------------------------------------------------------------
-    | Session Key
-    |--------------------------------------------------------------------------
-    |
-    | Specify the session key used when fetching the cart.
-    |
-    */
-    'session_key' => 'lunar_cart',
-
     /*
     |--------------------------------------------------------------------------
     | Fingerprint Generator
@@ -23,20 +12,6 @@ return [
     |
     */
     'fingerprint_generator' => GenerateFingerprint::class,
-
-    /*
-    |--------------------------------------------------------------------------
-    | Auto create a cart when none exists for user.
-    |--------------------------------------------------------------------------
-    |
-    | Determines whether you want to automatically create a cart for a user if
-    | they do not currently have one in the session. By default this is false
-    | to minimise the amount of cart lines added to the database.
-    |
-    */
-    'auto_create' => false,
-
-    'allow_multiple_per_order' => false,
 
     /*
     |--------------------------------------------------------------------------
--- a/packages/core/config/cart.php
+++ b/packages/core/config/cart.php
@@ -36,6 +36,8 @@ return [
     */
     'auto_create' => false,
 
+    'allow_multiple_per_order' => false,
+
     /*
     |--------------------------------------------------------------------------
     | Authentication policy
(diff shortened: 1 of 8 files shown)


packages/core/config/cart.php
```diff
@@ -36,6 +36,8 @@
     */
     'auto_create' => false,
 
+    'allow_multiple_per_order' => false,
```
glennjacobs: You can only have one cart per-order, so this probably wants to be rephrased, e.g.

```
allow_multiple_orders_per_cart
```

glennjacobs: We should probably include a comment description also.

glennjacobs:
@alecritson should we also use the config when placing an order via the cart session manager?

See https://github.com/lunarphp/lunar/blob/8e7814826fed7767240f4690d3a04753dd50d6bd/packages/core/src/Managers/CartSessionManager.php#L240

glennjacobs:
As this setting only applies to the cart session manager, and not the cart itself, maybe we should make that clear in the config? Should Cart Session Manager specific config be separated somehow?

packages/core/src/Managers/CartSessionManager.php
```diff
@@ -135,6 +135,10 @@ private function fetchOrCreate(bool $create = false, bool $estimateShipping = fa
             config('lunar.cart.eager_load', [])
         )->find($cartId);
 
+        if ($cart->hasCompletedOrders() && ! config('lunar.cart.allow_multiple_per_order', false)) {
```
glennjacobs: This config needs to be updated also to `cart_session`.

packages/core/src/Managers/CartSessionManager.php
```diff
@@ -25,13 +25,18 @@ public function __construct(
         //
     }
 
+    public function allowsMultipleOrdersPerCart(): bool
+    {
+        return config('lunar.cart.allow_multiple_per_order', false);
```
glennjacobs: Still the wrong config entry?

packages/core/config/cart_session.php
```diff
@@ -0,0 +1,38 @@
...
+    'session_key' => 'lunar_cart',
+
+    /*
+    |--------------------------------------------------------------------------
+    | Auto create a cart when none exists for user.
+    |--------------------------------------------------------------------------
+    |
+    | Determines whether you want to automatically create a cart for a user if
+    | they do not currently have one in the session. By default this is false
+    | to minimise the amount of cart lines added to the database.
```
glennjacobs: "cart lines" sounds wrong, I think it should just be "carts".

packages/core/config/cart_session.php
```diff
@@ -0,0 +1,38 @@
...
+    */
+    'auto_create' => false,
+
+    /*
+    |--------------------------------------------------------------------------
+    | Allow Carts to have multiple orders associated.
+    |--------------------------------------------------------------------------
+    |
+    | Determines whether the same cart instance will be returned if there is already
+    | a completed orders associated to the cart which is retrieved in the session.
```
glennjacobs: Should be "a completed order" not "orders".

glennjacobs (approved):

PR #1857, opened 2024-07-08
Adds the ability to specify the order statuses for the navigation count

glennjacobs:
You can now add the statuses the admin panel should use for the order count.

### DIFF
--- a/packages/admin/config/panel.php
+++ b/packages/admin/config/panel.php
@@ -46,6 +46,6 @@ return [
     | to include in the count below.
     |
     */
-    'order_count_statuses' => ['in-progress'],
+    'order_count_statuses' => ['payment-received'],
 
 ];
--- a/packages/admin/config/panel.php
+++ b/packages/admin/config/panel.php
@@ -35,4 +35,17 @@ return [
     |
     */
     'scout_enabled' => true,
+
+    /*
+    |--------------------------------------------------------------------------
+    | Navigation counts
+    |--------------------------------------------------------------------------
+    |
+    | The admin panel will show a count of orders in the left navigation.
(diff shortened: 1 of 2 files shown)


packages/admin/config/panel.php
```diff
@@ -35,4 +35,17 @@
...
+    |--------------------------------------------------------------------------
+    | Navigation counts
+    |--------------------------------------------------------------------------
+    |
+    | The admin panel will show a count of orders in the left navigation.
+    | This is based upon specific order statuses. You can define the statuses
+    | to include in the count below.
+    |
+    */
+    'order_count_statuses' => ['in-progress'],
```
alecritson: @glennjacobs Lunar doesn't ship with an `in-progress` status

glennjacobs: Good point haha!

alecritson:
Should we format this so it looks a bit nicer?
image

glennjacobs:
> Should we format this so it looks a bit nicer? image

Do you mean the number format? So it's `558,265`?

alecritson:
> > Should we format this so it looks a bit nicer? image
> 
> Do you mean the number format? So it's `558,265`?

Yeah, I guess it would need to take into account Locale as not everyone format numbers like that.

glennjacobs:
> > > Should we format this so it looks a bit nicer? image
> > 
> > 
> > Do you mean the number format? So it's `558,265`?
> 
> Yeah, I guess it would need to take into account Locale as not everyone format numbers like that.

Exactly, this could turn a simple PR into a much bigger issue, as then we'd have to check every count format in the admin panel.

alecritson (approved):

PR #1863, opened 2024-07-14
Fixing a misconfiguration problem for ListField type

xaimes:
1. Create a new attribute in the group
2. Select the type: List Field
3. Save
4. Activate
5. After entering the product displayed an error:

image

This happened because a null was written to the database, not an array, and the configuration is loaded in the function Lunar\Admin\Support\FieldTypes::getFilamentComponent.
https://github.com/lunarphp/lunar/blob/2b4d127e3d475cffa22f46c7a3e8bbc54cac87b3/packages/admin/src/Support/FieldTypes/ListField.php#L14

The solution to this problem, is to check during the creation if the value is null, if so, we replace it with an empty array.

The problem will also arise, for example, with the Youtube field and others that have no configuration.


xaimes:
Now I see that the problem has been solved here: #1853

while I would leave the configuration change from null to arrays, so that it does not generate problems in similar cases in the future.

alecritson:
Thanks @xaimes I appreciate you submitting this PR however it looks like this has been updated in another PR so there was a bit of overlap https://github.com/lunarphp/lunar/commit/fd2f972c79f68814f26479ef6f1b56993c7fb724 . Either way it should be solved now :)

PR #1874, opened 2024-07-18
Fixed an error in installation document

georgyKurian:
Updated the doc to add the missing LunarUser interface to the Laravel User model. 

Without the interface, it was causing an error on user login (Cart merge error)

### DIFF
--- a/docs/core/installation.md
+++ b/docs/core/installation.md
@@ -29,13 +29,14 @@ You may need to update your app's `composer.json` to set `"minimum-stability": "
 
 ### Add the LunarUser Trait
 
-Some parts of the core rely on the `User` model having certain relationships set up. We've bundled these into a trait which you must add to any models that represent users in your database.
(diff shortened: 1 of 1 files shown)


docs/core/installation.md
```diff
@@ -29,13 +29,14 @@ You may need to update your app's `composer.json` to set `"minimum-stability": "
 
 ### Add the LunarUser Trait
 
-Some parts of the core rely on the `User` model having certain relationships set up. We've bundled these into a trait which you must add to any models that represent users in your database.
+Some parts of the core rely on the User model having certain relationships set up. We have bundled these into a trait and an interface, which you must add to any models that represent users in your database.
 
 ```php
 use Lunar\Base\Traits\LunarUser;
+use Lunar\Base\LunarUser as BaseLunarUser;
```
wychoong:
```php
use Lunar\Base\LunarUser as LunarUserInterface;
// ...

class User extends Authenticatable implements LunarUserInterface
```

personally think this would be better as it tells clearly its an interface instead of "base class"

glennjacobs: I agree. Can we do that, please?

PR #1875, opened 2024-07-19
Handle cart value mismatch on Stripe

alecritson:
There are rare cases where the cart value is not the same as the amount which has been captured, in this instance we need to refund the amount and reset the cart.


Saifallak:
> There are rare cases where the cart value is not the same as the amount which has been captured, in this instance we need to refund the amount and reset the cart.

How could this happen or when?

PR #1877, opened 2024-07-22
Tweak Stripe logic

alecritson:


packages/stripe/src/Jobs/ProcessStripeWebhook.php
```diff
@@ -0,0 +1,77 @@
...
+
+        if (! $cart) {
+            Log::error(
+                "Unable to find cart with intent {$this->paymentIntentId}"
+            );
+
+            return;
+        }
+
+        $payment = Payments::driver('stripe')->cart($cart->calculate())->withData([
```
glennjacobs: Do we need to set the cart instance here? And I don't think the cart will always be available, so we're just setting it to null.

packages/stripe/src/StripePaymentType.php
```diff
@@ -52,7 +52,16 @@ public function __construct()
      */
     final public function authorize(): PaymentAuthorize
     {
-        $this->order = $this->cart->draftOrder ?: $this->cart->completedOrder;
+        $this->order = $this->order ?: ($this->cart->draftOrder ?: $this->cart->completedOrder);
+
+        if ($this->order && $this->order->placed_at) {
```
glennjacobs, line 57: If the order has already been placed, should we be firing a success response again?

packages/stripe/src/Jobs/ProcessStripeWebhook.php
```diff
@@ -0,0 +1,77 @@
...
+            return;
+        }
+
+        $payment = Payments::driver('stripe')->withData([
+            'payment_intent' => $this->paymentIntentId,
+        ]);
+
+        if ($order) {
+            $payment = $payment->order($order)->authorize();
+            PaymentAttemptEvent::dispatch($payment);
```
glennjacobs: Should we now be checking for null here, before firing the event?

alecritson: Looks like the payment type already sends this event

packages/stripe/src/Http/Controllers/WebhookController.php
```diff
@@ -38,25 +36,9 @@ public function __invoke(Request $request): JsonResponse
...
-                'message' => $error,
-            ], 400);
-        }
-
-        $payment = Payments::driver('stripe')->cart($cart->calculate())->withData([
-            'payment_intent' => $paymentIntent,
-        ])->authorize();
-
-        PaymentAttemptEvent::dispatch($payment);
+        ProcessStripeWebhook::dispatch($paymentIntent, $orderId)->delay(now()->addMinute());
```
glennjacobs: I can't see any check for the webhook event type. We don't want to process the order on a failure or similar.

alecritson: If we don't process a failure then we don't log the failed transaction and it won't appear in Lunar.

docs/core/upgrading.md
```diff
@@ -18,6 +18,16 @@ php artisan migrate
...
 
+## 1.0.0-alpha.34
+
+### Medium Impact
+
+#### Stripe Addon
+
+The Stripe driver will now check whether an order has a value for `placed_at` against an order and if so, no further processing will take place.
+
+Additionally, the logic in the webhook has been moved to a Job which is dispatched with a delay of 1 minute, this is to allow storefronts to manually process a payment intent, in addition to the webhook, without having to worry about overlap.
```
glennjacobs: Do we want to be more specific and say it uses the job "queue"?

packages/stripe/README.md
```diff
@@ -224,7 +227,7 @@ Stripe::getCharges(string $paymentIntentId);
 
 ## Webhooks
 
-The plugin provides a webhook you will need to add to Stripe. You can read the guide on how to do this on the Stripe website [https://stripe.com/docs/webhooks/quickstart](https://stripe.com/docs/webhooks/quickstart).
+The plugin provides an optional webhook you may add to Stripe. You can read the guide on how to do this on the Stripe website [https://stripe.com/docs/webhooks/quickstart](https://stripe.com/docs/webhooks/quickstart).
```
glennjacobs: Should we use the phrase "add-on", as I think we wanted to avoid "plugin" as that is what Filament uses.

packages/stripe/README.md
```diff
@@ -80,6 +80,9 @@ Make sure you have the Stripe credentials set in `config/services.php`
 'stripe' => [
     'key' => env('STRIPE_SECRET'),
     'public_key' => env('STRIPE_PK'),
+    'webhooks' => [
+        'payment_intent' => env('STRIPE_WEBHOOK_PAYMENT_INTENT'),
```
glennjacobs: Can we change this to `STRIPE_WEBHOOK_SIGNING_SECRET`?

glennjacobs (approved):

packages/stripe/src/StripePaymentType.php
hectnandez: Why didn't you return true then?

PR #1879, opened 2024-07-24
Ensure cart is calculated

alecritson:
Currently there is an edge case where if a cart is not calculated before `$cart->getShippingOption()` is called then it will return the incorrect shipping option from the shipping manifest when using the shipping add-on.

We're also not calculating the cart before we try and validate against it which could lead to false validation checks as all the cart information is not available.

This PR will call `calculate` when attempting to fetch the cart's shipping option and also before validation takes place when creating an order.

### DIFF
--- a/packages/core/src/Models/Cart.php
+++ b/packages/core/src/Models/Cart.php
@@ -572,7 +572,7 @@ class Cart extends BaseModel
      */
     public function getShippingOption(): ?ShippingOption
     {
-        return ShippingManifest::getShippingOption($this);
+        return ShippingManifest::getShippingOption($this->calculate());
     }
 
     /**
@@ -592,6 +592,8 @@ class Cart extends BaseModel
         bool $allowMultipleOrders = false,
         ?int $orderIdToUpdate = null
     ): Order {
+        $this->refresh()->recalculate();
+
         foreach (config('lunar.cart.validators.order_create', [
             ValidateCartForOrderCreation::class,
         ]) as $action) {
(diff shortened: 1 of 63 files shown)


packages/core/src/Models/Cart.php
```diff
@@ -592,18 +592,20 @@ public function createOrder(
         bool $allowMultipleOrders = false,
         ?int $orderIdToUpdate = null
     ): Order {
+        $cart = $this->refresh()->recalculate();
```
glennjacobs, line 595: Is it important to use recalculate here, as apposed to calculate?

alecritson: This was how it was previously and I think it ensures we have the latest version of the cart to create an order from.

glennjacobs (approved):

packages/core/src/Models/Cart.php
```diff
@@ -572,7 +572,7 @@ public function setShippingOption(ShippingOption $option, bool $refresh = true):
      */
     public function getShippingOption(): ?ShippingOption
     {
-        return ShippingManifest::getShippingOption($this);
+        return ShippingManifest::getShippingOption($this->calculate());
```
wychoong, line 575: @alecritson what is the need for calling `calculate()` here? for what this method concern, it should just need `refresh()` if it wants the latest shipping address and shipping option

currently I'm facing an issue where we use $cart->getShippingOption() in a calculate pipeline and its causing infinite loop

PR #1882, opened 2024-07-24
Feature - Add new helper localeUrl

lguichard:
When you manage multi language it might be useful to have an helper localeUrl to display localize slug


glennjacobs (changes requested):
It looks good, but could we add something to the docs so other developers are aware of it?

glennjacobs (approved):

PR #1884, opened 2024-07-29
Fix - Error product null if create product variant record on resource

lguichard:


glennjacobs:
@lguichard how do we reproduce this error?

lguichard:
@glennjacobs If the dev create a custom ProductVariantResource and use Attributes component to render attribute data field, the EditRecord generate product null error in create record process. 

 The error is introduce in this pr : https://github.com/lunarphp/lunar/commit/ecefeeba6f2313584a07fad78f956b093c7ea391

alecritson (approved):

PR #1888, opened 2024-07-30
Set payment order to null when setting the cart

alecritson:
There is a potential for a payment provider to have access to a previously set order when calling the `order()` method on a payment type.

Since we always need a cart but not always an order we should be okay with setting the order to `null` when calling the `cart()` method.


glennjacobs (approved):

wychoong:
currently I dont have issue with this change, but this feels odd to make cart and order mutually exclusive. say if I want to chain `->order($order)->cart($cart)`, it will not be possible after this change

would it be better to check if the order belongs to the cart before setting to null?

PR #1889, opened 2024-07-30
Feature -  SpatieHasMedia interface must be use in ProductVariant

lguichard:
This feature is usefull if the dev need to add media in the variant. Currently, the ProductVariant use onlymedia_product_variant to link images inherit from Product


glennjacobs:
Lunar doesn't currently offer media at the ProductVariant level, so this would be wrong. If you need such a thing, you can extend the model etc in your application (model extending update will be merged soon).

PR #1894, opened 2024-08-05
Fixed and added additional config for File Fieldtype

maurice-ellis:
This pull request fixes the File Attribute and its issue of receiving a `Livewire\Exceptions\CorruptComponentPayloadException` and not uploading the file. Additionally, this fieldtype can handle multiple files and configuration options were added: File Types, Multiple Files, Min Files, and Max Files.

Of importance is the change in the handling of the dehydrated state for the File Attribute in  [AttributeData](https://github.com/maurice-ellis/lunar/blob/3e19fc1a7f6a82ceef814e74c0cd71249b798e9c/packages/admin/src/Support/Forms/AttributeData.php#L65) that happens after the file upload.

Before dehydrating, files are uploaded by [BaseFileUpload](https://github.com/filamentphp/filament/blob/b083225530c336ae86cf288a276d139dd0057b82/packages/forms/src/Components/BaseFileUpload.php#L130) and a string or array is returned, which was not properly handled previouslly.

### DIFF
--- a/packages/admin/resources/lang/en/fieldtypes.php
+++ b/packages/admin/resources/lang/en/fieldtypes.php
@@ -53,5 +53,19 @@ return [
     ],
     'file' => [
         'label' => 'File',
+        'form' => [
+            'file_types' => [
+                'label' => 'Allowed File Types',
+            ],
+            'multiple' => [
+                'label' => 'Allow Multiple Files',
+            ],
+            'min_files' => [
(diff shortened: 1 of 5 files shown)


alecritson (approved):
Thanks @maurice-ellis I had a little play and it seemed to work well 🙇

maurice-ellis:
@alecritson Great, glad I can help!

PR #1897, opened 2024-08-12
Store address information that comes from Stripe

alecritson:
When using Stripe frontend tools, such as Express checkout, we don't always get the complete address information sent back. This can cause a problem for Lunar as we're potentially storing a small subset of what is actually on the payment intent.

This PR looks to add a new action which updates the billing and shipping addresses on Lunar during placement, using the payment intent as the source of truth.

This is enabled/disabled via a new config option

```
// config/lunar/stripe.php

<?php

return [
    // ...
    'sync_addresses' => true,
];
```

### DIFF
--- a/packages/stripe/src/Actions/StoreAddressInformation.php
+++ b/packages/stripe/src/Actions/StoreAddressInformation.php
@@ -33,7 +33,7 @@ class StoreAddressInformation
             $shippingAddress->city = $stripeShipping->city;
             $shippingAddress->state = $stripeShipping->state;
             $shippingAddress->postcode = $stripeShipping->postal_code;
-            $shippingAddress->country_id = $country->id;
+            $shippingAddress->country_id = $country?->id;
             $shippingAddress->contact_phone = $paymentIntent->shipping->phone;
             $shippingAddress->save();
         }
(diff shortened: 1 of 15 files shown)


packages/stripe/src/Actions/StoreAddressInformation.php
```diff
@@ -0,0 +1,56 @@
...
+        if ($paymentIntent->shipping && $stripeShipping = $paymentIntent->shipping->address) {
+            $country = Country::where('iso2', $stripeShipping->country)->first();
+            $shippingAddress->first_name = explode(' ', $paymentIntent->shipping->name)[0];
+            $shippingAddress->last_name = explode(' ', $paymentIntent->shipping->name)[1] ?? '';
+            $shippingAddress->line_one = $stripeShipping->line1;
+            $shippingAddress->line_two = $stripeShipping->line2;
+            $shippingAddress->city = $stripeShipping->city;
+            $shippingAddress->state = $stripeShipping->state;
+            $shippingAddress->postcode = $stripeShipping->postal_code;
+            $shippingAddress->country_id = $country->id;
```
glennjacobs: What happens if we don't have a country?

alecritson: Currently assumes we will, considering we auto seed the countries on install and there's no way to manage them it's unlikely they would have been removed or changed. 

If one really didn't exist we wouldn't be able to assume anything so it'd have to bail out, we don't have any country management so no defaults etc.

glennjacobs:
@alecritson should this have a test?

alecritson:
> @alecritson should this have a test?

Test has been added

glennjacobs (approved):

glennjacobs (approved):

PR #1899, opened 2024-08-13
Add Postgres support for query sorting

alecritson:
This PR looks to add a new `orderBySequence` macro to the Eloquent Builder class which adds PostgreSQL support when needing to order records from the query builder based on an array of ID's in a certain order.

### DIFF
--- a/packages/admin/src/Support/Pages/BaseListRecords.php
+++ b/packages/admin/src/Support/Pages/BaseListRecords.php
@@ -45,7 +45,7 @@ abstract class BaseListRecords extends ListRecords
 
             $query->when(
                 ! $ids->isEmpty(),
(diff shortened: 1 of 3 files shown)


packages/admin/src/Support/Pages/BaseListRecords.php
```diff
@@ -45,7 +45,7 @@ protected function applySearchToTableQuery(Builder $query): Builder
 
             $query->when(
                 ! $ids->isEmpty(),
-                fn ($query) => $query->orderByRaw("field(id, {$placeholders})", $ids->toArray()) // TODO: Only supports MySQL
+                fn ($query) => $query->orderBySequence($ids->toArray()) // TODO: Only supports MySQL
```
glennjacobs: The comment can be removed now

glennjacobs (approved):

PR #1904, opened 2024-08-14
Add cart stock validators to config

alecritson:


docs/core/reference/carts.md
```diff
@@ -285,6 +285,8 @@ The other available methods are as follows:
 
 ### Add a cart line
 
+> When adding purchasables to a cart, Lunar will run validation against the quantity and stock levels to ensure the item can be added. You can view these validators in `config/lunar/cart.php`.
```
glennjacobs: This is quite light weight. We don't explain what happens if a validator fails for example and how to handle it.

glennjacobs (approved):

PR #1914, opened 2024-08-20
Add Translations to product type.

Saifallak:
Add Translations to product type.


Saifallak:
Why closed, product type should be translatable

PR #1928, opened 2024-08-30
Manually map models on migration instead of loading via directory

alecritson:
Currently if you publish the migrations they can fail due to trying to read a directory relative to the the vendor folder. This PR instead opts to manually map the models to prevent this issue completely.

### DIFF
--- a/packages/core/database/migrations/2024_03_15_100000_remap_polymorphic_relations.php
+++ b/packages/core/database/migrations/2024_03_15_100000_remap_polymorphic_relations.php
@@ -55,7 +55,7 @@ class RemapPolymorphicRelations extends Migration
                 $class => \Lunar\Facades\ModelManifest::getMorphMapKey($class),
             ]
         );
-        
+
         $tables = [
             'attributables' => ['attributable_type'],
             'attributes' => ['attribute_type'],
--- a/packages/core/database/migrations/2024_03_15_100000_remap_polymorphic_relations.php
+++ b/packages/core/database/migrations/2024_03_15_100000_remap_polymorphic_relations.php
@@ -1,25 +1,61 @@
 <?php
 
 use Illuminate\Support\Facades\Schema;
-use Lunar\Base\BaseModel;
 use Lunar\Base\Migration;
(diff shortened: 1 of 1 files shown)


packages/core/database/migrations/2024_03_15_100000_remap_polymorphic_relations.php
```diff
@@ -1,20 +1,56 @@
...
-use Lunar\Base\BaseModel;
 use Lunar\Base\Migration;
-use Spatie\StructureDiscoverer\Discover;
 
 class RemapPolymorphicRelations extends Migration
 {
     public function up()
     {
-        $modelClasses = collect(
-            Discover::in(__DIR__.'/../../src/Models')
```
wychoong, line 13: what about `base_path('vendor/lunarphp/core/src/Models')`
https://laravel.com/docs/11.x/helpers#method-base-path

glennjacobs: Yeah, I would have taken that approach also.

alecritson: I tried this but it wouldn't work for the testsuite as Orchestra loads an instance of a Laravel app within itself so the base path would be relative to that install, which doesn't have a vendor directory

glennjacobs: I guess it's only a migration, not something that will need to be maintained. So it's fine.

alecritson: Can you clarify what's fine, the proposed change here or the current change in the PR?

glennjacobs: I'm happy with the PR as it is. If the base_path() won't work, then I don't think we worry about it.

glennjacobs (approved):

glennjacobs (approved):

PR #1932, opened 2024-09-02
Extending models breaks pipelines

repl6669:
This PR just aims to prove that in certain cases model extending breaks pipelines.

I just swapped `Cart` and `Order` models, let's call them `CustomCart` and `CustomOrder`.

**Problem:**
The problem appears to be double serialization of certain casted attributes, specifically `$shippingBreakdown` and `$taxBreakdown`, because at first the model is `Cart`, but when it goes through pipelines, it is resolved as `CustomCart` and this namespace corrupts attribute serialization, so when it tries to set eg. `$order->shipping_breakdown`, it tries to set a json encoded value instead of the `ShippingBreakdown` class.

Specifically this part of `ShippingBreakdown` / `TaxBreakdown` gets called when it should not.

```php
{
    return json_encode(
        $this->set($model, $key, $value, $attributes)
    );
}
```

**Possible solution:**
Resolve correct models in pipelines and probably everywhere possible like this:
```php
use Lunar\Models\Contracts\Cart as CartContract;
use Lunar\Models\Contracts\OrderLine as OrderLineContract;

final class CreateOrder extends AbstractAction
{
    /**
     * Execute the action.
     */
    public function execute(
        CartContract $cart,
        ?int $orderIdToUpdate = null
    ): self {

    // ...
    
    // Instead of new OrderLine
    $orderLine = App::make(OrderLineContract::class);
```

I am ready to do the work with tests. I just need you to confirm, that this is the right solution in your opinion. I already tested this solution here: https://github.com/dystcz/lunar-api/pull/166 by swapping the necessary pipelines with updated container resolution.

### DIFF
--- a/packages/core/database/factories/CartFactory.php
+++ b/packages/core/database/factories/CartFactory.php
@@ -2,9 +2,9 @@
 
 namespace Lunar\Database\Factories;
 
-use Lunar\Models\Cart;
 use Lunar\Models\Channel;
 use Lunar\Models\Currency;
+use Lunar\Tests\Core\Stubs\Models\Cart;
 
 class CartFactory extends BaseFactory
 {
--- a/packages/core/database/factories/OrderFactory.php
+++ b/packages/core/database/factories/OrderFactory.php
@@ -4,7 +4,7 @@ namespace Lunar\Database\Factories;
 
 use Lunar\Base\ValueObjects\Cart\TaxBreakdown;
 use Lunar\Models\Channel;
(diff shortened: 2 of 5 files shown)


repl6669:
Here is the failing test: https://github.com/lunarphp/lunar/actions/runs/10658743988/job/29540287076?pr=1932

repl6669:
Possible implementation here: https://github.com/lunarphp/lunar/pull/1933

PR #1933, opened 2024-09-02
Fix container resolution and morph map for extended models

### LINKED ISSUE #1930: Unable to create order
- Lunar version: 1.0-Beta
- Laravel Version: 11
- PHP Version: 8.2.1
- Database Driver & Version:

### Expected Behaviour:
Creating order from basket should work

### Actual Behaviour:
Getting Shipping breakdown must be instance of Lunar\Base\ValueObjects\Cart\ShippingBreakdown.

Works in previously versions.

### Steps To Reproduce:

  // Create a new cart for the user
            $cart = Cart::create([
                'currency_id' => Currency::first()->id,
                'channel_id' => Channel::first()->id,
                'user_id' => $user->id,
            ]);

            $cart->lines()->create([
                'purchasable_type' => ProductVariant::class,
                'purchasable_id' => $this->product->variants()->first()->id,
                'quantity' => 1,
            ]);

            // Set the billing address for the cart
            $cart->setBillingAddress([
                'first_name' => $user->name,
                'last_name' => $user->name,
                'line_one' => '123',
                'city' => 'Greenified',
                'postcode' => 2600,
                'country_id' => 61,
            ]);

            // Create the order from the cart
            $order = $cart->createOrder();

repl6669:
This PR aims to fix https://github.com/lunarphp/lunar/issues/1930 as proposed here: https://github.com/lunarphp/lunar/pull/1932

**Problem:**
The problem appears to be double serialization of certain casted attributes, specifically `$shippingBreakdown` and `$taxBreakdown`, because at first the model is `Cart`, but when it goes through pipelines, it is resolved as `CustomCart` and this namespace corrupts attribute serialization, so when it tries to set eg. `$order->shipping_breakdown`, it tries to set a json encoded value instead of the `ShippingBreakdown` class.

Specifically this part of `ShippingBreakdown` / `TaxBreakdown` gets called when it should not.

```php
{
    return json_encode(
        $this->set($model, $key, $value, $attributes)
    );
}
```

**Possible solution:**
Resolve correct models in pipelines:
```php
use Lunar\Models\Contracts\Order as OrderContract;
use Lunar\Models\Contracts\OrderLine as OrderLineContract;

public function handle(OrderContract $cart, Closure $next): mixed
{
    // ...
    
    // Instead of new OrderLine
    $orderLine = App::make(OrderLineContract::class);

    // ...
}
```

- [x] Sample implementation
- [x] Tests


repl6669:
This change https://github.com/lunarphp/lunar/commit/edb019fe38356896178faac23658ae8ff90cc2fc was the missing piece to get green in our projects.

repl6669:
I know this PR is huge, but here is the gist.

I basically went through the whole codebase and did just a few things:

1. Replace concrete model implementations in method arguments (`Lunar\Models\Product` → `Lunar\Models\Contracts\Product`) as well as method returns types
2. Instead of directly calling static properties on a concrete implementation, it is now prefixed with `::modelClass()` (`Lunar\Models\Order::find()` → `Lunar\Models\Order::modelClass()->find()`
3. Introduce new test scenario where all models are swapped (if `LUNAR_TESTING_REPLACE_MODELS` env var is true, it will bind `TestModelManifest` singleton which registers stub models), Github test workflow is updated as well to reflect this change
4. There is a new `phpunit.extending.xml` which can be used for local testing as follows ` ./vendor/bin/pest -c phpunit.extending.xml --testsuite core --bail`
5. Test `"can get new draft order when cart changes"` was skipped, because it seems to be a false positive

I also tested that everything works in https://github.com/dystcz/lunar-api, where I extend almost every Lunar model and it works great.

packages/core/src/LunarServiceProvider.php
```diff
@@ -161,6 +162,10 @@ public function register(): void
         });
 
         $this->app->singleton(ModelManifestInterface::class, function ($app) {
+            if ($app->environment('testing') && env('LUNAR_TESTING_REPLACE_MODELS', false)) {
```
wychoong: should use 
```php
    ModelManifest::replace(
            \Lunar\Models\Contracts\Product::class,
            \Lunar\Tests\Core\Stubs\Models\Product::class
        );
```
in test suit setup to make sure it is as close as user's app

tests/core/Unit/Base/ModelManifestTest.php
```diff
@@ -33,9 +33,14 @@
...
+    $expected = env('LUNAR_TESTING_REPLACE_MODELS', false)
+    ? \Lunar\Tests\Core\Stubs\Models\Product::class
+    : \Lunar\Models\Product::class;
+
     expect(
         ModelManifest::get(\Lunar\Models\Contracts\Product::class)
-    )->toBe(Product::class);
+    )->toBe($expected);
 
     ModelManifest::replace(
```
wychoong, line 42: line 45-52 seems redundant now

packages/core/src/LunarServiceProvider.php
alecritson: Yeah I don't think this should be in the LunarServiceProvider. If something is needed in the test env it should be set up on the TestCase class or similar.

alecritson:
@theimerj Looks like there was some overlap with a PR, are you able to take a look at the conflicts?

repl6669:
> @theimerj Looks like there was some overlap with a PR, are you able to take a look at the conflicts?

Hey @alecritson, I will get to it soon. Thank you @wychoong for pointing these things out.

packages/core/src/LunarServiceProvider.php
repl6669: Removed from SP and moved to "base" `TestCase` which is extended by all internal `TestCases`, which call the `replaceModelsForTesting` method in `getEnvironmentSetUp` method.

https://github.com/lunarphp/lunar/blob/f98a015639b2a8944039700b21203e5f255ca90b/tests/TestCase.php#L21-L39

tests/core/Unit/Base/ModelManifestTest.php
repl6669: Now testing against `\Lunar\Tests\Core\Stubs\Models\Custom\CustomProduct::class`

https://github.com/lunarphp/lunar/blob/113b3d65ea64416f8db3270f426f4baa073b4c19/tests/core/Unit/Base/ModelManifestTest.php#L35-L50

alecritson:
Thanks for this PR, I've had a little look and we've discussed it internally briefly, one point we raised was the introduction of `modelClass()` when using eloquent such as:

```php
Language::modelClass()::getDefault();
```

I think this feels a bit off and not something we'd really want throughout the codebase as it doesn't really feel like a natural use of Eloquent and is a bit of a maintenance overhead we'd rather not introduce.

Was the main purpose of this so that you are always interacting with the extended model? Since I don't think the panel code should be concerned with this and using the Lunar models should just work as well. There might have been an issue with resolving the correct model by Lunar but hopefully this should be resolved in #1949 as there was a bug with this.

Additionally I think that same PR should resolve the issue with the shipping breakdown hydration 🤔

repl6669:
1. Ok, I understand your concern, but Lunar's responsibility is to ensure that it always resolves the correct concrete implementation. To me it basically seems wrong to use concrete implementations of Lunar models throughout the codebase  when you want to offer the ability to replace them. And the way to go around this problem is basically trying to find the Lunar model in ancestors and go from there. What if the replaced model does not extend Lunar model and just implements the contract? Unlikely, but there is no check that the concrete implementation has to extend a Lunar model. I just went with the `::modelClass()`, because it was short. There are other ways to resolve the concrete class out of the container.

2. I already checked #1949 and am not sure that it will work with more than one level of inheritance which may happen. I am not saying that this is the best practice, but I am sure you get my point. 

3. At the same time this part seems to me like an introduction of possible performance issues if this kind of rehydration happens too many times.
https://github.com/lunarphp/lunar/blob/047bfb2e9051906471e67810bc22011265457d34/packages/core/src/Base/Traits/HasModelExtending.php#L25-L30
Plus do we know that this works with all types of relationships? Eg. "through" relationships and so on? Shouldn't this use the `::replicate()` method provided by laravel, just modified to replicate into a different class to make it more robust? Is this even necessary?

The solution introduced in this PR seems tedious at first, which it is (I went through the process), but if you comply with the correct way of resolving implementations out of the container in the future as well, the problems introduced are basically mitigated for good.

Lastly I find the introduction of tests which cover the functionality with extended models in mind is quite important in order to globally catch possible bugs which may arise in the future.

--- 

Bottom line: I tested the #1949 briefly locally and it really seems to have resolved the discovered issues. However, the test coverage is not ideal and I personally believe that the whole codebase should rely on automatic dependency injection more (at the very least in all the method calls where you can type hint the interfaces instead of concrete implementations). This alone mitigates a lot of problems.

Let me know how to proceed. We can scratch this PR all together. We can keep some of the changes and scrach the rest you don't like. Your call. Please let me know and I will act accordingly. Thank you.

alecritson:
> 1. Ok, I understand your concern, but Lunar's responsibility is to ensure that it always resolves the correct concrete implementation. To me it basically seems wrong to use concrete implementations of Lunar models throughout the codebase  when you want to offer the ability to replace them. And the way to go around this problem is basically trying to find the Lunar model in ancestors and go from there. What if the replaced model does not extend Lunar model and just implements the contract? Unlikely, but there is no check that the concrete implementation has to extend a Lunar model. I just went with the `::modelClass()`, because it was short. There are other ways to resolve the concrete class out of the container.

I think you're correct in the sense that Lunar should be able to resolve the concrete class, but the underlying core code does not need to resolve to those classes, Lunar should should still be able to use it's own models. 

The point of model extending is to allow developers to bring their own model classes, which extend the Lunar ones, to add functionality on top of Lunar. But also for Lunar itself to not be concerned about that and still use the base models under the hood. For example if the developer adds their own model method, the Lunar core is never going to try and call that method so it doesn't really need to know about it.

I think if we wanted to go down the path of Lunar using the concrete class then really we should use dependency injection.


> 2. I already checked [Fix extended model resolving #1949](https://github.com/lunarphp/lunar/pull/1949) and am not sure that it will work with more than one level of inheritance which may happen. I am not saying that this is the best practice, but I am sure you get my point.

We are currently not concerned about more than one level of inheritance and would consider it unsupported.

> 3. At the same time this part seems to me like an introduction of possible performance issues if this kind of rehydration happens too many times.
>    https://github.com/lunarphp/lunar/blob/047bfb2e9051906471e67810bc22011265457d34/packages/core/src/Base/Traits/HasModelExtending.php#L25-L30
>    
>    Plus do we know that this works with all types of relationships? Eg. "through" relationships and so on? Shouldn't this use the `::replicate()` method provided by laravel, just modified to replicate into a different class to make it more robust? Is this even necessary?

Yeah I understand what you're saying here and I updated the PR to use the `replicate` method, thanks for the heads up about that. I don't foresee any performance issues.

> 
> The solution introduced in this PR seems tedious at first, which it is (I went through the process), but if you comply with the correct way of resolving implementations out of the container in the future as well, the problems introduced are basically mitigated for good.
> 
> Lastly I find the introduction of tests which cover the functionality with extended models in mind is quite important in order to globally catch possible bugs which may arise in the future.
> 
> Bottom line: I tested the #1949 briefly locally and it really seems to have resolved the discovered issues. However, the test coverage is not ideal and I personally believe that the whole codebase should rely on automatic dependency injection more (at the very least in all the method calls where you can type hint the interfaces instead of concrete implementations). This alone mitigates a lot of problems.
> 
> Let me know how to proceed. We can scratch this PR all together. We can keep some of the changes and scrach the rest you don't like. Your call. Please let me know and I will act accordingly. Thank you.


In short I think the best things to take away from the PR are:

- Using DI where possible and type methods to their contracts
- Using extended models in the test suite seems like good coverage.

The only thing we wouldn't include is the `::modelClass()` call.

repl6669:
Hello @alecritson, thank you for your reply.

> I think you're correct in the sense that Lunar should be able to resolve the concrete class, but the underlying core code does not need to resolve to those classes, Lunar should should still be able to use it's own models.
> 
> The point of model extending is to allow developers to bring their own model classes, which extend the Lunar ones, to add functionality on top of Lunar. But also for Lunar itself to not be concerned about that and still use the base models under the hood. For example if the developer adds their own model method, the Lunar core is never going to try and call that method so it doesn't really need to know about it.
> 
> I think if we wanted to go down the path of Lunar using the concrete class then really we should use dependency injection.

Understood. Your approach actually [made me refactor](https://github.com/dystcz/lunar-api/pull/178) Lunar API models, because creating deeply nested inheritance is not a way to go.

> We are currently not concerned about more than one level of inheritance and would consider it unsupported.

Ok, if only one level of inheritance will be supported by Lunar it should probably be written in the docs and a check could be introduced in the model manifest. I can prepare something.

> Yeah I understand what you're saying here and I updated the PR to use the `replicate` method, thanks for the heads up about that. I don't foresee any performance issues.

Great to hear. That should minimize the performance issue I guess.

> In short I think the best things to take away from the PR are:
> 
> * Using DI where possible and type methods to their contracts
> * Using extended models in the test suite seems like good coverage.
> 
> The only thing we wouldn't include is the `::modelClass()` call.

Ok, I will merge conflicts and rework the PR removing `::modlClass()`.

Thanks again. Have a great day.

glennjacobs:
@alecritson conflicts resolved and all tests passing. This will need a thorough review.

glennjacobs (approved):

PR #1937, opened 2024-09-04
Support selection of the highest tax rate in the basket for shipping

ryanmitchell:
Currently the shipping tables add-on only supports returning the default tax class/rate.

However countries can have different requirements for how tax should be calculated on shipping depending on what is in the basket, for example in the UK shipping should take the most expensive tax rate in the basket.

This PR updates the add-on with a new config option (`lunar.shipping-tables.shipping_rate_tax_calculation`) which currently has two options: `default` for selecting the default rate, or `highest` for selecting the highest rate in the basket.


ryanmitchell:
Any thoughts on this one?

glennjacobs (approved):

PR #1938, opened 2024-09-04
Allow extending of page edit record forms

ryanmitchell:
This PR adds the ability to extend base edit record pages.

Our use case is the ability to add to the inventory pages, which is why I've applied this to them.

### DIFF
--- a/packages/admin/src/Support/Pages/BaseEditRecord.php
+++ b/packages/admin/src/Support/Pages/BaseEditRecord.php
@@ -9,6 +9,7 @@ abstract class BaseEditRecord extends EditRecord
 {
     use Concerns\ExtendsFooterWidgets;
     use Concerns\ExtendsFormActions;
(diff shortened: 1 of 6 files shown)


packages/admin/src/Support/Pages/BaseEditRecord.php
```diff
@@ -9,6 +9,7 @@ abstract class BaseEditRecord extends EditRecord
 {
     use Concerns\ExtendsFooterWidgets;
     use Concerns\ExtendsFormActions;
+    use Concerns\ExtendsForms;
```
wychoong, line 12: should apply the same to `BaseCreateRecord`

ryanmitchell: done

glennjacobs (approved):

PR #1944, opened 2024-09-09
fix extended model morph map by checking parent class map

wychoong:
currently after model extension, when loading relationship it is unable to retrieve correct morph type

setup
```php
  public function boot(ShippingModifiers $shippingModifiers): void
{
    \Lunar\Facades\ModelManifest::replace(
        \Lunar\Models\Contracts\Product::class,
        \App\Models\Product::class,
    );
}
```

laravel will be looking for `App\Models\Product` in the morphMap array but it will not be able to resolve to `product` since Lunar registered 'product' => 'Lunar\Models\Product
https://github.com/lunarphp/lunar/blob/2b31dcabc4cc7f169051be63648836b75bf91d18/packages/core/src/Base/ModelManifest.php#L153-L176

earlier a workaround would be override the morphMap
```php
Relation::morphMap([
    'product' => \App\Models\Product::class,
]);
```

this fix resolve this by searching the morphMap array using the Lunar model's FQCN
it also checks if the extended model is already manually registered in morphMap by the user

### DIFF
--- a/packages/core/src/Base/ModelManifest.php
+++ b/packages/core/src/Base/ModelManifest.php
@@ -143,7 +143,7 @@ class ModelManifest implements ModelManifestInterface
         return 'Lunar\\Models\\'.$shortName;
     }
 
-    public function isLunarModel(BaseModel $model): bool
+    public function isLunarModel(string|BaseModel $model): bool
     {
         $class = (new \ReflectionClass($model));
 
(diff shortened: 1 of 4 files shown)


alecritson:
Thanks @wychoong it seems to do the trick, I added a simple test and also tried to clean the logic on the trait a little. Let me know your thoughts and hopefully we can get this in!

wychoong:
> Thanks @wychoong it seems to do the trick, I added a simple test and also tried to clean the logic on the trait a little. Let me know your thoughts and hopefully we can get this in!

looks good to me

glennjacobs (approved):

PR #1945, opened 2024-09-10
Fix missing product variant table

ken717w:


alecritson:
Thanks @ken717w I think in this instance we don't want the table to show since it's managed via a widget so it's likely the whole method can be removed 🤔

ken717w:
> Thanks @ken717w I think in this instance we don't want the table to show since it's managed via a widget so it's likely the whole method can be removed 🤔

So how do we create and manage product variants in the dashboard? Originally I only see this in the Variants tab:
Screenshot 2024-09-10 at 22-39-47 Variants - Lunar

I also tried to create a variant manually in database, and when I go back to the dashboard, tabs like Pricing and Inventory are gone. After some digging I can see that `ManageProductInventory` and `ManageProductPricing` are not being registered to the navigation if there are multiple variants. Although this makes sense, but I can find nowhere to update the price and stock value of each variant.

glennjacobs:
> > Thanks @ken717w I think in this instance we don't want the table to show since it's managed via a widget so it's likely the whole method can be removed 🤔
> 
> So how do we create and manage product variants in the dashboard? Originally I only see this in the Variants tab: Screenshot 2024-09-10 at 22-39-47 Variants - Lunar
> 
> I also tried to create a variant manually in database, and when I go back to the dashboard, tabs like Pricing and Inventory are gone. After some digging I can see that `ManageProductInventory` and `ManageProductPricing` are not being registered to the navigation if there are multiple variants. Although this makes sense, but I can find nowhere to update the price and stock value of each variant.

To create variants you'll want to set the Product Options first, e.g. Colour -> Red, Blue, Green. It will then creste the variants for you, which you can then go on to edit.

PR #1947, opened 2024-09-12
helper function to get model morph name

wychoong:
to simplify the usage of getting relationship alias 

```php
// before
(new Collection)->getMorphClass()

// after
Collection::morphName()
```


alecritson:
I think this is good to have we'd just need to update the existing codebase so it's consistent.

wychoong:
> I think this is good to have we'd just need to update the existing codebase so it's consistent.

done. I only changed those using `(new Model)->getMorphClass`, those already have the instance i remain `$record->getMorphClass()`

added test as well.

alecritson (approved):

PR #1948, opened 2024-09-12
fix price breaks not storing in correct currency factor

wychoong:
also added `->required()` to shipping method


alecritson:
Thanks @wychoong Should we also be considering the Min. Spend field as well?

wychoong:
> Thanks @wychoong Should we also be considering the Min. Spend field as well?

Min. Spend field is actually min_quantity field on the table, and currently it is hardcoded to multiply 100.
do you want me to update it to use currency factor too?

alecritson:
I think so, since the shipping addon uses this as a price check 🤔

wychoong:
> I think so, since the shipping addon uses this as a price check 🤔

I have updated both to use the currency factor. Also fixed my previous change which uses default currency. now it uses the currency selected at the repeater 😶‍🌫️ 
image

glennjacobs (approved):

PR #1956, opened 2024-09-17
Fix line fetching in order creation

maurice-ellis:
I'm updating the line fetching in `Lunar\Pipelines\Order\Creation\CreateOrderLines` and `Lunar\Pipelines\Order\Creation\MapDiscountBreakdown` to match the one used in `Lunar\Actions\Carts\GetExistingCartLine` to see if the line meta matches as well.

This PR will help be able to handle products when similar products are added instead of merged and quantity updated.


alecritson:
Thanks @maurice-ellis Are you able to provide a test case to show what this is fixing within it's context?

maurice-ellis:
@alecritson I added a test case. I believe I did that correctly.

The issue is when you have 2 of the same purchasables, it causes an undefined key error in generating the breakdown because the line does not exist because the `$cartLinesMappedToOrderLines` only had one order line. By comparing the meta as well, it generates 2 separate lines and the error is resolved.

Not sure how to exactly test for that, but did create a case.

glennjacobs (approved):

PR #1961, opened 2024-09-19
Fix/shipping min quantity factor

wychoong:
changes:
- min_quantity now check on shipping method to determine should it divide by currency factor
- min_quantity field label now reflect according to method
- move min_quantity as last field
- ~now assume weight is by KG~

after
image

image


maurice-ellis:
Don't assume weight, we are imperial in the US

wychoong:
> Don't assume weight, we are imperial in the US

the focus of this PR is on the shipping method chargeBy is not considered when converting min_quantity

to properly calculate shipping cost for chargeBy weight requires more effort and should be a separate PR

wychoong:
changed it back to min_quantity / 100 for weight so that it is same as previous behavior

alecritson:
Trying to get through the backlong leading up to v1.0.0. 

It looks like this PR has swapped the order of price/min_spend, is that intentional?

wychoong:
> Trying to get through the backlong leading up to v1.0.0.
> 
> It looks like this PR has swapped the order of price/min_spend, is that intentional?

Yes it is. It is so that when changing shipping method the fields will be in the same order, just the label are updated

wychoong:
Sorry, it’s been too long. I think it’s just that it reads better to me. Can just revert the change if have to

glennjacobs (approved):

PR #1962, opened 2024-09-19
Update Cart.php

wychoong:
currently I'm facing an issue where we use `$cart->getShippingOption()` in a cart calculate pipeline and its causing infinite loop

for what this method concern, it should just need `refresh()` or `load('shippingAddress')` to get latest shipping address and shipping option

### DIFF
--- a/packages/core/src/Models/Cart.php
+++ b/packages/core/src/Models/Cart.php
@@ -506,7 +506,7 @@ class Cart extends BaseModel implements Contracts\Cart
 
     public function getShippingOption(): ?ShippingOption
     {
-        return ShippingManifest::getShippingOption($this->refresh());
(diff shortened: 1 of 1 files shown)


packages/core/src/Models/Cart.php
```diff
@@ -506,7 +506,7 @@ public function setShippingOption(ShippingOption $option, bool $refresh = true):
 
     public function getShippingOption(): ?ShippingOption
     {
-        return ShippingManifest::getShippingOption($this->calculate());
+        return ShippingManifest::getShippingOption($this->refresh());
```
wychoong:
```suggestion
        return ShippingManifest::getShippingOption($this->load('shippingAddress'));
```

PR #1966, opened 2024-09-24
doc: add remark for custom model definition

wychoong:
when a Lunar model is extended/replaced, the media definition config should be updated. Or standard definition will be used

p/s: not sure should the media definition to change to string alias or internally auto resolve to lunar model fqcn

### DIFF
--- a/packages/core/config/media.php
+++ b/packages/core/config/media.php
@@ -4,26 +4,13 @@ use Lunar\Base\StandardMediaDefinitions;
 
 return [
 
-    /*
-    |--------------------------------------------------------------------------
-    | Media Definition
-    |--------------------------------------------------------------------------
-    |
-    | Specify which media definition should be used when generating media.
-    |
-    | note: if extended Lunar's model, you should update the key,
-    |       or default definition will be used
-    |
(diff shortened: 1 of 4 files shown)


alecritson:
I think it would be more beneficial if the media definition worked with extended models. My thinking is that you should be able to just remove model extending in one place without having to then update a bunch of config etc in order to remove it if that makes sense.

wychoong:
> I think it would be more beneficial if the media definition worked with extended models. My thinking is that you should be able to just remove model extending in one place without having to then update a bunch of config etc in order to remove it if that makes sense.

Totally agree. In config you prefer alias or fqcn?

alecritson:
> > I think it would be more beneficial if the media definition worked with extended models. My thinking is that you should be able to just remove model extending in one place without having to then update a bunch of config etc in order to remove it if that makes sense.
> 
> Totally agree. In config you prefer alias or fqcn?

I think alias should be sufficient since that's set up by Lunar so unlikely to cause issues and maybe less confusing for people using extending.

wychoong:
> alias should be sufficient

done

glennjacobs (approved):

PR #1970, opened 2024-09-26
Fix number field setter

### LINKED ISSUE #1969: Inconsistent datatype of number product attributes in database
- Lunar version: 1.0.0-beta.2
- Laravel Version: 11.22.0
- PHP Version: 8.3.11

### Expected Behaviour:
Running SQL query `SELECT JSON_EXTRACT(attribute_data, '$.my_number_field') AS my_number_field FROM lunar_products;` should show `my_number_field` in integers.

### Actual Behaviour:
For new products which I didn't change any of the attributes the above query gives:
```JSON
{
	"value": 0,
	"field_type": "Lunar\\FieldTypes\\Number"
}
```

While if I've changed `my_number_field` to 1, the query gives:
```JSON
{
	"value": "1",
	"field_type": "Lunar\\FieldTypes\\Number"
}
```

If I change it back to 0, the query gives:
```JSON
{
	"value": "0",
	"field_type": "Lunar\\FieldTypes\\Number"
}
```

### Steps To Reproduce:
1. Create a product attribute of type `Lunar\FieldTypes\Number`
2. Create 2 products, and update this attribute in one of them
3. Run the SQL query above

ken717w:
Do data type casting from String to either int or float.

This should fixes #1969.

### DIFF
--- a/packages/core/src/FieldTypes/Number.php
+++ b/packages/core/src/FieldTypes/Number.php
@@ -54,6 +54,12 @@ class Number implements FieldType, JsonSerializable
             throw new FieldTypeException(self::class.' value must be numeric.');
         }
 
+        if (strpos($value, '.') !== false) {
+            $value = floatval($value);
+        } else {
+            $value = intval($value);
+        }
+
         $this->value = $value;
     }
 
--- a/packages/core/src/FieldTypes/Number.php
+++ b/packages/core/src/FieldTypes/Number.php
@@ -54,13 +54,7 @@ class Number implements FieldType, JsonSerializable
             throw new FieldTypeException(self::class.' value must be numeric.');
         }
 
-        if (strpos($value, '.') !== false) {
-            $value = floatval($value);
-        } else {
-            $value = intval($value);
-        }
-
-        $this->value = $value;
+        $this->value = $value + 0;
     }
 
     /**
--- a/packages/core/src/FieldTypes/Number.php
+++ b/packages/core/src/FieldTypes/Number.php
@@ -54,7 +54,7 @@ class Number implements FieldType, JsonSerializable
             throw new FieldTypeException(self::class.' value must be numeric.');
         }
 
-        $this->value = $value + 0;
+        $this->value = $value ? $value + 0 : '';
     }
 
     /**
--- a/packages/core/src/FieldTypes/Number.php
+++ b/packages/core/src/FieldTypes/Number.php
@@ -46,15 +46,21 @@ class Number implements FieldType, JsonSerializable
     /**
      * Set the value of this field.
      *
-     * @param  int|float  $value
+     * @param  string|int|float|null  $value
      */
     public function setValue($value)
     {
-        if ((! is_numeric($value)) && $value != '') {
+        if (blank($value)) {
+            $this->value = 0;
+            return;
+        }
+
+        if (!is_numeric($value)) {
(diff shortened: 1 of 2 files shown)


packages/core/src/FieldTypes/Number.php
```diff
@@ -54,6 +54,12 @@ public function setValue($value)
             throw new FieldTypeException(self::class.' value must be numeric.');
         }
 
+        if (strpos($value, '.') !== false) {
```
alecritson: Instead of this if/else statement we could also do something like this:

```php
$this-value = $value + 0;

"1.29" + 0 = 1.29
"129" + 0 = 129
```

ken717w: Agree, that's simpler.

ken717w: @alecritson Just committed the change.

alecritson (dismissed):

ken717w:
The tests failed, are there some sort of linting rules?

alecritson:
> The tests failed, are there some sort of linting rules?

Spotted that, pushed a quick patch which should sort it, it was due to `$value` being an empty string. Although not sure if that will work if you pass a `0` thinking about it 😱

packages/core/src/FieldTypes/Number.php
```diff
@@ -54,7 +54,7 @@ public function setValue($value)
             throw new FieldTypeException(self::class.' value must be numeric.');
         }
 
-        $this->value = $value;
+        $this->value = $value ? $value + 0 : '';
```
wychoong: does returning '' solve the initial issue?

will setting type hint to property/return/argument solve the issue?

glennjacobs: @alecritson @wychoong 

Would this approach maybe be better?

```php
/**
 * Set the value of this field.
 *
 * @param  string|int|float|null  $value
 */
public function setValue($value)
{
    if ($value === '' || $value === null) {
        $this->value = 0;
        return;
    }

    if (!is_numeric($value)) {
        throw new FieldTypeException(self::class.' value must be numeric.');
    }

    // Cast to int if the value is a whole number, otherwise use float
    $this->value = floor($value) == $value ? (int) $value : (float) $value;
}

```

wychoong: Would you be ok to use blank() instead to handle for “ “

glennjacobs: Actually, yes, blank() would work. Good idea.

packages/core/src/FieldTypes/Number.php
```diff
@@ -46,15 +46,21 @@ public function getValue()
     /**
      * Set the value of this field.
      *
-     * @param  int|float  $value
+     * @param  string|int|float|null  $value
      */
     public function setValue($value)
     {
-        if ((! is_numeric($value)) && $value != '') {
+        if (blank($value)) {
```
alecritson: Would something like this be a bit better? Prevents us having to having double `if` statements, having to do a `return` as an escape and also means the property assignment is in one place.

```php
if (! blank($value) && ! is_numeric($value)) {
    throw new FieldTypeException(self::class.' value must be numeric.');
}

// Cast to int if the value is a whole number, otherwise use float
$this->value = blank($value) ? 0 : (floor($value) == $value ? (int) $value : (float) $value);
```

glennjacobs (approved):

PR #1975, opened 2024-10-02
Fix exception when updating order line

lguichard:
After the Model Extending change, we have missing that the order line observer trigger an exception with the morph name.

``
class_implements(): Class product_variant does not exist and could not be loaded
``


alecritson:
Thanks @lguichard although this is a duplicate of #1964

PR #1979, opened 2024-10-05
fix: delete child collection item

ferdiunal:
It gives an error because the collection to be deleted is linked to the ‘Customer Group’ table and the foreign key index.

image


alecritson:
Thanks @ferdiunal Can you provide some steps on how to reproduce as I currently can't seem to :)

rjvandoesburg:
I can reproduce the issue simply by creating a new root collection and then adding a single collection to that root.
When i try to delete the collection (not the root) I get the same error.

I have to manually delete the record from `lunar_collection_customer_group` before i can delete the collection via the admin.

This should be solvable by adding `cascadeOnDelete()` on the constraints, this also allows developers to delete records outside of Lunar.

glennjacobs:
An issue still resides if you try to delete a parent collection, as it won't remove the customer group relations for the child collections.

I'm not too sure why this is still an issue, as we have the detach in the observer, see here
https://github.com/lunarphp/lunar/blob/1.x/packages/core/src/Observers/CollectionObserver.php

glennjacobs:
Thanks for the submission, but I think this is the wrong approach. I will open a separate PR to resolve this, after I've consulted @alecritson

PR #1986, opened 2024-10-09
Make Product name and SKU searchable

ryanmitchell:
If you use the database search driver (ie dont use scout) in the admin the product table is not searchable by SKU or by product name, which are the two ways you are most likely to search this table.

This PR updates that.


glennjacobs:
@alecritson I think this looks OK, but figured I'd get you to double-check :)

alecritson (approved):

Issue #1996, opened 2024-10-20
Product variant attributes can not be edited if product has no options


AidasK:
- Lunar version: 1.0.0-beta3
- Laravel Version: 10.48.22
- PHP Version: 8.3.9
- Database Driver & Version: Mysql

### Expected Behaviour:
I should be able to fill in product variant attribute data event though product has only one product variant (e.g. it has no options)

### Actual Behaviour:
I am not able to fill in any product variant attributes.

### Steps To Reproduce:
1. Assign some custom attributes to product variant.
2. Create a product and you won't be able to fill in any custom attribute_data on product_variant.

On lunar 0.8 it was possible to fill in this data, not we can't. Solution might be to create same attributes for the product too, but it create additional code complexity as you must check two places for attributes in a product and in a product variant.

glennjacobs:
That's a good point, it should support that.

fredmarmillod:
I've just tried on a fresh Lunar starterkit install and it works with Text or TranslatedText field types but others like Toggle or Number triggers a validation error. I'm using **v1.0.0-beta17**

<img width="932" alt="Image" src="https://github.com/user-attachments/assets/e9e2811f-7390-426b-9352-cbfa1879a77d" />

Steps to reproduce:

1. Install Lunar starterkit
2. Create a attribute group
3. Add a Number field to the group
4. Add the group to a product type
5. Create a new product
6. Try to save the product

PR #1997, opened 2024-10-20
Fix/product variant attribute filling

AidasK:
Not expecting this to be merged, but this is just for an idea how it can be solved https://github.com/lunarphp/lunar/issues/1996


glennjacobs:
I've converted this to a draft as it's not ready yet. We do need a solution for this. Would you like to continue the development?

alecritson:
We've since updated how variant attributes are managed, if this is still deemed relevant feel free to reopen.

PR #1998, opened 2024-10-22
Fix/remove cart customer check exception

AidasK:
If anonymous customer creates a cart and then tries to login it it throws an exception and page crashes. From database perspective customer can have multiple users so this check should not be performed at all.


glennjacobs:
If a user logs in, Lunar assumes they are associated to a customer and if not throws an exception because a user must belong to a customer. 

We don't support users placing orders without being associated to a customer.

I'm going to close this issue, but if you think this is wrong or think it could be improved, please let me know.

PR #2004, opened 2024-10-25
Fix an issue where the AmountOff percentage is cast to an int, even t…

### LINKED ISSUE #2003: AmountOff percentage is cast to an int, even though it could contain a decimal value
The admin panel allows you to enter a decimal value when creating an AmountOff Discount with a percentage discount. However, it gets cast to an int in the AmountOff class, which results in the percentage being rounded and the discount amount being slightly off.

You could argue that usually percentage discounts are given as whole numbers, but since the admin panel already supports decimal values, I think this should be changed.

I'll include a PR to address this.

tvlokven:
The admin panel allows you to enter a decimal value when creating an AmountOff Discount with a percentage discount. However, it gets cast to an int in the AmountOff class, which results in the percentage being rounded and the discount amount being slightly off.

You could argue that usually percentage discounts are given as whole numbers, but since the admin panel already supports decimal values, I think this should be changed.

Fixes #2003

### DIFF
--- a/packages/admin/src/Filament/Resources/CustomerResource/RelationManagers/OrdersRelationManager.php
+++ b/packages/admin/src/Filament/Resources/CustomerResource/RelationManagers/OrdersRelationManager.php
@@ -5,8 +5,8 @@ namespace Lunar\Admin\Filament\Resources\CustomerResource\RelationManagers;
 use Filament\Tables;
 use Filament\Tables\Table;
 use Lunar\Admin\Filament\Resources\OrderResource;
-use Lunar\Admin\Support\RelationManagers\BaseRelationManager;
 use Lunar\Admin\Filament\Resources\OrderResource\Pages\ManageOrder;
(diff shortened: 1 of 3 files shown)


alecritson:
Thanks @tvlokven I think this is valid, although I'm curious why you overrode the 10% discount test instead of adding to it? Maybe this is a good use case for Pest datasets? https://pestphp.com/docs/datasets

tvlokven:
@alecritson thanks for having a look. I thought that having both a test for 10.5% and 10% would be redundant, I didn't want to copy-paste the entire test with some different values. Your suggestion of using a Pest dataset sounds like a good idea, I've added a new commit where I did this. I was struggling a little bit with how best to pass the parameters to the test (using arrays, making quantities dynamic maybe instead of hardcoding 1 and 2, etc). But in the end I stuck with the easiest approach, which I thought was the most clear.

PR #2008, opened 2024-11-01
Adds attributes support for the Price Model

abishekrsrikaanth:
This PR adds attribute data to the `Price` model.

This will be useful for storing price related meta data for conditions such as title and description related to pricing. This is in relation to a discussion #2007 to allow adding titles and descriptions for Shipping rates and Rate breaks. 

Since the column on the database is nullable, it should not break or affect any existing UI. The attributes related to pricing can be added on a need basis as mentioned in the Use case on the discussion here: #2007 

A PR is in the works


glennjacobs:
I don't believe this is the solution we're looking for. Using attributes would be very difficult to implement in the admin UI. 

Let's continue the discussions on the Feature Request discussion.

PR #2009, opened 2024-11-04
Add Spanish localization files

dualklip:
This commit introduces new Spanish localization files for various components within the admin and shipping packages. These include translation strings for customer groups, product variants, shipping methods, and more, etc...


glennjacobs:
Many thanks. I don't speak Spanish, so I'll rely on the community to keep this updated.

glennjacobs (approved):

PR #2011, opened 2024-11-05
Adds KeyValue to OrderItemsTable for lunar_order_lines meta data

shanerbaner82:
If adding meta data to a line item there is currently no visibility on the backend for the order line item meta data.

For example, we run a custom printing shop and want to allow users to enter custom data for custom prints:

line_1: USDOT 1234567
line_2: GVW 12334566
...

We need a way to see this data after the order is placed without having to go into the db.


glennjacobs:
Things I'm thinking here is, ...

- some projects may not wish to show the meta information, maybe we should make it optional somehow?
- If a meta entry has no value, do we still want to show it?

glennjacobs:
@alecritson do we want to take over this PR?

alecritson:
> @alecritson do we want to take over this PR?

Happy for us to.

PR #2025, opened 2024-11-28
Fix deleting a discount should delete relations

ryanmitchell:
Deleting a discount should delete its pivots (channels, customerGroups and brands) otherwise you run into foreign key exceptions.


alecritson:
Should we also be considering other relationships, such as users and purchasables,

ryanmitchell:
@alecritson done!

alecritson (approved):

PR #2046, opened 2025-01-01
Fixes customer group in Discounts

### LINKED ISSUE #2045: Customer groups of cart customer are not considered by Discounts
- Lunar version: 1.0.0-beta.4
- Laravel Version: 11.31.0
- PHP Version: 8.2.25
- Database Driver & Version: 10.6.20-MariaDB

### Expected Behaviour:
Discount available to a specific customer group should be applied

### Actual Behaviour:
Discount available to a specific customer group is not applied

### Steps To Reproduce:
- Add a discount to some products
- Make it available to a specific customer group only
- Login with a user of that specific customer group 
- Add a product from discounted to cart 
- Discount is not applied

yasirmturk:
Fixes https://github.com/lunarphp/lunar/issues/2045

### DIFF
--- a/packages/core/src/Managers/DiscountManager.php
+++ b/packages/core/src/Managers/DiscountManager.php
@@ -123,7 +123,7 @@ class DiscountManager implements DiscountManagerInterface
         }
 
         if ($this->customerGroups->isEmpty()) {
-            if ($customerGroups = $cart->customer?->customerGroups) {
+            if ($cart && $customerGroups = $cart->customer?->customerGroups) {
                 $this->customerGroup($customerGroups);
             } elseif ($defaultGroup = CustomerGroup::getDefault()) {
                 $this->customerGroup($defaultGroup);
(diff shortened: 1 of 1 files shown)


alecritson (changes requested):
Thanks @yasirmturk I think we should try to avoid nested if statements and `elseif`'s

yasirmturk:
I agree @alecritson , what do you suggest? conditional or null colescing statement?

alecritson:
> I agree @alecritson , what do you suggest? conditional or null colescing statement?

I kind of think that we should always consider the customer's groups, regardless, so maybe something like this:

```php
if ($cart && $customerGroups = $cart->customer?->customerGroups) {
    $this->customerGroup($customerGroups);
}

if ($this->customerGroups->isEmpty() && $defaultGroup = CustomerGroup::getDefault()) {
    $this->customerGroup($defaultGroup);
}
```

What do you think?

alecritson (approved):

Issue #1992, opened 2024-10-18
ListField bug, does not allow 0 as a key


AidasK:
- Lunar version: 1.0.0-beta.3
- Laravel Version: 10
- PHP Version: 8.3
- Database Driver & Version: Mysql

### Expected Behaviour:
It should be allowed to start a list with a key 0 as in regular array

### Actual Behaviour:
It deletes all the values if there is a key 0 and also deletes all the duplicate fields if they have the same key.

https://github.com/user-attachments/assets/b1e4bd3c-e37c-4cf6-82b5-372225b5218c

### Steps To Reproduce:
Create an attribute with ListField type and fill it with key 0

shepard153:
Same thing on beta.6. In my opinion this should be KeyValue field not ListField. List field should be separate component with list of inputs

PR #2052, opened 2025-01-02
Remove live method from ListField

alecritson:
Having the `->live()` method on the `ListField` component was causing issues when trying to interact with the values. There is no good reason to use the live method on this field and removing it seems to fix the problem outlined in #1992


wychoong:
can you test if the issue in #1787 is solved with latest filament/alpinejs after this change?

glennjacobs:
> can you test if the issue in #1787 is solved with latest filament/alpinejs after this change?

Seems fine, so will merge.

glennjacobs (approved):

PR #2053, opened 2025-01-02
fix compare price decimal

wychoong:


alecritson:
Thanks @wychoong Can you provide either a test case or instructions on how to reproduce the issue this is fixing? thanks.

wychoong:
> Thanks @wychoong Can you provide either a test case or instructions on how to reproduce the issue this is fixing? thanks.

Before fix, when saving value, eg: 1000, it will display as 10.00 after refresh, as it will save as it is but display divide by factor. It should have same handling as `price` field

alecritson (approved):

PR #2064, opened 2025-01-16
Bring search package into monorepo

alecritson:
This PR looks to introduce the new search package into the monorepo


wychoong:
i know this might be early. but can it play nicely with model extension?

alecritson:
> i know this might be early. but can it play nicely with model extension?

So the package doesn't return Product models, just what is from the index. I have been using it on a site with model extending and the only caveat for now is you may have to add your own model references to the search config alongside `Lunar\Models\{class}` e.g.

```php
'facets' => [
    \Lunar\Models\Product::class => [],
    \App\Models\Product::class => [],
]
```

wychoong:
I think it should be handled like the indexer config. I’m not too sure if there is a use case for having both references 🤔

alecritson:
> I think it should be handled like the indexer config. I’m not too sure if there is a use case for having both references 🤔

There's some area's in scout where the config is referenced directly so not sure how that could be overcome at the moment.

glennjacobs (approved):

PR #2072, opened 2025-01-21
Addition of getCartQuantity method on the CartSessionManager

chrispage1:
This PR contains two updates to the `CartSessionManager` class

1) Abstraction of cart ID retrieval so anyone can get the current cart ID

2) Addition of a `getCartQuantity` method which will allow fetching the current quantity of all line items with minimal queries.

3) Addition of `onlyVisible` to the customerGroups scope to be able to limit products down based on visibility

The reason for this addition is that to load the basket quantity at the top of the page uses lots more queries than is necessary, if I am indeed simply fetching the basket quantity.

I have also added a test for this.

Thanks!


glennjacobs:
Thanks @chrispage1, could we add something to the docs to cover these new methods?

packages/core/src/Managers/CartSessionManager.php
```diff
@@ -137,6 +140,31 @@ private function fetchOrCreate(bool $create = false, bool $estimateShipping = fa
...
+     */
+    public function getCartQuantity(): int
+    {
+        if (! $cartId = $this->getCartId()) {
+            return 0;
+        }
+
+        return CartLine::query()
+            ->where('cart_id', $cartId)
+            ->where('purchasable_type', ModelManifest::getMorphMapKey(ProductVariant::class))
```
glennjacobs, line 157: We can't assume that products will always be of type ProductVariant. 

Why do we care about the purchasable type when getting the total cart quantity?

glennjacobs: @alecritson do you see any reason for filtering when getting the cart quantity?

alecritson: I guess yeah if it's purchasable then should be considered in the count.

chrispage1:
Hey, I've closed this just because I opted for a bespoke system in the end rather than Lunar and can't really remember the exact reasoning behind the PR 😄

PR #2075, opened 2025-01-22
Fix forceDelete product

wychoong:


packages/admin/src/Support/Actions/Products/ForceDeleteProductAction.php
```diff
@@ -20,24 +20,16 @@ public function setUp(): void
...
-                    $record->variants()->delete();
+                    $record->variants()->get()->each->forceDelete();
 
                     $record->productOptions()->detach();
 
                     $record->associations()->delete();
 
                     $record->channels()->detach();
 
-                    $record->clearMediaCollection('images');
```
wychoong, line 39: this is unnecessary as handled by Spatie's Media observer for cleanup

packages/admin/src/Support/Actions/Products/ForceDeleteProductAction.php
```diff
@@ -20,24 +20,16 @@ public function setUp(): void
...
-                    $record->variants()->each(function (Model $variant) {
-                        $variant->values()->detach();
-
-                        $variant->prices()->delete();
-                    });
-
                     $record->urls()->delete();
 
-                    $record->variants()->delete();
+                    $record->variants()->get()->each->forceDelete();
```
wychoong: cleanup to use ProductVariant observer for handling relationship

packages/admin/src/Support/Actions/Products/ForceDeleteProductAction.php
```diff
@@ -20,24 +20,16 @@ public function setUp(): void
...
 
                     $record->productOptions()->detach();
 
                     $record->associations()->delete();
 
                     $record->channels()->detach();
 
-                    $record->clearMediaCollection('images');
-
                     return $record->forceDelete();
```
wychoong: the whole code could be simplified to just this one line if we register a Product observer for handling the relationships.

user land can register own observer as well, i think 🤔

packages/core/src/Observers/MediaObserver.php
```diff
@@ -27,7 +27,11 @@ protected function ensureOnlyOnePrimary(Media $media, bool $isDelete = false): v
             return;
         }
 
-        $owner = $media->model()->sole();
+        $owner = $media->model()->first();
+
+        if (blank($owner)) {
+            return;
+        }
```
wychoong, line 34: this solves an issue where when force deleting a product with media.

using `sole` doesn't trigger 'ModelNotFoundException' when deleting media in ManageProductMedia RelationManager, thus it was not caught last time. However it triggers when force delete product. This is likely due to the sequence of triggering the observer deleting/deleted.

packages/admin/src/Support/Actions/Products/ForceDeleteProductAction.php
wychoong: @alecritson need your input on this, not sure is ProductObserver is purposely left out

alecritson: @wychoong Yeah I think ultimately having this kind of logic in the panel is a mistake, we should really be using observers, whether within the model itself or in a separate class.

wychoong: @alecritson Alright, got it. 
Can you review the part for media observer? Will prepare a separate pr for the product observer

packages/core/src/Observers/ProductObserver.php
```diff
@@ -11,11 +11,29 @@ class ProductObserver
...
+
+            $product->productOptions()->detach();
+
+            $product->associations()->delete();
+
+            $product->channels()->detach();
+
+            $product->tags()->detach();
+        } else {
+            $product->variants()->get()->each->delete();
```
alecritson, line 31: I appreciate that the model events need to be triggered, what happens however if there is a large amount of variants 🤔 We currently do not enforce a maximum permutations so just thinking of the overhead.

alecritson:
Thanks @wychoong We had a small discussion about soft deletes internall recently so will bring @glennjacobs on this to get his views once he is back from leave.

packages/core/src/Observers/ProductObserver.php
wychoong: maybe can dispatch into queue to handle. if not dispatching the events, will have to duplicate the codes for handling models related to prodduct variants

alecritson: Yeah, part of me thinks it might be useful to create actions which handle the clean up which can then be put on the job queue. It would keep it out of the observers that way. Bit out of the scope of this PR though.

glennjacobs: For the moment this seems to work and I think there would have to be a crazy amount of variants for it to fail. Sure it might take a few secs, but that's to be expected if you've got like 2,000 variants!

PR #2077, opened 2025-01-23
Added the ability to pass the channel to the StorefrontSessionManager

chrispage1:
Currently, when the StorefrontSessionManager is initialised, there is no way of setting the defaults.

In my use case, I'm determining the channel based on the URL structure and market that is defined (separate to Lunar).

So in my middleware, I want to initialise the storefront with the channel but there isn't actually a way to do this. This adds the option to supply a channel

### DIFF
--- a/packages/core/src/Managers/StorefrontSessionManager.php
+++ b/packages/core/src/Managers/StorefrontSessionManager.php
@@ -36,16 +36,17 @@ class StorefrontSessionManager implements StorefrontSessionInterface
 
     /**
      * Initialise the manager
-     *
-     * @param protected SessionManager
      */
     public function __construct(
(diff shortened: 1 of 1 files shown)


alecritson:
Thanks @chrispage1 I don't imagine this would be required since you should be able to do:

```php
\Lunar\Facades\StorefrontSession::setChannel('webstore');
\Lunar\Facades\StorefrontSession::setChannel(
    \Lunar\Models\Channel::first()
);
```

or if you're worried about the default channel being set beforehand you can manually set the channel in the session

```php
Session::put(\Lunar\Facades\StorefrontSession::getSessionKey().'_channel', 'webstore');
```

PR #2080, opened 2025-01-23
Use correct base price in PricingResponse

se09deluca:
Base price shouldn't be customerGroup scoped, so use $basePrice where customer_group_id is false.

The basePrice is initially retrieved correctly:
https://github.com/lunarphp/lunar/blob/467b49765d827ce937c7ce91043147195546ec59/packages/core/src/Managers/PricingManager.php#L198

then is not used in PricingResponse, instead is newly retrieved loosing the condition that customer_group_id must be false.
https://github.com/lunarphp/lunar/blob/467b49765d827ce937c7ce91043147195546ec59/packages/core/src/Managers/PricingManager.php#L220

That makes potentially get the wrong price as basePrice

### DIFF
--- a/packages/core/src/Managers/PricingManager.php
+++ b/packages/core/src/Managers/PricingManager.php
@@ -217,7 +217,7 @@ class PricingManager implements PricingManagerInterface
 
         $this->pricing = new PricingResponse(
             matched: $matched,
-            base: $prices->first(fn ($price) => $price->tier == 1),
+            base: $basePrice,
             tiered: $prices->filter(fn ($price) => $price->tier > 1),
             customerGroupPrices: $prices->filter(fn ($price) => (bool) $price->customer_group_id)
         );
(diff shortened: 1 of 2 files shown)


se09deluca:
I think the issue is present even in the newest 1.x branch. If you'll accept this PR, i'll create a new on based on that branch.

alecritson:
@se09deluca Thanks for the fix, so is this PR valid or do you need to resubmit?

se09deluca:
> @se09deluca Thanks for the fix, so is this PR valid or do you need to resubmit?

@alecritson 
Yes, from my point of view this PR is valid for the 0.8x branch. I experienced a bug in an ecommerce based on that version of lunar, so this would fix it in 0.8x branch. Later, if this PR will be accepted, i will create a new similar PR for 1.x branch.

It's ok ?

glennjacobs:
@se09deluca as this is making a change to pricing, could we add a test please?

se09deluca:
> @se09deluca as this is making a change to pricing, could we add a test please?

I added a test. I'm not sure that's the best way to do it. 
In the first part it checks that the base price is correct, in the second part it simulates the price fetching as the PricingManager would do, then from the price collection it takes the price with the old method and demonstrates that the value is different from the base price.

alecritson (approved):

PR #2089, opened 2025-01-31
Fix variant switcher when overriding ProductVariant model

rjvandoesburg:
Hi there!

In our application we extend the ProductVariant model like so
```
ModelManifest::replace(
    \Lunar\Models\Contracts\ProductVariant::class,
   App\Models\ProductVariant::class,
);

```

When trying to switch using the button on the top right when editing a variant:

Screenshot 2025-01-31 at 10 54 39

The following error is generated.

Screenshot 2025-01-31 at 10 52 19

After some digging I found that in the closure there was a hard reference to the original model, this caused the evaluation in [getRecordUrl](https://github.com/filamentphp/filament/blob/3.x/packages/tables/src/Table/Concerns/HasRecordUrl.php#L29) to not fill the variable properly. Switching to `record` makes it work even when extending

### DIFF
--- a/packages/admin/src/Filament/Widgets/Products/VariantSwitcherTable.php
+++ b/packages/admin/src/Filament/Widgets/Products/VariantSwitcherTable.php
@@ -88,9 +88,9 @@ class VariantSwitcherTable extends TableWidget
 
     protected function getTableRecordUrlUsing(): ?Closure
     {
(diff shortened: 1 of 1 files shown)


alecritson:
Thanks! @rjvandoesburg 🙇

alecritson (approved):

PR #2092, opened 2025-02-06
Allow variant media page to be extended

ryanmitchell:
The variant media page currently can't be extended as it overrides the form() method, this PR enables it by using the correct getDefaultForm method.


alecritson:
@ryanmitchell Looks like the other PR has introduced some conflicts :)

ryanmitchell:
It can be closed as the other PR allows it to be extended.

PR #2093, opened 2025-02-07
Update variant media page to allow multiple images

ryanmitchell:
This PR updates the variant media screens to allow multiple images to be attached to the product, rather than just a primary image. This aligns with the behaviour before Lunar 1.x.

It now looks more like this:
CleanShot 2025-02-07 at 07 45 02@2x

As this is now a relation manager we've lost the shout boxes you had in place, if there's a way to still have those I think it would be useful. Let me know if I've missed how this could be done.

### DIFF
--- a/packages/admin/resources/lang/en/relationmanagers.php
+++ b/packages/admin/resources/lang/en/relationmanagers.php
@@ -95,6 +95,9 @@ return [
             'create' => [
                 'label' => 'Create Media',
             ],
+            'detach' => [
+                'label' => 'Detach',
+            ],
             'view' => [
                 'label' => 'View',
             ],
--- a/packages/admin/resources/lang/en/relationmanagers.php
+++ b/packages/admin/resources/lang/en/relationmanagers.php
@@ -89,6 +89,9 @@ return [
         'title' => 'Media',
         'title_plural' => 'Media',
         'actions' => [
+            'attach' => [
+                'label' => 'Attach Media',
+            ],
             'create' => [
                 'label' => 'Create Media',
             ],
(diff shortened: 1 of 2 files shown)


alecritson:
Thanks @ryanmitchell Looking good, I don't think there is a way to add the shout notices to the relation manager, however we can always add them to the attach media form? Also just a couple of points:

1 - Breadcrumbs seem to be incorrect now, we can't get back to the product properly
2 - There is no way (that I can see) to detach media
3 - If there is no `small` conversion, we don't display anything, I think we should fallback to `getUrl`, even if it's not optimum it will at least show something.

I was curious also if Filament had a dropdown component which could include images, however it doesn't seem the case...

ryanmitchell:
Thanks.

1. Brought back breadcrumbs
2. Added a detach row action
3. Fallback to getUrl()

Yeah I had wanted the dropdown to be a thumbnail selection but I couldn't see a way to do it unfortunately.

alecritson:
What do you think to adding the Shout components to this form?

image

ryanmitchell:
Sure, easily done. Those are back with the same logic as before.

alecritson:
I don't think it's quite there. I'm not sure if we need the message for now having any images attached, since it's clear to see on the relation manager. I think we maybe just need a message for when the dropdown is empty, so say all available media has been selected perhaps. Also adding the section has introduced a double panel situation 

image

ryanmitchell:
Makes sense. Section removed; shout content and logic updated.

alecritson:
Thanks @ryanmitchell I tweaked the logic a little, I think this makes sense, how about you?

ryanmitchell:
Looks good thank you.

alecritson (approved):

PR #2101, opened 2025-02-12
Don't force lowercase search term for `generate_search_term_expression`

### LINKED ISSUE #2094: When trying to create a Product Association, the search engine does not find any products.
- Lunar version: 1.0.0-beta.9
- Laravel Version: 11.41.3
- PHP Version: 8.4.3
- Database Driver & Version: PostgreSQL 17 (Also tested with MySQL 8.0.33)

### Expected Behaviour:
In the Filament panel, when creating a `Product Association` with a previously selected product, in the product drop-down list you should do a search, regardless of case and find existing products.

### Actual Behaviour:
when I try to create a `Product Association`, when searching for a product to relate to this one, in the Select component called 'product', it is not able to find anything, it makes no difference to me if I search with all lowercase or by writing the name of a product exactly the same, no product is found.

I actually tested this without using Laravel Scout.

### Steps to reproduce:
1. In the `config/lunar/panel.php` file, set the 'scout_enabled' parameter to false.
2. Create 2 products.
3. In one of them, edit and try to create a `Product Association` by searching in the Select component's search box by typing the name of your other product.

alecritson:
Currently when scout is disabled, the search term is being forced to lowercase, meaning when doing a `where` query in the database with `LIKE` it's not matching any results.

This PR stops this behaviour so products can be matched.

Fixes #2094

### DIFF
--- a/packages/admin/src/helpers.php
+++ b/packages/admin/src/helpers.php
@@ -91,7 +91,7 @@ if (! function_exists('get_search_builder')) {
             /** @var Connection $databaseConnection */
             $databaseConnection = $query->getConnection();
 
-            $search = generate_search_term_expression($search, true, $databaseConnection);
+            $search = generate_search_term_expression($search, false, $databaseConnection);
 
             foreach (explode(' ', $search) as $searchWord) {
                 $query->where(function (Builder $query) use ($model, $searchWord) {


glennjacobs:
As MySQL LIKE is case-insensitive, how does it fix it?

glennjacobs:
I see that when using JSON it becomes a binary string, that IS case-sensitive.

glennjacobs:
I'm not sure this is the correct solution, as we don't want a case-sensitive search. 

I would imagine we want to set both the field being searched upon and the keyword to be lowercase, to achieve a case-insensitive search. 

So we'd retain the lowercasing of the keyword, but update the query to also lowercase the field in the LIKE comparison.

By using `LOWER()` in the query we can achieve this, and a bonus is that it's compatible with both MySQL and PostgreSQL.

glennjacobs:
Closing in favour of new PR

Issue #2071, opened 2025-01-21
Can not implement a Filament Repeater as attribute field


tdwesten:
- Lunar version: v1.0.0-beta.9
- Laravel Version: v11
- PHP Version: 8.3.15
- Database Driver & Version: Mysql 

### Expected Behaviour:
I can use the Filament Repeater field to create more complex data structures based on a array of objects.

### Actual Behaviour:

When I create a custom attribute field using the Filament Repeater field, which extends from the Lunar List field, it doesn't render existing data and throws this error:
```
Filament\Forms\ComponentContainer::getRawState(): Return value must be of type Illuminate\Contracts\Support\Arrayable|array, stdClass returned
```

### Steps To Reproduce:

1. Create a new field: 
```php
<?php

namespace App\Domain\Lunar\Fields;

use Lunar\FieldTypes\ListField;

class ProductDimensionsField extends ListField {}
```

2. Create a field type:
```php
<?php

namespace App\Domain\Lunar\FieldTypes;

use App\Domain\Lunar\Synthesizers\ProductDimensionsFieldSynth;
use Filament\Forms\Components\Component;
use Filament\Forms\Components\Repeater;
use Filament\Forms\Components\TextInput;
use Lunar\Admin\Support\FieldTypes\ListField;
use Lunar\Models\Attribute;

class ProductDimensionsFieldType extends ListField
{
    /**
     * Specify the Synthesizer for this field type
     */
    protected static string $synthesizer = ProductDimensionsFieldSynth::class;

    /**
     * Define the Filament component for rendering
     */
    public static function getFilamentComponent(Attribute $attribute): Component
    {
        return Repeater::make($attribute->handle)
            ->collapsed()
            ->itemLabel(fn (array $state): ?string => self::getLabel($state))
            ->columns(3)
            ->schema([
                TextInput::make('length')
                    ->numeric()
                    ->label('Length')
                    ->required(),
                TextInput::make('depth')
                    ->numeric()
                    ->label('Depth')
                    ->required(),
                TextInput::make('height')
                    ->numeric()
                    ->label('Height')
                    ->required(),
                TextInput::make('label')
                    ->label('Label')
                    ->required(),
            ])
            ->addable()
            ->deletable()
            ->reorderable();
    }

    /**
     * Get the label for the repeater item
     */
    protected static function getLabel(array $state): ?string
    {
        return sprintf(
            '%s | Price: %s - Updated: %s',
            $state['label'],
[...]

PR #2105, opened 2025-02-14
Feat: Generalize attribute data field state handling in Filament

### LINKED ISSUE #2091: Custom Attribute File Field Type Fails to Save Due to Strict Type Comparison in AttributeData
- Lunar version: 1.0.0-beta.6
- Laravel Version: 11.35.1
- PHP Version: 8.3.15

### Expected Behaviour:

When creating a custom attribute file field type by following the documentation, the file should be correctly saved, and its reference should be stored in the database.

### Actual Behaviour:

When attempting to save a custom attribute file field type, an error occurs. The issue seems to stem from the AttributeData class, specifically in the ->mutateDehydratedStateUsing method. There is an if condition that checks:
`if ($attribute->type == FileFieldType::class)`
Since a custom attribute file field type is not strictly equal to FileFieldType::class, this condition does not execute.

### Steps To Reproduce:

1. Follow the Lunar documentation to create a custom attribute field type for a file upload.
2. Attempt to save a record with this custom file field type.
3. Observe that an error occurs because the if condition does not account for custom file field types.

repl6669:
## Description

This PR aims to generalize how attribute data fields are handled in Filament and fix #2091 and #2071.

1. Component state for all fields is set to field value by using the `->getValue()` method.
2. Instance of the field with its value set is returned after dehydration

## Todo
- [x] Test that fields are rendered correctly on `ProductEditPage`
- [x] Test that the form is saved correctly for all Lunar fields as well as custom, user defined fields
    - [x] Test custom `RepeaterField`
    - [x] Test custom `BuilderField`


alecritson:
@repl6669 Is this still being worked on?

repl6669:
Hello @alecritson, I will give it some more time this weekend. I just need to find out how to properly test it.

glennjacobs (changes requested):
@repl6669 From my QA this appears to break the file upload field type.

image

And the ListField no longer saves its values.

PR #2110, opened 2025-02-14
Fix error in collection deletion

glennjacobs:
When deleting a collection that has child collections in the admin panel, it will error. I believe this is because the child-collections attempt to be removed by nestedset, but their associated customer groups etc are not being removed.

The simplest and safest approach might be to simply not allow the deletion of collections if they have child collections.

This PR introduces a quick solution for that, but there may be a nicer way.


wychoong:
> this is because the child-collections attempt to be removed by nestedset, but their associated customer groups etc are not being removed.

getting this error when try to delete a collection with childs
```bash
> Collection::whereIn('id', [21])->get()->each->delete()

   Illuminate\Database\QueryException  SQLSTATE[23000]: Integrity constraint violation: 14
51 Cannot delete or update a parent row: a foreign key constraint fails (`ems_klef`.`lunar
_collection_customer_group`, CONSTRAINT `lunar_collection_customer_group_collection_id_for
eign` FOREIGN KEY (`collection_id`) REFERENCES `lunar_collections` (`id`)) (Connection: ma
riadb, SQL: delete from `lunar_collections` where (`lunar_collections`.`collection_group_i
d` = 1 and `lunar_collections`.`_lft` between 67 and 71) and `lunar_collections`.`collecti
on_group_id` = 1).
```

since now we already have Collection observer to cleanup the related customerGroups, discounts, etc.
we might just need to override the `deleteDescendants` method to `$this->descendants->each->delete()` the child model instead of delete by query `$this->descendants()->delete()`

https://github.com/lazychaser/laravel-nestedset/blob/fa3f77bb96709eb6b394f66b54eb08f0d8d47d63/src/NodeTrait.php#L622-L643

vitalijalbu:
I also have this issue, any fixes?? Thanks

alecritson (dismissed):
I think this is fine as a failsafe, just need to sort out translations.

alecritson (approved):

PR #2124, opened 2025-02-26
Laravel 12 Support

### LINKED ISSUE #2121: Add support for Laravel 12
- Lunar version: latest
- Laravel Version: 12.x
- PHP Version: 8.4.0

### Expected Behaviour:
Allow to be installed in a laravel 12 project.

### Actual Behaviour:
  Problem 2
    - Root composer.json requires lunarphp/core * -> satisfiable by lunarphp/core[dev-main].
    - lunarphp/core dev-main requires laravel/framework ^10.0|^11.0 -> found laravel/framework[v10.0.0, ..., 10.x-dev, v11.0.0, ..., 11.x-dev] but it conflicts with your root composer.json require (^12.0).

### Steps To Reproduce:
Fresh laravel 12 project and try to install this package.

alecritson:
Fixes #2121 

- Introduces support for Laravel 12.
- Adds PHP 8.4 to automated tests.
- Move filament-input-select-affix into admin panel
- Changes root composer.json to use the `lunarphp/lunarmono` name
- Updates docs to show L12 compatibility
- Introduces new Laravel Installer starter kit installation approach
- Updates VitePress docs to latest version
- Fixes some Larastan errors with CartSessionManager


packages/admin/composer.json
```diff
@@ -15,7 +15,7 @@
     "minimum-stability": "dev",
     "require": {
         "lunarphp/core": "self.version",
-        "filament/filament": "^3.2.25",
+        "filament/filament": "^3.2.25|^3.3",
```
adevade: This change should not be necessary, right? Because it's a minor bump.

glennjacobs: Yeah, I think you're right, we should be able to stick with `^3.2.25`.

.github/workflows/static-analysis.yml
```diff
@@ -12,8 +12,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php: ["8.2", "8.3"]
-        laravel: ["^10.0", "^11.0", "^12.0"]
+        php: ["8.2", "8.3", "8.3"]
```
adevade: Probably a typo? `8.4`

glennjacobs: ta :-)

packages/core/src/Managers/CartSessionManager.php
```diff
@@ -75,7 +75,7 @@ public function forget(bool $delete = true): void
             );
         }
 
-        unset($this->cart);
+        $this->cart = $this->createNewCart();
```
glennjacobs: @alecritson can I have your thoughts on this?

alecritson: Not sure if this is correct? If we are forgetting a cart it doesn't always mean we want to pick up a new one.

glennjacobs: In which case the constructor may need to be updated.

PR #2127, opened 2025-02-27
Add ability to filter Customers by Customer Group

glennjacobs:
Previously there was no way to view the customers associated to a specific customer group. This update introduces a filter to the customers listing screen so that you can filter by customer group.


alecritson:
@glennjacobs Is it worth this being a multiple select filter and also show what customer groups the customer belongs to on the table?

glennjacobs:
> @glennjacobs Is it worth this being a multiple select filter and also show what customer groups the customer belongs to on the table?

Sorted.

alecritson (approved):

PR #2054, opened 2025-01-02
Add variant attributes to product edit page

alecritson:
Fixes #1996

PR #2134, opened 2025-03-15
Fix updating single variant attributes from the product form

### LINKED ISSUE #1996: Product variant attributes can not be edited if product has no options

fredmarmillod:
Fixes #1996

This is based on https://github.com/lunarphp/lunar/pull/2054

In the current implementation, the variant attributes are not properly hydrated in the ProductResource. 
This change creates a ProductFilamentProxy that extends the Product model and adds the variant attributes to the product using an accessor and the `$appends` property.

I created a new ProductFilamentProxy class to make this change as non-breaking as possible. It is probably not the best solution but it works and can then be refactored.

### DIFF
--- a/packages/admin/src/Filament/Resources/ProductResource.php
+++ b/packages/admin/src/Filament/Resources/ProductResource.php
@@ -32,8 +32,8 @@ use Lunar\Admin\Support\Tables\Columns\TranslatedTextColumn;
 use Lunar\FieldTypes\Text;
 use Lunar\FieldTypes\TranslatedText;
 use Lunar\Models\Attribute;
-use Lunar\Models\Contracts\Product;
 use Lunar\Models\Currency;
+use Lunar\Models\ProductFilamentProxy;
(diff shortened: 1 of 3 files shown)


glennjacobs:
Thanks, but I have solved in the original PR with a simpler approach.

PR #2135, opened 2025-03-17
Allow multiple extension to be registered for the same page/resource

alecritson:
Currently when registering extensions in the panel, they are identified by the class to be extended as the array key, for example:

```php
LunarPanel::extensions([
  CustomerResource::class => ExtensionA::class,
]);
```

This means you cannot register muliple extensions at once and would need to call `extensions` multiple times:

```php
LunarPanel::extensions([
  CustomerResource::class => ExtensionA::class,
]);

LunarPanel::extensions([
  CustomerResource::class => ExtensionB::class,
]);
```

This PR looks to introduce changes so you can pass an array of extensions instead, meaning it can all be done in one call:

```php
LunarPanel::extensions([
  CustomerResource::class => [ExtensionA:class, ExtensionB::class],
]);
```

Other notable tweaks

- Added a check that the extension class exists before registering to minimise exceptions
- Use `app($extension)` instead of `new $extension` to initialise the extension class.

### DIFF
--- a/packages/admin/src/LunarPanelManager.php
+++ b/packages/admin/src/LunarPanelManager.php
@@ -258,10 +258,28 @@ class LunarPanelManager
     public function extensions(array $extensions): self
     {
         foreach ($extensions as $class => $extension) {
-            $this->extensions[$class][] = new $extension;
+            if (! is_array($extension)) {
+                $extension = [$extension];
+            }
+
+            $extensions = collect($extension)->reject(
+                fn ($extension) => ! class_exists($extension)
+            )->map(
+                fn ($extension) => app($extension)
+            );
+
+            $this->extensions[$class] = [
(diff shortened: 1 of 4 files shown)


packages/admin/src/LunarPanelManager.php
```diff
@@ -258,12 +258,30 @@ protected function defaultPanel(): Panel
     public function extensions(array $extensions): self
     {
         foreach ($extensions as $class => $extension) {
-            $this->extensions[$class][] = new $extension;
+            if (! is_array($extension)) {
+                $extension = [$extension];
+            }
+
+            $extensions = collect($extension)->reject(
```
glennjacobs: `$extensions` is used in `foreach ($extensions` so we shouldn't be reusing it.

glennjacobs (approved):

PR #2137, opened 2025-03-18
Improve non-Scout searching

### LINKED ISSUE #2094: When trying to create a Product Association, the search engine does not find any products.

glennjacobs:
Fixes https://github.com/lunarphp/lunar/issues/2094

Currently, searching in the admin panel without Laravel Scout enabled isn't working 100%.

This PR aims to solve that by allowing a case insensitive search on JSON fields. To achieve this we introduce a builder macro to enhance Laravel's query builder.

- [x] Relation fields
- [x] Global search
- [x] Table search
- [ ] Document changing scout default to `false`


packages/admin/src/LunarPanelProvider.php
```diff
@@ -181,4 +176,42 @@ protected function registerLunarSynthesizer(): void
...
+                $last = array_pop($path);
+                foreach ($path as $part) {
+                    $jsonExpr .= "->'$part'";
+                }
+                $jsonExpr .= "->>'$last'";
+
+                return $boolean === 'or'
+                    ? $this->orWhereRaw("LOWER($jsonExpr) LIKE ?", [$searchTerm])
+                    : $this->whereRaw("LOWER($jsonExpr) LIKE ?", [$searchTerm]);
+            } else {
```
alecritson: `} else {` is redundant here

packages/admin/src/helpers.php
```diff
@@ -127,3 +118,34 @@ function get_search_builder($model, $search): Laravel\Scout\Builder|Builder
         }
     }
 }
+
+if (! function_exists('search_lunar_attributes')) {
+    function search_lunar_attributes(Builder &$query, string $search, string $modelMorphName): void
+    {
+        $attributes = Attribute::whereAttributeType(
```
alecritson: Should we use something like `Blink` here? If this gets called multiple times and depending how many attributes there are, could be a bottleneck?

packages/admin/src/helpers.php
```diff
@@ -127,3 +118,34 @@ function get_search_builder($model, $search): Laravel\Scout\Builder|Builder
         }
     }
 }
+
+if (! function_exists('search_lunar_attributes')) {
```
alecritson: Not sure on the naming here, it makes it seem like it should return results, but it's just adding logic to a query

packages/admin/src/LunarPanelProvider.php
```diff
@@ -181,4 +176,42 @@ protected function registerLunarSynthesizer(): void
         \Lunar\Admin\Support\Facades\AttributeData::synthesizeLivewireProperties();
         Livewire::propertySynthesizer(PriceSynth::class);
     }
+
+    protected function registerBuilderMacros(): void
+    {
+        Builder::macro('whereJsonContainsInsensitive', function (
```
alecritson, line 182: Could do with a unit test or sorts to ensure the query returned is what you're expecting based on input given.

packages/admin/src/helpers.php
```diff
@@ -127,3 +118,34 @@ function get_search_builder($model, $search): Laravel\Scout\Builder|Builder
...
+    function search_lunar_attributes(Builder &$query, string $search, string $modelMorphName): void
```
alecritson: We shouldn't need to pass by reference here.

packages/admin/src/LunarPanelProvider.php
glennjacobs: Yeah, that could work. We can't test in MySQL or Postgres, so probably the best approach!

packages/admin/src/helpers.php
glennjacobs: I'll check this, but pretty sure it was for a reason.

alecritson:
Currently this doesn't appear to work properly for multiple paths. For example:

```php
Product::query()->whereJsonContainsInsensitive('attribute_data', ['name', 'description'], 'value')->toRawSql()
```

Yields the following query in MySQL

```sql
select * from "lunar_products" where LOWER(json_unquote(json_extract(attribute_data, '$."name"'))) LIKE '%value%' and LOWER(json_unquote(json_extract(attribute_data, '$."name"."description"'))) LIKE '%value%' and "lunar_products"."deleted_at" is null
```

glennjacobs:
No longer required, fixed via Filament update.

PR #2138, opened 2025-03-19
All cart items stock check should consider shippability

ryanmitchell:
The `allCartItemsAreInStock` check in ShippingRateResolver needs to consider whether the items are shippable or not as well as the stock level.

This PR updates the logic to that.

### DIFF
--- a/packages/table-rate-shipping/src/Resolvers/ShippingRateResolver.php
+++ b/packages/table-rate-shipping/src/Resolvers/ShippingRateResolver.php
@@ -60,7 +60,7 @@ class ShippingRateResolver
         $shippingMeta = $cart->shippingEstimateMeta;
 
         $this->allCartItemsAreInStock = ! $this->cart->lines->first(function ($line) {
-            return $line->purchasable->stock < $line->quantity;
+            return $line->purchasable->shippable && ($line->purchasable->stock < $line->quantity);
         });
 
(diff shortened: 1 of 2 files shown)


glennjacobs:
@ryanmitchell why would stock calculations care about whether something can be shipped? Surely it should simply be based upon Purchasability?

ryanmitchell:
If something can't be shipped then it shouldnt affect the availability of a shipping rate, its being delivered to the customer in a different way (eg email?)

glennjacobs:
> If something can't be shipped then it shouldnt affect the availability of a shipping rate, its being delivered to the customer in a different way (eg email?)

That's quite opinionated. I might want to sell a digital product but limit it to 20 items. 

I think the answer is to simply set Purchasability to "Always".

glennjacobs:
Sorry, just seen this is on the shipping resolver :D

glennjacobs:
Right, makes more sense now haha. Yep, this seems fine then.

alecritson:
Thanks @ryanmitchell Can we add an assertion on the `tests/shipping/Unit/Resolvers/ShippingRateResolverTest.php` to ensure this works as intended. Or let me know if you want me to take this over and have a go.

ryanmitchell:
@alecritson that's a test added

alecritson (approved):

Issue #2139, opened 2025-03-19
Postgres SQLSTATE[42883] error when entering the customers page


CasSchuurman1:
- Lunar version: 1.0.0-beta installed like this composer require lunarphp/lunar:"^1.0.0-beta" -W
- Laravel Version: 12
- PHP Version: 8.4
- Database Driver & Version: Postgres 15.12

### Expected Behaviour:
The page should function correctly, as it does with MySQL.

### Actual Behaviour:
This error shows up
SQLSTATE[42883]: Undefined function: 7 ERROR: could not identify an equality operator for type json LINE 1: select distinct "lunar_customer_groups".* from "lunar_custom... ^ (Connection: pgsql, SQL: select distinct "lunar_customer_groups".* from "lunar_customer_groups" left join "lunar_customer_customer_group" on "lunar_customer_groups"."id" = "lunar_customer_customer_group"."customer_group_id" order by "lunar_customer_groups"."name" asc) 

### Steps To Reproduce:
1. Set up a Lunar project with a PostgreSQL database.  
2. Navigate to the /lunar/customer page.  
3. Observe the behavior.

tiznull:
I had the same error, please approve this merge. Tested the fix and it works.

alecgarcia:
I also ran into this issue, tested the fix, and can confirm that it works. 

Thanks!

PR #2146, opened 2025-03-25
fix: postgresql not loading in customer page fixed lunarphp/lunar#2139

CasSchuurman1:
This pull request fixes the error from issue #2139.

### DIFF
--- a/packages/admin/src/Filament/Resources/CustomerResource.php
+++ b/packages/admin/src/Filament/Resources/CustomerResource.php
@@ -207,7 +207,13 @@ class CustomerResource extends BaseResource
             ->filters([
                 Tables\Filters\SelectFilter::make('customer_group')
                     ->label(__('lunarpanel::customergroup.label'))
-                    ->relationship('customerGroups', 'name')
+                    ->relationship(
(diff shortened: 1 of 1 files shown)


tiznull:
Please approve this pull request. I had the same error, I tested the fix and it works.

alecritson (approved):
Ready for the next release, thank you for this @CasSchuurman1

shanebrightbulb:
I've just updated and tested the customer group filter on Postgre and I'm getting the following SQL error:

`SQLSTATE[42702]: Ambiguous column: 7 ERROR: DISTINCT ON "id" is ambiguous LINE 1: ...om "customers" where (exists (select distinct on ("id", "nam... ^ (Connection: pgsql, SQL: select count(*) as aggregate from "customers" where (exists (select distinct on ("id", "name", "handle", "default") * from "customer_groups" inner join "customer_customer_group" on "customer_groups"."id" = "customer_customer_group"."customer_group_id" where "customers"."id" = "customer_customer_group"."customer_id" and "customer_groups"."id" in (1))))`

PR #2151, opened 2025-03-28
fix: saving single variant attribute from product page

fredmarmillod:
This PR fixes an issue with single variant product attribute saving when attribute types are not Text or Translatable Text.

https://github.com/lunarphp/lunar/issues/1996#issuecomment-2760878461

**Steps to reproduce:**

1. Install Lunar starterkit
2. Create a attribute group
3. Add a Number field to the group
4. Add the group to a product type
5. Create a new product
6. Try to save the product

**Expected behavior:**

The product should save successfully.

**Actual behavior:**

It triggers a validation error.

### DIFF
--- a/packages/admin/src/Filament/Resources/ProductResource.php
+++ b/packages/admin/src/Filament/Resources/ProductResource.php
@@ -226,8 +226,7 @@ class ProductResource extends BaseResource
 
     protected static function getAttributeDataFormComponent(): Component
     {
-        return Attributes::make()
-            ->statePath('attribute_data');
+        return Attributes::make();
     }
 
     protected static function getVariantAttributeDataFormComponent(): Component
--- a/packages/admin/src/Filament/Resources/ProductResource.php
+++ b/packages/admin/src/Filament/Resources/ProductResource.php
@@ -234,7 +234,7 @@ class ProductResource extends BaseResource
     {
         return Attributes::make()
             ->using(ProductVariant::class)
-            ->statePath('variant.attribute_data');
+            ->relationship('variant');
     }
 
     public static function getDefaultTable(Table $table): Table
--- a/packages/admin/src/Filament/Resources/ProductResource.php
+++ b/packages/admin/src/Filament/Resources/ProductResource.php
@@ -226,14 +226,15 @@ class ProductResource extends BaseResource
 
     protected static function getAttributeDataFormComponent(): Component
     {
-        return Attributes::make();
+        return Attributes::make()
+            ->statePath('attribute_data');
     }
 
     protected static function getVariantAttributeDataFormComponent(): Component
     {
         return Attributes::make()
             ->using(ProductVariant::class)
-            ->relationship('variant');
+            ->statePath('variant.attribute_data');
     }
 
     public static function getDefaultTable(Table $table): Table
--- a/packages/admin/src/Support/Forms/Components/Attributes.php
+++ b/packages/admin/src/Support/Forms/Components/Attributes.php
@@ -87,8 +87,7 @@ class Attributes extends Forms\Components\Group
                         $sectionFields[] = AttributeData::getFilamentComponent($field);
                     }
                     $groupComponents[] = Forms\Components\Section::make($group['model']->translate('name'))
-                        ->schema($sectionFields)
-                        ->statePath('attribute_data');
+                        ->schema($sectionFields);
                 }
 
                 return $groupComponents;
@@ -101,12 +100,6 @@ class Attributes extends Forms\Components\Group
             }
 
             foreach ($state as $key => $value) {
-                if ($key === 'attribute_data') {
-                    $state[$key] = array_map(fn ($item) => $item->getValue(), $value);
-
-                    continue;
-                }
-
                 if (! $value instanceof \Lunar\Base\FieldType) {
                     continue;
                 }
--- a/packages/admin/src/Support/Forms/Components/Attributes.php
+++ b/packages/admin/src/Support/Forms/Components/Attributes.php
@@ -20,6 +20,8 @@ class Attributes extends Forms\Components\Group
     {
         $this->modelClassOverride = $modelClass;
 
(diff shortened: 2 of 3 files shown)


alecritson:
Hi @fredmarmillod Thanks for this, unfortunately it doesn't seem to fix the issue for me. I still get the validation error, which I agree is a problem 🤔

fredmarmillod:
You're right, I've retried and it works only if the product type only has variant attributes configured. When we add product attributes, it triggers a validation error only for those. It seems like there is a conflict with the `attribute_data` state path.

fredmarmillod:
@alecritson The changes appear to be working correctly now. Could you please review the implementation and let me know if you notice any potential issues?

packages/admin/src/Support/Forms/Components/Attributes.php
```diff
@@ -20,14 +20,16 @@ public function using(string $modelClass): self
...
+
         return $this;
     }
 
     protected function setUp(): void
     {
         parent::setUp();
 
-        $this->key('attributeData'.$this->modelClassOverride);
+        $this->key('attributeData');
```
wychoong: could have just override the getKey and prefix it 

```php
public function getKey(): ?string
{
    $key = $this->evaluate($this->key);

    return 'attributeData' . $key;
}
```

packages/admin/src/Support/Forms/Components/Attributes.php
```diff
@@ -88,7 +90,7 @@ protected function setUp(): void
                     }
                     $groupComponents[] = Forms\Components\Section::make($group['model']->translate('name'))
                         ->schema($sectionFields)
-                        ->statePath('attribute_data');
+                        ->statePath($modelClass === ProductVariant::class ? 'attribute_data' : null);
                 }
```
wychoong, line 95: this is setting every single fields' statePath to be 'attribute_data', should be removed

packages/admin/src/Support/Forms/Components/Attributes.php
```diff
@@ -20,14 +20,16 @@ public function using(string $modelClass): self
...
         $this->modelClassOverride = $modelClass;
 
+        $this->key('attributeData'.$this->modelClassOverride);
+
         return $this;
     }
 
     protected function setUp(): void
     {
         parent::setUp();
```
wychoong, line 31:
```php
        $this->statePath('attribute_data');
``` 

just add this will tell filament this whole Group field statePath is attribute_data

packages/admin/src/Filament/Resources/ProductResource.php
```diff
@@ -226,7 +226,8 @@ protected static function getTagsFormComponent(): Component
 
     protected static function getAttributeDataFormComponent(): Component
     {
-        return Attributes::make();
+        return Attributes::make()
+            ->statePath('attribute_data');
```
wychoong: after having the `statePath` in `setUp`, we don't need this in every Attributes field usage, but we can still override it if need to. Lunar should assume all usage of this field is attribute_data, and if used user land dev should override it

packages/admin/src/Support/Forms/Components/Attributes.php
```diff
@@ -100,6 +102,12 @@ protected function setUp(): void
                 return $state;
             }
 
+            if ($this->modelClassOverride === ProductVariant::class) {
+                $state = [
+                    'attribute_data' => array_map(fn ($item) => $item->getValue(), $state['attribute_data']),
+                ];
+            }
```
wychoong, line 110:
```suggestion
```

once everything in place, we don't need this anymore

wychoong:
@alecritson can you have a look at my comments, as this is a blocker issue.
or your prefer I make a new PR

fredmarmillod:
@wychoong Thank you for reviewing this PR, I'll make the necessary changes

packages/admin/src/Support/Forms/Components/Attributes.php
wychoong, line 31:
```suggestion
    protected function setUp(): void
    {
        parent::setUp();
        
        $this->statePath('attribute_data');

```

packages/admin/src/Support/Forms/Components/Attributes.php
fredmarmillod: I tried to remove that as well as the statePath from the Section component but for some reason it's not working. If I `dd($state)` in the `mutateStateForValidationUsing` method I can see that it contains the whole variant properties along with the attributes. And the attributes are sent in the `attribute_data` which is correct, but they are also sent as top level properties which triggers an error when saving the variant.

```
"id" => 2
"product_id" => 2
"tax_class_id" => 1
"attribute_data" => [
    "size" => Lunar\FieldTypes\Number {#5371 ▶}
    "color" => Lunar\FieldTypes\Text {#4798 ▶}
]
"tax_ref" => null
"unit_quantity" => 1
"min_quantity" => 1
... other variant properties
"size" => Lunar\FieldTypes\Number {#5371 ▶}
"color" => Lunar\FieldTypes\Text {#4798 ▶}
```

That's why I've added code to cleanup to `$state` before saving.

wychoong: did you remove the statePath in line 92-95

fredmarmillod: Yes I did

fusionconsulting:
I've run into this issue as well. I've tried the changes above and run into the same issue as the failed test. Happy to test any changes that might resolve the issue. I'm not familiar enough with the code to offer a PR yet but I'm looking through the code in the mean time.

wychoong:
@fredmarmillod @fusionconsulting i made another PR #2159 to fix the issue

alecritson:
Your time and energy on this is appreciated, however we opted to go for the approach @wychoong went for as it seemed to fill a few more gaps with attributes and editing. 🙏

PR #2159, opened 2025-04-03
Fix saving product single variant

wychoong:
fixes bug introduced in #2054

### DIFF
--- a/packages/admin/src/Support/Forms/Components/Attributes.php
+++ b/packages/admin/src/Support/Forms/Components/Attributes.php
@@ -2,6 +2,7 @@
 
 namespace Lunar\Admin\Support\Forms\Components;
 
+use Closure;
 use Filament\Forms;
 use Illuminate\Database\Eloquent\Model;
 use Livewire\Component as Livewire;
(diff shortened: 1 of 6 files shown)


packages/admin/src/Support/Forms/Components/Attributes.php
```diff
@@ -110,5 +134,15 @@ protected function setUp(): void
...
+
+        $this->mutateRelationshipDataBeforeSaveUsing(static function (Attributes $component, array $data): array {
+            return [
+                $component->getAttributeDataField() => $data,
+            ];
+        });
+
+        $this->mutateRelationshipDataBeforeFillUsing(static function (Attributes $component, array $data): array {
+            return $data[$component->getAttributeDataField()] ?? [];
+        });
```
wychoong, line 146: when setting the component `->relationship('variant')` during editing product with single variant, the relationship override the statepath, causing issue when saving or hydrating as filament attempt to get the value from variant->`attribute-handle` or saving to wrong column

adding these two mutate methods tell filament to get the correct field

alecritson:
Thanks @wychoong I think it seems to work well. The only thing I was also looking at was hiding the attributes when we have more than one variant.

wychoong:
@alecritson done

glennjacobs (approved):

PR #2171, opened 2025-04-22
Refactor order reference generator

### LINKED ISSUE #2164: Order references are limited to 9999 orders
After that Lunar starts clashing references with this error:

```
Unique violation: 7 ERROR:  duplicate key value violates unique constraint "lunar_orders_reference_unique"...
```

The issue has 2 roots:

1) This is trying to `max()` on a string:

``` php
 DB::RAW('MAX(reference) as reference')
```

The problem is that, because the next one is `2025-04-10000`, when generating the next code it always find 9999 as the last one, because in string comparison: 

```
2025-04-9999... > 2025-04-10000...
```

And also from this line in the [OrderReferenceGenerator class](https://github.com/lunarphp/lunar/blob/e9bf055f76bb44582a09fad1e0e621fd79668679/packages/core/src/Base/OrderReferenceGenerator.php#L38C9-L38C77):

``` php
return $year.'-'.$month.'-'.str_pad($increment, 4, 0, STR_PAD_LEFT);
````

### PR DESCRIPTION
**Pull request by alecritson**
Fixes #2164 by making some assumptions about the format and then casting to unsigned so the `MAX` function works as expected. This change also allows the generator code to be slim-lined.

### DIFF
--- a/tests/core/Unit/Base/OrderReferenceGeneratorTest.php
+++ b/tests/core/Unit/Base/OrderReferenceGeneratorTest.php
@@ -56,3 +56,31 @@ function can_increment_order_reference_by_default()
 
     expect($result)->toEqual($order->created_at->format('Y-m').'-0002');
 }
+
+test('can generate high reference numbers', function () {
+    Order::factory()->create([
+        'reference' => '2024-04-5000000',
+        'placed_at' => now(),
+    ]);
+
+    Order::factory()->create([
+        'reference' => '2025-04-9999',
+        'placed_at' => now(),
+    ]);
+
+    Order::factory()->create([
+        'reference' => '2025-04-10000',
+        'placed_at' => now(),
+    ]);
+
+    $order = Order::factory()->create([
+        'reference' => null,
+        'placed_at' => now(),
+    ]);
+
+    expect($order->reference)->toBeNull();
+
+    $result = app(OrderReferenceGenerator::class)->generate($order);
+
+    expect($result)->toEqual($order->created_at->format('Y-m').'-10001');
+});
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: tests/core/Unit/Base/OrderReferenceGeneratorTest.php [hunk 1]
```diff
@@ -56,3 +56,31 @@ function can_increment_order_reference_by_default()
...
+    $order = Order::factory()->create([
+        'reference' => null,
+        'placed_at' => now(),
+    ]);
+
+    expect($order->reference)->toBeNull();
+
+    $result = app(OrderReferenceGenerator::class)->generate($order);
+
+    expect($result)->toEqual($order->created_at->format('Y-m').'-10001');
```
wychoong, line 87: will this test break after year 2025?

alecritson: 😃

antonioribeiro:
Hello @alecritson, thanks for your PR! 

About [this code](https://github.com/lunarphp/lunar/pull/2171/files#diff-bec634531fe52cc81eeb8de3de07b796895acd606bfb3648d6e360de9b5db302R19): 

```sql
$rawSelect = $connection != 'sqlite' ?
  'MAX(CAST(SUBSTRING(reference from 9) as UNSIGNED))' :
  'MAX(CAST(substr(reference, 9) as UNSIGNED))';
```

This can be a performance bottleneck as it forces the database to perform these operations on every single row in the table

alecritson:
After internal discussion we have decided on a different approach

alecritson:
The new approach can be seen on #2175

PR #2172, opened 2025-04-24
Make discount table columns toggleable

ryanmitchell:
This PR adds toggling to the discount table columns, and adding coupon and created at columns, hidden by default.

It makes the name, coupon and date fields sortable.

And it adds a unique validation on the coupon field, as currently you can add discount codes with the same coupon which throws an unhandled database error.

### DIFF
--- a/packages/admin/src/Filament/Resources/DiscountResource.php
+++ b/packages/admin/src/Filament/Resources/DiscountResource.php
@@ -209,7 +209,8 @@ class DiscountResource extends BaseResource
                 __('lunarpanel::discount.form.coupon.label')
             )->helperText(
                 __('lunarpanel::discount.form.coupon.helper_text')
-            );
+            )
+            ->unique(ignoreRecord: true);
     }
 
     protected static function getMaxUsesFormComponent(): Component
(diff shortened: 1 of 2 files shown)


packages/admin/src/Filament/Resources/DiscountResource.php
```diff
@@ -359,21 +360,39 @@ protected static function getTableColumns(): array
...
+                ->sortable()
+                ->toggleable(),
             Tables\Columns\TextColumn::make('ends_at')
                 ->label(__('lunarpanel::discount.table.ends_at.label'))
-                ->date(),
+                ->date()
+                ->sortable()
+                ->toggleable(),
+            Tables\Columns\TextColumn::make('coupon')
+                ->label(__('lunarpanel::discount.table.name.coupon.label'))
```
alecritson: Looks like this is missing or incorrect?

ryanmitchell: Thanks, I've updated that.

alecritson (approved):

PR #2175, opened 2025-04-28
Refactor order reference generator

alecritson:
This PR changes the way order references are generated. Currently the format YYYY-MM-XXXXXX creates difficulty when attemping to determine the next reference in the sequence due to being a string.

The decision was to move away from any sort of queries when determining the next reference and instead just use the ID of the order and allow developers to customise certain formatting options within config, so in most cases there should be no need to use a custom generator for small changes.


docs/core/upgrading.md
```diff
@@ -18,6 +18,55 @@ php artisan migrate
...
+
+#### Order reference generation changes
+
+The current order reference generator uses the format `YYYY-MM-{X}` which has been implemented since the early days of when Lunar was called GetCandy.
+
+This approach to formatting is not great for order references and can lead to anomalies when attempting to determine the next reference in the sequence.
+
+The new format uses the Order ID and adds leading zeros and an optional prefix i.e.
+
+Assuming order ID is 19560
```
wychoong: the example is based on 19560 or 1965?

glennjacobs (dismissed):

PR #2176, opened 2025-04-28
Add Two-Factor Authentication to the panel

alecritson:
This PR looks to introduce the ability for Staff members to set up Two-Factor Authentication for their account and also gives developers the option to enforce it for all members, or disable it entirely.

Two-Factor Authentication is quite standard these days and  really should be available on Lunar as standard.


composer.json
```diff
@@ -37,6 +37,7 @@
         "spatie/laravel-medialibrary": "^11.12.7",
         "spatie/laravel-permission": "^6.12",
         "spatie/php-structure-discoverer": "^2.3.1",
+        "stephenjude/filament-two-factor-authentication": "dev-main",
```
glennjacobs, line 40: Can we use a release? `0.0.23` seems to be the latest. Bit of an odd release number, would `0.*` suffice perhaps?

packages/admin/src/LunarPanelManager.php
```diff
@@ -33,9 +33,14 @@
 use Lunar\Admin\Filament\Widgets\Dashboard\Orders\PopularProductsTable;
 use Lunar\Admin\Http\Controllers\DownloadPdfController;
 use Lunar\Admin\Support\Facades\LunarAccessControl;
+use Stephenjude\FilamentTwoFactorAuthentication\TwoFactorAuthenticationPlugin;
 
 class LunarPanelManager
 {
+    protected bool $twoFactorAuthEnforced = false;
+
+    protected bool $twoFactorAuthDisabled = false;
```
wychoong, line 42: is there a need for having 2 properties?

alecritson: Having both set to `false` will mean that each staff member will have the option to enable 2FA on their account

`twoFactorAuthEnforced` - Means when a staff member logs in, if they don't have 2FA set up they will be forced to do so

`twoFactorAuthDisabled` - 2FA will not be installed and Staff won't have the option to enable it nor will it be enforced.

glennjacobs (approved):

PR #2180, opened 2025-05-07
Add event for stripe webhook

alecritson:
Currently if an Order or Cart model cannot be found when processing an intent, we just log the error.  This approach doesn't give developers a chance to respond to the problem in their apps and therefore this change will dispatch an event so listeners are able to respond if needed.

### DIFF
--- a/packages/stripe/README.md
+++ b/packages/stripe/README.md
@@ -381,6 +381,22 @@ const submit = async () => {
 ```
 ---
 
+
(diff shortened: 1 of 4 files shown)


glennjacobs:
@alecritson could we add a test for this?

alecritson:
> @alecritson could we add a test for this?

Donezo

PR #2186, opened 2025-05-14
Fix StorefrontSessionManager not storing Currency in session

adevade:
This PR updates the StorefrontSessionManager to store the currency in the session.


glennjacobs (changes requested):
This sets the currency in the session, but it doesn't load the currency from the session in an initialisation method.

We could do with a `$this->initCurrency();`

glennjacobs:
@adevade I see we also have #2187 , so do we still need this PR?

adevade:
@glennjacobs Yeah, no I think these changes are included in the other PR as well.

I based it on this branch.

glennjacobs:
Closing in favour of #2187

PR #2187, opened 2025-05-15
Refactor StorefrontSessionManager

adevade:
> [!NOTE]
> Builds on #2186 *(I think the changes are included here as well..?)*

This PR introduces significant updates to the `StorefrontSession` system, focusing on improving type safety, enhancing initialization logic, and refining test configurations. 

### Updates to `StorefrontSession` Interface and Manager:

- **General Cleanup**:
  - Reordered methods and fixed missing/inconsistent/wrong types.

- **Initialization Enhancements**:
  - Refactored initialization methods (`initChannel`, `initCustomerGroups`, `initCurrency`, and `initCustomer`) in `StorefrontSessionManager` to streamline session setup and ensure default values are applied when session data is missing.

### Updates to `StorefrontSession` Facade:

- **Facade Method Definitions**:
  - Added static method definitions to the `StorefrontSession` facade for improved IDE support and type hinting.

### Testing Improvements:

- **Test File Cleanup**:
  - Remove duplicated code and add general setup to `beforeEach()`.

- **Test Initialization**:
  - Added a `beforeEach` hook in `StorefrontSessionManagerTest` to set up the authentication user configuration before each test. (`tests/core/Unit/Managers/StorefrontSessionManagerTest.php`)

- **Test Improvements**:
  - Added more tests and expectations for better test coverage.

#### Unused methods

There's two methods on `StorefrontSessionManager` (`resetCustomerGroups()` and `forget()`) that were not tested, and `forget()` was not working in the old implementation (now fixed). Should these be included, and if so, should we add methods for resetting the other properties as well?

#### Automatic initialization 

When the class is instantiated it automatically inits `Channel`, `CustomerGroup`, `Currency` and `Customer` (if found). This is already mentioned in the docs, but maybe should be left to the developer to manually init only the parts they want? If so we could add a `init()` helper to keep the current behaviour.


glennjacobs:
@alecritson this PR has a lot of updates to tidy the StorefrontSessionManager, but it also includes a fundamental update to currencies.

Previously currencies were not initialised in the storefront session manager, nor were they stored in the session. This goes against the fundamental purpose of the manager.

I can't see any issues, but putting it past you just in case :)

alecritson:
> @alecritson this PR has a lot of updates to tidy the StorefrontSessionManager, but it also includes a fundamental update to currencies.
> 
> Previously currencies were not initialised in the storefront session manager, nor were they stored in the session. This goes against the fundamental purpose of the manager.
> 
> I can't see any issues, but putting it past you just in case :)

I think it's probably fine

alecritson (approved):

PR #2202, opened 2025-05-27
Add amount check to payment type

alecritson:
This PR adds a new check into the Stripe PaymentType to ensure that the PaymentIntent amount equals the Order total.

Generally frontends will need to ensure they update the PaymentIntent amount when carts change, there could be situations where this hasn't happened and the order should not be marked as processed.


glennjacobs:
I agree we probably need a check to ensure orders are fully paid, perhaps it should be more of a UI thing in the admin system though?

If someone wants to build a store that takes a deposit for example, this would prevent it.

As long as the transaction is logging the amount taken, then we should be good.

@alecritson considering this, should we close this PR?

alecritson:
> I agree we probably need a check to ensure orders are fully paid, perhaps it should be more of a UI thing in the admin system though?
> 
> If someone wants to build a store that takes a deposit for example, this would prevent it.
> 
> As long as the transaction is logging the amount taken, then we should be good.
> 
> @alecritson considering this, should we close this PR?

What happens in a situation when an order comes in as placed but the payment intent amount is less than what the order total is? In Lunar it would currently be marked as paid with a successful status.

I understand this would stop someone taking a deposit however I feel that this is somewhat of a rare case and could instead be allowed with config/parameters passed to allow partial intents.

glennjacobs (changes requested):
OK, I think if the payment manager can have the option to allow partial payments (e.g. deposits), but default to expecting the full amount, that would work.

PR #2205, opened 2025-06-03
Auto document facades

repl6669:
Facades in Lunar are generally "undocumented", meaning they do not have proper docblocks for static methods. This can be easily solved by introducing a new workflow which runs Laravel [facade-documenter](https://github.com/laravel/framework/blob/12.x/.github/workflows/facades.yml) automatically.

* Added `@see` to all facades where missing
* Added `document-facades` workflow which runs `laravel/facade-documenter` and updates facade docblocks

Example automatic of the new workflow: https://github.com/lunarphp/lunar/pull/2205/commits/bbee19830af13619bafbcff3672132a97d3be871.


glennjacobs:
Thanks @repl6669, this is something I'd actually wanted for a while now 👍

glennjacobs (approved):

PR #2214, opened 2025-06-16
Allow BuyXGetY discount conditions to be collection or variant, rewards to be variant

ryanmitchell:
This PR extends the BuyXGetY discount conditions to support `Collection` or `Variant` as a condition, and `Variant` as a reward:

CleanShot 2025-06-16 at 12 01 00@2x

This allows you to choose collections or variants that the discount applies to, not just products, and rewards can be scoped to specific variants, not just products.

As its no longer 'Product Conditions' and `Product Rewards`, I've also changed lang strings for some elements on the page to make things match the new approach.

## Summary by CodeRabbit

* **New Features**
  * Discount conditions and rewards now support collections and product variants, expanding beyond just products.
  * Discount management UI allows selection of collections, products, and variants as discountables.
  * New interface for thumbnail images ensures consistent display across collections, products, and variants.
  * Added new scope to filter discounts by collections.

* **Improvements**
  * Admin interface uses clearer, more general terminology for discounts, conditions, and rewards.
  * Discount filtering now includes collections, enhancing discount applicability.

* **Bug Fixes**
  * Case-insensitive search is standardized in the admin panel.

* **Documentation**
  * Updated documentation to reflect new discountable terminology, relationships, and upgrade instructions.

* **Chores**
  * Refactored tests and codebase to use new discountable naming conventions and relationships.
  * Database migration renames and updates discount-related tables and columns for consistency.


glennjacobs (changes requested):
@ryanmitchell I'm not against this idea, in fact I can see it would be handy, but the implementation feels wrong at the moment.

The relation table is called discount_purchasables and so you'd expect the entries to be a purchasable. To now start putting non-purchasables in is confusing.

I'd probably do the database tables differently now, with perhaps discount_conditions & discounts_rewards... it would need some proper thought, but it doesn't help us right now. That sort of change would have to be done in v2.

Can you have a think about how we can implement this, but keep things clean?

ryanmitchell:
Yeah I agree.

Currently there are 4 types of 'purchasable' - conditions, rewards, exclusions and limitations, so that would mean 4 tables, and probably more as other discount types are made. Keeping one table makes more sense to me and is more flexible when it's just the 'type' changing each time.

Some suggestions for what we could rename `discount_purchasables` to:

`discount_connections`
`discount_discountables`
`discount_relations`

or just simply:
`discountables` (<-- this would be my preference as its a convention used in Lunar already)

Do any of these sit well with you?

glennjacobs:
Yeah, discountables I think would work. Would you be willing to make the change in this PR?

ryanmitchell:
Yep absolutely, leave it with me and I'll get it updated today.

ryanmitchell:
@glennjacobs thats updated. theres a failing test but its not related to this code - seem to be a random failure.

glennjacobs:
@alecritson you happy with this route?

alecritson:
> @alecritson you happy with this route?

From reading it seems sound, have you pulled down and tried it out in the panel?

ryanmitchell:
If you're happy with this direction of travel I'd love to go a step further and allow variants to be selected on Buy X get Y too.

glennjacobs:
> If you're happy with this direction of travel I'd love to go a step further and allow variants to be selected on Buy X get Y too.

Sure, I don't see any issue with that.

ryanmitchell:
thanks @glennjacobs thats been done

glennjacobs (changes requested):
When a variant has no associated media, I get the following error

> Call to undefined relationship [media] on model [Lunar\Models\ProductVariant].

ryanmitchell:
Yeah I'm seeing that too - looking at the image column youre using there doesnt seem to be an option to prevent it being shown. Is the best approach to extend the column and have a lunar version of it?

glennjacobs:
> Yeah I'm seeing that too - looking at the image column youre using there doesnt seem to be an option to prevent it being shown. Is the best approach to extend the column and have a lunar version of it?

Any of the stuff in here work? <https://github.com/filamentphp/filament/blob/3.x/packages/tables/src/Columns/ImageColumn.php>

Thinking `visibility`, `defaultImageUrl()`, etc.

I'd prefer not to extend if we can.

ryanmitchell:
Unfortunately not, as the image column is applying eager loading and thats where the error comes from. As another option could we map the 'images' media collection on variants to 'media' ? Failing that I think it would be a case of having a custom column that returns the image using formatStateUsing

glennjacobs:
If we need to get an image from different models, perhaps we need a contract, like `HasImages` and then we can consistently get the appropriate thumbnail URL.

ryanmitchell:
I could be misunderstanding what you are suggesting, but a contract wont help as this is happening at the query level, not at the model level. Do we really need the image column at all?

glennjacobs:
> I could be misunderstanding what you are suggesting, but a contract wont help as this is happening at the query level, not at the model level. Do we really need the image column at all?

I'm suggesting if we had a standard way of getting a thumbnail URL from models, regardless of whether they're directly implementing Spatie Media or not, then we could use that with a custom column.

ryanmitchell:
Ah, following now. Ok, this is getting bigger! I'll work on that.

ryanmitchell:
@glennjacobs let me know what you think of the recent changes

packages/admin/src/Filament/Resources/DiscountResource/RelationManagers/ProductRewardRelationManager.php
```diff
@@ -64,19 +76,26 @@ public function getDefaultTable(Table $table): Table
                     return $data;
                 }),
             ])->columns([
-                Tables\Columns\SpatieMediaLibraryImageColumn::make('purchasable.thumbnail')
-                    ->collection(config('lunar.media.collection'))
-                    ->conversion('small')
-                    ->limit(1)
-                    ->square()
+                ThumbnailImageColumn::make('discountable_id')
```
wychoong, line 80: Is the media image eager loaded?
Since going standardizing lunar models to have same method to retrieve image, can we explore a simpler method? Eg: ->via(‘discountable’) that configure resolveThumbnailUrlUsing

ryanmitchell: I don’t follow sorry.

wychoong: something along this

```php
// I might be wrong, but using 'discountable_id' only load media during getting image url
// causing N+1
ThumbnailImageColumn::make('discountable') 
      ->through('discountable')
```

```php
class ThumbnailImageColumn extends ImageColumn
{
    protected \Closure $resolveThumbnailUrlUsing;

    protected function setUp(): void
    {
        parent::setUp();
       $this->label('');   // this is so that image column label default no label
    }
    
    public function through(string $through): self {
        $this->resolveThumbnailUrlUsing(fn (?Model $record) => $record?->{$through}?->getThumbnailImage());
        
        return $this;
    }
    public function getImageUrl(?string $state = null): ?string
    {
        // ....
    }

    public function resolveThumbnailUrlUsing(\Closure $callback): self
    {
        // ....
    }
}
```

ryanmitchell: That will still cause n+1, which I think is unavoidable in this case due to the media relation being different across the different models. Its the balance between trying to create something generic enough for others to extend versus something specific to the current use case.

glennjacobs: I do think showing a thumbnail image helps, and I can't imagine the lazy loading will become a problem until you literally have 100's of discountables, but some might I guess.

Part of me wonders if we could have a column type that lazy-loads in the UI. So it requests the image after the page has loaded and only when the row is visible. But are we trying to fix a problem that perhaps doesn't exist?

Maybe we run with the current approach and go from there?

@alecritson be good to have your input on this one.

wychoong: > Maybe we run with the current approach and go from there?

agree

glennjacobs (changes requested):
Could we add something to the upgrade guide to explain the change and the breaking changes, please?

And do the Discount docs also need updating?

ryanmitchell:
Yes no problem - thats done

docs/core/upgrading.md
```diff
@@ -18,6 +18,14 @@ php artisan migrate
 
 Lunar currently provides bug fixes and security updates for only the latest minor release, e.g. `0.8`.
 
+## 1.0.0-beta.23
```
glennjacobs: It would be beta.24 now

glennjacobs (approved):

PR #2221, opened 2025-06-23
test: provide failing test case to prove lazy loading exceptions are thrown

jordyvanderhaegen:


glennjacobs:
@alecritson I've add a "quick fix" to this, but I'm a bit concerned about the `$cartLine->cart` usage. Typically we're loading the cart and then looping around the cart lines. So for a cart line to then load up the cart is probably wrong. Really any cart line loaded should already have the cart relation set somehow I guess.

See https://github.com/lunarphp/lunar/pull/2221/files#diff-260215554a74ed5b0d812ad5f731b85e261340625f81727e604adb29fd4d5ab0R26

glennjacobs:
Following on from this, I have added `Model::preventLazyLoading(true);` to the entire Core testsuite and it's breaking in a fair few places. 

If we are going to support this (which I think we should), then we should fix it all.

alecritson:
> @alecritson I've add a "quick fix" to this, but I'm a bit concerned about the `$cartLine->cart` usage. Typically we're loading the cart and then looping around the cart lines. So for a cart line to then load up the cart is probably wrong. Really any cart line loaded should already have the cart relation set somehow I guess.
> 
> See https://github.com/lunarphp/lunar/pull/2221/files#diff-260215554a74ed5b0d812ad5f731b85e261340625f81727e604adb29fd4d5ab0R26

Yeah price calculation has always been a bit of a beast, I think as well due to using pipelines we only really send the CartLine model though and sometimes we need the context of the Cart itself. Trouble then is people could be calling these kind of actions at any point and with `Model::preventLazyLoading(true)` set you end up having to do something like `with(['cartLine.purchasable.prices.currency'])` which can seem a bit much.

Part of me thinks if we're going to try and introduce `Model::preventLazyLoading(true)` it might be worth revisiting pricing and affected area's to see what can be overhauled/improved rather than patching with a bunch of eager loading.

glennjacobs:
@jordyvanderhaegen to support `Model::preventLazyLoading(true);` in v1 I think is going to be pretty difficult/impossible unfortunately. We would need to re-think certain aspects of the system, especially how we handle prices.

For this reason, I'm going to close this PR and will add it to our v2 objectives.

PR #2223, opened 2025-06-24
Remove Laravel 10 support and remove Filament version lock

wychoong:
I suppose this pr https://github.com/stephenjude/filament-two-factor-authentication/pull/61 is to fix the issue
since its been merged, it should be unlocked

also is there a reason for locking other deps also?


glennjacobs:
@wychoong the problem we've had generally is that Filament does an update and then it breaks Lunar. So we wanted to lock the versions and update on a regular basis when we're happy there's no issues in the newer version.

Maybe that's the wrong approach, I don't know. But it seems reasonable to me.

wychoong:
> wanted to lock the versions and update on a regular basis

Ideally this would work, if there is a GA to weekly update dependencies and pin latest working version, not limiting to Filament. 

Pinning specific version will stop everyone from getting deps bug fixes and only lunar team will face the bugs when doing updates. If unlocked, user will hit the bug with their own test and if annoyed enough they can submit a fix to either filament or lunar 😆 or pin to their last working version

glennjacobs:
@wychoong I'm going to drop L10 support, as it's becoming harder to support. 

So I'll take over this PR and do it as part of this.

glennjacobs:
@alecritson can you give this a check to ensure I've not missed anything when removing L10 support?

alecritson (approved):

PR #2225, opened 2025-06-25
Allow discount types to register relation managers

ryanmitchell:
This PR allow custom discount types to register relation managers.

My use case is I want to use the same managers as those used by BuyXGetY to allow rewards and conditions to be associated with the discount.


glennjacobs (changes requested):
@ryanmitchell it seems fine, but as we're technically breaking an interface, could you add something to the upgrade guide to cover it?

ryanmitchell:
@glennjacobs thats been done

glennjacobs (approved):

PR #2247, opened 2025-07-18
Increase delivery_instructions column length in address-related tables

### LINKED ISSUE #2245: Increase `delivery_instructions` Field Length Across Address Tables
- Lunar version: 1.0.0-beta22
- Laravel Version: 11.45.1
- PHP Version: 8.3.23
- Database Driver & Version: MySQL 8.0.42

### Expected Behaviour:
The `delivery_instructions` field in the `lunar_addresses`, `lunar_cart_addresses`, and `lunar_order_addresses` tables should allow users to enter longer messages without hitting character limits.

### Actual Behaviour:
Currently, the `delivery_instructions` field is defined as a `VARCHAR(255)`, which often results in a `Data too long for column` error when users try to provide more detailed instructions for delivery.

### Steps To Reproduce:
1. Attempt to submit an address form (e.g. as a guest) with a longer `delivery_instructions` message (e.g. over 255 characters).
2. Observe SQL exception: `SQLSTATE[22001]: Data too long for column 'delivery_instructions'`.

Huncsuga:
This pull request updates the `delivery_instructions` column in the `lunar_addresses`, `lunar_cart_addresses`, and `lunar_order_addresses` tables from `VARCHAR(255)` to `TEXT`.

This change ensures that users can provide longer delivery instructions without encountering SQL errors such as:

`SQLSTATE[22001]: Data too long for column 'delivery_instructions'`

Fixes #2245

## Summary by CodeRabbit

* **New Features**
  * Increased the maximum length of delivery instructions to 1000 characters for addresses, cart addresses, and order addresses.


wychoong:
> SQLSTATE[22001]: Data too long for column 'delivery_instructions'

There is no need to increase from VARCHAR(255) to TEXT, could be just 500
Or better, your input field should limit input to 255

Huncsuga:
@wychoong Why do you think TEXT wouldn't be need here? Do you see any drawback to using it, or do you think 500 characters would be enough for all future cases?

wychoong:
> Do you see any drawback to using it, 

yes. depends on database engine (read/write performance, storage)

> or do you think 500 characters would be enough for all future cases?

will it be enough for **all future cases**? maybe, maybe not. but it needs to be reasonable.
TEXT support 65,535 characters. is it reasonable for delivery instructions? I don't think so.

refer to your issue
> Attempt to submit an address form (e.g. as a guest) with a longer delivery_instructions message (e.g. over 255 characters).

it's apparent there is no validation and limit on user input. are we going to increase from TEXT 65,535 characters to LONGTEXT? (well, this is exaggerating)

KingaOrsii:
We actually had a user (in production) who entered a longer instruction (over 255 characters) and encountered a server error due to the column size constraint. That’s what prompted us to consider increasing the limit.

If we do decide to go beyond 255, I believe it makes more sense to raise it to at least 1,000 or even 2,000 characters. Increasing it only slightly (e.g., to 500) doesn’t provide much practical benefit, as longer inputs would still be cut off and result in a poor user experience.

So if you suggest enforcing a limit, @wychoong, I’d propose setting it to 2,000 characters.

BUT

Storage-wise, there’s not much difference between VARCHAR(2000) and TEXT for this use case. Both handle ~2000 chars efficiently, and the impact on performance or disk space is minimal. The choice mainly depends on whether we want a fixed length (VARCHAR) or more flexibility (TEXT), and so I think the second option is better.

glennjacobs:
I would suggest we use `VARCHAR(1000)`. There's no real reason for this to be a TEXT field. Varchar will be more optimised for this usage.

Re: the server error - was this on the frontend or the admin panel?

Huncsuga:
Alright, I’ll update it as suggested.
The server error happened on the frontend.

Huncsuga:
@glennjacobs  I’ve made the requested changes.

glennjacobs (approved):

PR #2249, opened 2025-07-20
Update CollectionResource.php

tiznull:
We have over 15,000 collections in the main catalog. The problem with **ancestors** is that it shows the linear collections that are on the same level. This makes the breadcrumb 15,000+ collection names. Using **CHILDREN** only shows the children of the collection that you clicked into. The page loads faster and the title bar with the breadcrumb is more readable.

- Documentation updates
- Automated tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated breadcrumb trail logic to display direct children of a collection instead of its ancestors.

### DIFF
--- a/packages/admin/src/Filament/Resources/CollectionResource.php
+++ b/packages/admin/src/Filament/Resources/CollectionResource.php
@@ -47,7 +47,7 @@ class CollectionResource extends BaseResource
             ]) => $collection->group->name,
         ];
 
-        foreach ($collection->ancestors as $childCollection) {
+        foreach ($collection->children as $childCollection) {
             $crumbs[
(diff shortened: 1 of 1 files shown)


alecritson (approved):

glennjacobs (approved):

wychoong:
@glennjacobs @alecritson I think this change is broken. Breadcrumbs should diplay ancestors not childs


before this PR
image

after
image

image

this is the collection structure tree
image


this PR supposed to fix issue with 15,000+ collections slowness, but I do not understand how is this solving that when the change here is about the nested levels of collection
@tiznull can you please explain how is this fix your issue or can you provide a repo with the 15,000 collection seeder

PR #2250, opened 2025-07-21
Fix missing discountable_type checks in Discount scope methods

fredmarmillod:
## Problem

The `scopeProducts`, `scopeProductVariants`, and `scopeCollections` methods in the Discount model were missing `discountable_type` checks in their `whereDoesntHave('discountables')` subqueries. This could lead to incorrect filtering when discounts have mixed discountable types.

For example, if a discount had Product discountables but a query was looking for Collection-related discounts, the scope might incorrectly include or exclude that discount because it wasn't properly filtering by the specific discountable type.

## Solution

Added `whereDiscountableType()` checks to the `whereDoesntHave('discountables')` subqueries in all three scope methods:
`scopeCollections`: Now properly filters for `Collection::morphName()` discountable types
`scopeProducts`: Now properly filters for `Product::morphName()` discountable types
`scopeProductVariants`: Now properly filters for `ProductVariant::morphName()` discountable types

## Changes

**packages/core/src/Models/Discount.php:**
- Fixed `scopeCollections()` method to include discountable type check
- Fixed `scopeProducts()` method to include discountable type check
- Fixed `scopeProductVariants()` method to include discountable type check

**tests/core/Unit/Models/DiscountTest.php:**
- Added comprehensive test suite for `scopeCollections()`
- Added comprehensive test suite for `scopeProducts()`
- Added comprehensive test suite for `scopeProductVariants()`


fredmarmillod:
This pull request also fixes the scopeCollections. It was previously using the `discountables` relation instead of using the `lunar_collection_discount`table.

I also added a `scopeBrands` that works the same way as `scopeCollections`

glennjacobs:
@ryanmitchell could I have your input on this one?

ryanmitchell:
I understand the issue, but I don't understand why some much of the code has been changed, for example using the prefix. Is there a reason why this PR has gone that far?

For me it should just be a case of adding ->whereDiscounType() on the whereDoesntHave() subqueries.

fredmarmillod:
I started by just adding whereDiscountType on the whereDoesntHave and tested filtering with discounts having multiple limitations of different types but it triggered some mysql ambiguous column names error. That's why I added prefix. Then I noticed thaat brands and collection limitations are using `collection_discount` and `brand_discount` tables instead of the `discountables` table so I updated collection scope to use `collection_discount` and created a brand scope.

ryanmitchell:
hmm, then some of my work on https://github.com/lunarphp/lunar/pull/2214 was incorrect. I had assumed (seemingly incorrectly) that everything was stored in the `discountables` table.

The options are either: 

1. maintain the existing table structure but then `ProductConditionRelationManager` needs to be updated to put collections into collection_discount instead.

2. drop the collection_discount, brand_discount tables and move them to be morph relationships on discountables

To me (2) makes more sense as its already a morph table.

ryanmitchell:
@glennjacobs what approach would you prefer?
@fredmarmillod let me know if you need me to work on these changes

fredmarmillod:
@ryanmitchell Yes, that would be great, thanks

glennjacobs:
> @glennjacobs what approach would you prefer? @fredmarmillod let me know if you need me to work on these changes

I think we stick with the current tables, option 1.

ryanmitchell:
I've opened a PR building on this https://github.com/lunarphp/lunar/pull/2345 - so this one can be closed.

glennjacobs:
Closing in preference of #2345

PR #2255, opened 2025-07-24
Telemetry insights

alecritson:
The idea behind this PR is for Lunar to send anonymous usage insights to a predetermined endpoint. This will help us see how Lunar is being used and at what capacity.

It is completely anonoymous and you will be able to easily opt out of sending any stats by calling `Telemetry::optOut()` in your service provider. We will also only do the call once per day and will be on a deffered request so that it does not interfere with the performance of your storefront.

## Summary by CodeRabbit

* **New Features**
  * Introduced anonymous usage insights enabled by default; data is non-identifying, sent at most once per day, and can be opted out.
* **Documentation**
  * Added clear "Telemetry insights" and upgrade notes with opt-out instructions.
* **Behavior**
  * Telemetry is triggered automatically at app shutdown (when not running in console).
* **Tests**
  * Added comprehensive tests covering opt-out, run frequency, and payload contents.


packages/core/src/Base/TelemetryService.php
```diff
@@ -0,0 +1,79 @@
...
+    public function run(): void
+    {
+        if (! $this->shouldRun()) {
+            return;
+        }
+
+        Cache::forget($this->getCacheKey());
+
+        Cache::remember($this->getCacheKey(), 86400, function (): ?Carbon {
+            $response = Http::withHeader('Accept', 'application/json')
```
glennjacobs, line 73: The timeout for this will be the default Laravel setting of 30 secs.
Do we want to lessen this? And do we want to add any retries?

glennjacobs (approved):

PR #2259, opened 2025-07-25
Update AsAttributeData.php

tiznull:
When indexing data, attribute data can contain optional information that is not a specific field, that currently throws an exception. This fix just steps over these objects and allows the rest of the models to be indexed without stopping.

- Documentation updates
- Automated tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling when processing data fields, reducing the chance of errors if data is missing or incorrectly formatted.

### DIFF
--- a/packages/core/src/Base/Casts/AsAttributeData.php
+++ b/packages/core/src/Base/Casts/AsAttributeData.php
@@ -30,7 +30,7 @@ class AsAttributeData implements Castable
                 $returnData = new Collection;
 
                 foreach ($data as $key => $item) {
-                    if (! class_exists($item['field_type'])) {
+                    if (!is_array($item) || !isset($item['field_type']) || !class_exists($item['field_type'])) {
                         continue;
                     }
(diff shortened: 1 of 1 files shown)


alecritson:
Thanks @tiznull Can you provide a bit more information on how your attribute_data content looks that causes this error?.

tiznull:
Not sure how to track it down. We imported 250,000 products, 1.5 million customers, 450,000 orders. We also use attribute data to store additional information on the products and customers. The data is up to 15 years old. I am sure some of it is off. The imports are just using the Lunar docs for creating records.

alecritson:
Gotcha, I think we need to determine what is actually causing the issue you experienced and whether it's a data issue your side or if it is in fact an issue in Lunar.

tiznull:
You can close this ticket, we have non-usable data in the brands attribute_data column. It did not have the class fields in it. We updated the model to clear out that field.

PR #2270, opened 2025-08-14
Move to JSONB

### LINKED ISSUE #2261: Cannot attach collection from product detail (Postgres Issue)
- Lunar version: 1.0.0-beta25
- Laravel Version: 12.21.0
- PHP Version: 8.3.24
- Database Driver & Version: Postgresql v 16.2

### Expected Behaviour:

Each time I add a collection from a product detail, I get an error like this:

```
Undefined function: 7 ERROR:  could not identify an equality operator for type json
LINE 1: select distinct "lunar_collections".*, "lunar_collection_pro...
                        ^ (Connection: pgsql, SQL: select distinct "lunar_collections".*, "lunar_collection_product"."position" from "lunar_collections" left join "lunar_collection_product" on "lunar_collections"."id" = "lunar_collection_product"."collection_id" where "lunar_collections"."id" = 1 order by "lunar_collection_product"."position" asc limit 1)

SQLSTATE[42883]: Undefined function: 7 ERROR:  could not identify an equality operator for type json
```

I think the culprit is come the `attribute_data` column in `lunar_collections` table. I actually want to try to repair but I'm clueless with Filament. I can spot the source come from this file: https://github.com/lunarphp/lunar/blob/1.x/packages/admin/src/Filament/Resources/ProductResource/Pages/ManageProductCollections.php#L62

### Actual Behaviour:

I can search the collection from product detail and add the collection into the product detail smoothly.

### Steps To Reproduce:

1. Go to product detail
2. Go to collections
3. Create attach
4. Search collection then attach and you will see the error

https://github.com/user-attachments/assets/f351067d-bd90-4367-b6f8-5be34e2102b5

glennjacobs:
Closes #2261 

We were using standard JSON fields and when it comes to Postgres this field type is very basic, not much more than a simple text field.

JSONB is a "proper" JSON field, is supported properly in Filament and should perform a lot better.

It's worth noting MySQL & Sqlite will continue to use the same JSON field type - so no change there.

## Summary by CodeRabbit

- **Refactor**
  - Converted many metadata and attribute fields to PostgreSQL JSONB for improved performance and consistency across activity logs, addresses, attributes, brands, carts, collections, customers, discounts, orders, products, shipping, transactions, and media.

- **Chores**
  - Added a PostgreSQL-only migration to perform the conversions and raised the PostgreSQL minimum requirement from 9.2+ to 9.4+.

### DIFF
--- a/packages/core/database/migrations/2025_08_14_164000_switch_to_jsonb.php
+++ b/packages/core/database/migrations/2025_08_14_164000_switch_to_jsonb.php
@@ -2,8 +2,8 @@
 
 use Illuminate\Database\Migrations\Migration;
 use Illuminate\Database\Schema\Blueprint;
-use Illuminate\Support\Facades\Schema;
 use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Facades\Schema;
 
 return new class extends Migration
 {
--- a/packages/core/database/migrations/2025_08_14_164000_switch_to_jsonb.php
+++ b/packages/core/database/migrations/2025_08_14_164000_switch_to_jsonb.php
@@ -152,6 +152,7 @@ return new class extends Migration
 
         // Get the Lunar table prefix from config
         $prefix = config('lunar.database.table_prefix', 'lunar_');
-        return $prefix . $table;
+
+        return $prefix.$table;
     }
 };
--- a/packages/core/database/migrations/2025_08_14_164000_switch_to_jsonb.php
+++ b/packages/core/database/migrations/2025_08_14_164000_switch_to_jsonb.php
@@ -160,9 +160,7 @@ return new class extends Migration
             return $table;
         }
 
-        // Get the Lunar table prefix from config
-        $prefix = config('lunar.database.table_prefix', 'lunar_');
-
-        return $prefix.$table;
+        
+        return $this->prefix.$table;
     }
 };
--- a/packages/core/database/migrations/2025_08_14_164000_switch_to_jsonb.php
+++ b/packages/core/database/migrations/2025_08_14_164000_switch_to_jsonb.php
@@ -1,9 +1,9 @@
-
 <?php
 
 use Illuminate\Database\Migrations\Migration;
 use Illuminate\Database\Schema\Blueprint;
 use Illuminate\Support\Facades\Schema;
+use Illuminate\Support\Facades\DB;
 
 return new class extends Migration
 {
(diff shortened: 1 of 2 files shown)


packages/core/database/migrations/2025_08_14_164000_switch_to_jsonb.php
```diff
@@ -0,0 +1,168 @@
...
+        $nonLunarTables = ['activity_log', 'media'];
+
+        if (in_array($table, $nonLunarTables)) {
+            return $table;
+        }
+
+        // Get the Lunar table prefix from config
+        $prefix = config('lunar.database.table_prefix', 'lunar_');
+
+        return $prefix.$table;
```
wychoong:
```suggestion
        
        return $this->prefix.$table;
```

packages/core/database/migrations/2025_08_14_164000_switch_to_jsonb.php
```diff
@@ -0,0 +1,168 @@
...
+        }
+
+        foreach ($this->columnsToUpdate as $table => $columns) {
+            $fullTableName = $this->getTableName($table);
+
+            Schema::table($fullTableName, function (Blueprint $tableBlueprint) use ($columns) {
+                foreach ($columns as $column) {
+                    $columnBuilder = $tableBlueprint->jsonb($column['name']);
+
+                    if ($column['nullable']) {
```
alecritson: Couldn't this just simply be `$columnBuilder->nullable($column['nullable'])` ?

packages/core/database/migrations/2025_08_14_164000_switch_to_jsonb.php
```diff
@@ -0,0 +1,168 @@
...
+        }
+
+        foreach ($this->columnsToUpdate as $table => $columns) {
+            $fullTableName = $this->getTableName($table);
+
+            Schema::table($fullTableName, function (Blueprint $tableBlueprint) use ($columns) {
+                foreach ($columns as $column) {
+                    $columnBuilder = $tableBlueprint->json($column['name']);
+
+                    if ($column['nullable']) {
```
alecritson: Same here

packages/core/database/migrations/2025_08_14_164000_switch_to_jsonb.php
```diff
@@ -0,0 +1,153 @@
...
+                        ->change();
+                }
+            });
+        }
+    }
+
+    /**
+     * Get the full table name with appropriate prefix.
+     */
+    private function getTableName(string $table): string
```
alecritson, line 142: This can probably be simplified to avoid unnecessary if statements and returns:

```php
private function getTableName(string $table): string
{
    return in_array($table, ['activity_log', 'media']) ? $table : $this->prefix.$table;
}
```

alecritson (approved):

Issue #2265, opened 2025-08-12
Order Invoice PDF - No publishable resource


alecgarcia:
- Lunar version: 1.0.0-beta25
- Laravel Version: 11.45.1
- PHP Version: 8.4.11
- Database Driver & Version: MySQL - 8.0.32

### Expected Behaviour:
When following documentation on how to publish the view for the `Order Invoice PDF` it says to run the following command: `php artisan vendor:publish --tag=lunar.hub.views`

This results in a response of: `INFO  No publishable resources for tag [lunar.hub.views].`

Also the documentation states that the file should be in the following path but is wrong and needs to be updated as well. 
`resources/vendor/lunarpanel/pdf/order.blade.php`

### Actual Behaviour:
The command should publish a view to the following path: `resources/view/vendor/lunarpanel/pdf/order.blade.php`

### Steps To Reproduce:
Run Command: `php artisan vendor:publish --tag=lunar.hub.views`

PR #2281, opened 2025-08-19
Fix vendor publish command for pdfs

SecurID:
This will fix issue #2265 

I am not sure about the naming of the publish tag. 
I think lunar.pdf fits better as this will only target all pdfs currently (and in the future) in lunar. 

If accepted we need to edit the docs aswell. 

Docs: https://github.com/lunarphp/docs/pull/3


SecurID:
Just saw that the docs are in the same repo so I updated them aswell.

PR #2283, opened 2025-08-20
:bug: fix(console): switch to components->info()

chrillep:
components->line requires two arguments in Laravel 12, causing an ArgumentCountError with a single string. Using components->info preserves formatting and restores compatibility. The $this->line helper has a different API and is not interchangeable with components->line.


glennjacobs:
@alecritson I don't think we've revisited the addons manifest for a good while. I don't think it's being used. 

Guessing we left it in for when we *did* want to re-implement it?

chrillep:
> @alecritson I don't think we've revisited the addons manifest for a good while. I don't think it's being used.
> 
> Guessing we left it in for when we _did_ want to re-implement it?

to be fair i haven't checked the internals/usage of the command, i just ran it and it failed. So i fixed that :)

glennjacobs (approved):

PR #2287, opened 2025-08-25
fix: use logical and operator instead of binary

hosni:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where existing cart line items could be misidentified when no changes were present, leading to unintended merges or duplicates.
  * Cart lines are now matched only when there are truly no new, edited, or removed differences, improving the accuracy of item updates.
  * Results in more predictable cart behavior and smoother checkout consistency.


alecritson (approved):

hosni:
Hi @alecritson,
Could you please re-run the tests? One of them failed because it tries to create a country with code GB, which already exists in the DB.
I ran the tests locally and they all passed. The CI also passed all the other tests, so this one might be failing due to some interaction with another test, but I’m not sure why.

PR #2290, opened 2025-09-05
Add with trashed to orderline purchasable relation

alecritson:
Sometimes a purchasable might be soft deleted which causes errors in the panel when viewing an order. This PR looks to include the `withTrashed()` scope so that these orders can load. This shouldn't have any adverse effects on how orders work.

## Summary by CodeRabbit

* **Bug Fixes**
  * Order lines now retain references to associated items even if those items have been archived or deleted. This ensures historical orders, invoices, and reports display complete information without missing products or errors.
  * Improves consistency across order detail pages, admin views, and data exports by preserving access to product details for past purchases, aiding auditing and customer support scenarios.

### DIFF
--- a/packages/core/src/Models/OrderLine.php
+++ b/packages/core/src/Models/OrderLine.php
@@ -82,7 +82,7 @@ class OrderLine extends BaseModel implements Contracts\OrderLine
 
     public function purchasable(): MorphTo
     {
-        return $this->morphTo();
+        return $this->morphTo()->withTrashed();
     }
 
     public function currency(): HasOneThrough


glennjacobs:
@alecritson do you think this is the correct approach? If something has been deleted, then I guess it's for a reason.

Perhaps instead the link shouldn't be made available for deleted purchasables, or purchasables that don't provide a link?

alecritson:
> @alecritson do you think this is the correct approach? If something has been deleted, then I guess it's for a reason.
> 
> Perhaps instead the link shouldn't be made available for deleted purchasables, or purchasables that don't provide a link?

Yeah it usually is deleted for a reason but I think it's still valid for an order line to be able to relate to that purchasable. Having the link to the purchasable runs deeper in the panel than just providing a url:

- Thumbnail display
- Current stock level
- Variant options

Looks like we currently use  the null safe operator for the above but not the link but I think the information it returns is still valid for that order (assuming we don't store it on the order line itself). I guess if you don't provide a link to the deleted product/purchasable does it then become more of a chore to then go and find that product to reactivate it if you need to?

glennjacobs:
But it's deleted, so it can't have any stock and the values would be assumed to be old. It seems dangerous to use such data.

PR #2296, opened 2025-09-23
Update logic to clean up order lines

alecritson:
This PR adds additional logic to the `CleanUpOrderLines` pipeline to catch any duplicate purchasable lines which may have slipped through the net due to differing quantities/meta values.

## Summary by CodeRabbit

* **Bug Fixes**
  * Order line cleanup now uses per-line signatures to precisely compare against cart lines, removing only truly mismatched lines and preserving matching product and shipping lines.

* **Tests**
  * Added and expanded unit tests to validate persisted DB state (including same product with different metadata/prices) and replaced fragile in-memory checks with robust database assertions.

### DIFF
--- a/packages/core/src/Pipelines/Order/Creation/CleanUpOrderLines.php
+++ b/packages/core/src/Pipelines/Order/Creation/CleanUpOrderLines.php
@@ -3,6 +3,7 @@
 namespace Lunar\Pipelines\Order\Creation;
 
 use Closure;
+use Illuminate\Database\Eloquent\Casts\ArrayObject;
 use Lunar\Models\Contracts\Order as OrderContract;
 use Lunar\Models\Order;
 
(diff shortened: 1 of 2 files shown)


glennjacobs (approved):

packages/core/src/Pipelines/Order/Creation/CleanUpOrderLines.php
```diff
@@ -16,12 +16,32 @@ public function handle(OrderContract $order, Closure $next): mixed
         /** @var Order $order */
         $cart = $order->cart;
 
-        $purchasableIds = $cart->lines->pluck('purchasable_id');
+        // Build a set of "signatures" that uniquely identify each cart line
+        $cartSignatures = $cart->lines->map(function ($line) {
+            return $this->signature($line->purchasable_id, (array) $line->meta, $line->quantity);
```
wychoong, line 21: should include purchasable_type in signature too

glennjacobs: @alecritson can you do a PR to add that in? I think it makes sense.

PR #2305, opened 2025-10-02
fix scout search with soft deleted resource

wychoong:
To use Scout search with soft delete model, as per laravel doc

> set the soft_delete option of the config/scout.php configuration file to **true**:
> 
> 'soft_delete' => true,
>
> When this configuration option is true, Scout will not remove soft deleted models from the search index. Instead, it will set a hidden __soft_deleted attribute on the indexed record. Then, you may use the withTrashed or onlyTrashed methods to retrieve the soft deleted records when searching:

and lunar also mentioned this in the doc
> By default, scout has the setting soft_delete set to false. You need to make sure this is set to **true** otherwise you will see soft deleted models appear in your search results

however this doesnt work when searching in the panel with filament table. when set to true, the scout search will not return the records, regardless the table filter already selected "with trashed" or "only trashed"
this PR attemp to fix this

~p/s: this is just a quick fix, it might have a better way to detect is the with/only trashed filter is applied before using `->with/onlyTrashed` on the scout builder, to get a better search result.~


packages/admin/src/Support/Pages/BaseListRecords.php
```diff
@@ -32,7 +32,16 @@ protected function applySearchToTableQuery(Builder $query): Builder
...
         ) {
-            $ids = collect(static::getModel()::search($search)->take(100)->keys())->map(
+            $trashedFilter = collect($this->getTable()->getFilters())
+                ->firstWhere(fn ($filter) => $filter instanceof \Filament\Tables\Filters\TrashedFilter);
+
+            $scoutQuery = static::getModel()::search($search);
+
+            if (filled($state = $trashedFilter?->getState()['value'] ?? null)) {
+                $state ? $scoutQuery->withTrashed() : $scoutQuery->onlyTrashed();
+            }
```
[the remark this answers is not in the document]
wychoong: bot is wrong here.
check TrashedFilter for this. while the filter is apply for eloquent builder, the same should apply to Scout builder
https://github.com/filamentphp/filament/blob/3.x/packages/tables/src/Filters/TrashedFilter.php

packages/admin/src/Support/Resources/BaseResource.php
```diff
@@ -74,22 +74,13 @@ protected static function applyGlobalSearchAttributeConstraints(Builder $query,
             $scoutEnabled &&
             $isScoutSearchable
         ) {
-            $ids = collect(static::getModel()::search($search)->keys())->map(
+            $ids = collect(static::getModel()::search($search)->withTrashed()->keys())->map(
```
wychoong: if search using global search in the panel should not show the soft deleted records, this line can be reverted, and also `getGlobalSearchEloquentQuery` in packages/admin/src/Filament/Resources/ProductResource.php

glennjacobs: In v2 I want to introduce some additional states for things like Products. This should mean that anything that is soft deleted probably shouldn't display in the global search.

Could you update to suit?

wychoong: done

packages/admin/src/Support/Pages/BaseListRecords.php
```diff
@@ -32,21 +32,22 @@ protected function applySearchToTableQuery(Builder $query): Builder
...
 
-            $placeholders = implode(',', array_fill(0, count($ids), '?'));
+            $scoutQuery = static::getModel()::search($search);
 
-            $query->whereIn(
-                'id',
-                $ids
-            );
+            if (filled($state = $trashedFilter?->getState()['value'] ?? null)) {
+                $state ? $scoutQuery->withTrashed() : $scoutQuery->onlyTrashed();
```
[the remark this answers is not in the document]
wychoong: bot is wrong, just like previous coderabbit review

packages/admin/src/Support/Pages/BaseListRecords.php
```diff
@@ -39,21 +40,22 @@ protected function applySearchToTableQuery(Builder $query): Builder
...
-            $query->when(
-                ! $ids->isEmpty(),
-                fn ($query) => $query->orderBySequence($ids->toArray())
+            $ids = collect($scoutQuery->take(100)->keys())->map(
+                fn ($result) => str_replace(static::getModel().'::', '', $result)
             );
+
+            $query
+                ->whereIn('id', $ids)
+                ->orderBySequence($ids);
```
alecritson, line 58: Having another look at this with fresh eyes. I think we still need to call `->toArray()` on the ids as this will apparently fail on MariaDB

wychoong: I’ll have another look at it. 

Any specific errors you have in mind? Passing collection to query builder shouldn’t be an issue

alecritson: Nevermind, tested with MariaDB and the error that was supposedly going to bubble up..didn't.

alecritson (approved):

PR #2306, opened 2025-10-03
Fix: Prevents fatal errors for custom Purchasable models

ablunier:
This PR improves support for custom models that implement the Purchasable interface but are not standard ProductVariants.

Currently, using a custom purchasable can lead to two exceptions:
1. DiscountManager error: It assumes the purchasable always has a ->product->collections relationship, causing a null pointer exception if it doesn't.
2. Admin Panel error: The order details view assumes the purchasable has a getOptions() method, which is specific to ProductVariant.

This PR resolves these issues with minimal impact:
1. The DiscountManager call chain is made null-safe to gracefully handle purchasables without product collections.
2. The Purchasable interface now includes getOptions().

This makes the Purchasable interface more robust and truly extensible as intended. Any feedback is welcome.


glennjacobs:
@ablunier it looks like we have failing tests. Are you able to take a look? Thanks.

ablunier:
Tests are now green @glennjacobs
Sorry!

glennjacobs (approved):

glennjacobs:
Thanks @ablunier

PR #2314, opened 2025-10-15
Fix admin activity log feed rendering

shanebrightbulb:
This PR fixes 2 issues when rendering the 'Activity Log' component on the order screen in the admin. 

The first issue is visible when using Filament Notifications. The z-index of the comment form in the activity log is set to 20 which brings it in front of the Filament notifications sidebar when it is opened.

The second issue is when using dark mode. The date heading in the activity log is using the 'text-gray-300' Tailwind class, however this class isn't generated so the 'date' heading in the activity log is black and cannot be read.


packages/admin/resources/views/livewire/components/activity-log-feed.blade.php
```diff
@@ -26,7 +26,7 @@ class="inline-block w-8 h-8 rounded-full" />
                 role="list">
                 @foreach ($this->activityLog as $log)
                     <li class="relative py-8 ml-5">
-                        <p class="ml-8 font-bold text-gray-950 dark:text-gray-300">
+                        <p class="ml-8 font-bold text-gray-950 dark:text-gray-400">
```
wychoong: this is not the correct fix. probably is the css build is not updated

packages/admin/resources/views/livewire/components/activity-log-feed.blade.php
```diff
@@ -1,5 +1,5 @@
 <div class="px-2 pb-4 scroll-mt-32" id="lunar-panel-timeline">
-    <div class="relative flex items-end gap-4 mt-4 z-20">
+    <div class="relative flex items-end gap-4 mt-4 z-10">
```
wychoong: while this fix the issue but I dont think its the root cause. this fix essentially means in the panel we cannot use any z-index more than 10

glennjacobs: Do we even need a z-index on this? I removed it and it was fine 🤷‍♂️

Issue #2316, opened 2025-10-21
Missing translation support for EditShippingZone navigation label


VincentTeresa:
## Context
I'm working on a Spanish e-commerce site using Lunar with the table-rate-shipping plugin. I've successfully translated most of the admin panel to Spanish, but I've encountered an issue where one specific navigation label cannot be translated.

## What I'm Trying to Do
I need to translate the "Edit Shipping Zone" text that appears in the sub-navigation menu when editing a shipping zone. I've already successfully translated all other navigation items using the standard Laravel translation files in `lang/vendor/lunarpanel.shipping/es/`.

## The Problem
The `EditShippingZone` page class doesn't implement a `getNavigationLabel()` method, unlike its sibling pages. This causes Filament to auto-generate the label from the class name ("Edit Shipping Zone"), making translation impossible.

## What I've Checked
- Reviewed the documentation for translation support
- Checked existing issues for similar problems
- Verified that other pages in the same module work correctly
- Confirmed the issue exists in the latest version
- Attempted workarounds (extending classes, translation overrides)

## Steps to Reproduce
1. Install Lunar with table-rate-shipping plugin
2. Set application locale to any non-English language (e.g., `'locale' => 'es'`)
3. Add translations in `lang/vendor/lunarpanel.shipping/es/`
4. Navigate to `/admin/shipping-zones/{id}/rates`
5. Observe the sub-navigation menu

**Current Result:**
- "Edit Shipping Zone" (Not translated - hardcoded from class name)
- "Tarifas de envío" (Correctly translated)
- "Exclusiones de envío" (Correctly translated)

**Expected Result:**
All three navigation items should be translated.

### Affected File
`src/Filament/Resources/ShippingZoneResource/Pages/EditShippingZone.php`

### Working Example (from same package)
`ManageShippingRates.php` correctly implements translation:
```php
public static function getNavigationLabel(): string
{
    return __('lunarpanel.shipping::relationmanagers.shipping_rates.title_plural');
}
```

### Proposed Solution
Add the same pattern to `EditShippingZone.php`:
```php
public static function getNavigationLabel(): string
{
    return __('lunarpanel.shipping::shippingzone.pages.edit.label');
}
```

## Environment
- **Laravel:** 11.46.1
- **Lunar:** Latest version
- **Lunar Table Rate Shipping:** Latest version
- **PHP:** 8.2+
- **Filament:** v3

## Why This Matters
[...]

PR #2317, opened 2025-10-21
Add translation support for EditShippingZone navigation label

VincentTeresa:
This PR adds the missing getNavigationLabel() method to EditShippingZone to maintain consistency with other pages in the module (ManageShippingRates and ManageShippingExclusions). Fixes issue #2316


alecritson:
Thanks, when adding translations we need to ensure we add the English version at a minimum, if you're able to provide the other translations (even via a service like Google Translate) it would help as well. :)

VincentTeresa:
Of course! I already did it! Thanks for the heads-up.

glennjacobs:
Thanks @VincentTeresa

PR #2319, opened 2025-10-23
Add semicolon to example code

erashdan:
The example code has invalid syntax because it’s missing a semicolon.

- Documentation updates
https://docs.lunarphp.io/core/extending/payments.html


glennjacobs:
Thanks, I'm actually releasing updated docs very soon, so will ensure this fix is included.

glennjacobs:
I've updated in the new docs here https://github.com/lunarphp/docs

I'll be removing the old docs shortly.

PR #2326, opened 2025-10-31
Fix typo in extend method name for order summary infolist

wychoong:
could be a breaking change

Docs: https://github.com/lunarphp/docs/pull/1


alecritson:
I think it should be fine, although I think something in the upgrade guides would help anyone who is using the misspelt reference in their extensions.

glennjacobs (approved):

PR #2330, opened 2025-11-03
Fix default color and icon for state in CustomerGroupRelationManager

ChaDonSom:
Add support for when postgres casts booleans to php bool, thereby casting `false` to an empty string for the match case

- Documentation updates
- Automated tests

### DIFF
--- a/packages/admin/src/Filament/Resources/ProductResource/RelationManagers/CustomerGroupRelationManager.php
+++ b/packages/admin/src/Filament/Resources/ProductResource/RelationManagers/CustomerGroupRelationManager.php
@@ -81,15 +81,8 @@ class CustomerGroupRelationManager extends BaseRelationManager
             return Tables\Columns\IconColumn::make($column)->label(
                 __("lunarpanel::relationmanagers.customer_groups.table.{$column}.label")
             )
-                ->color(fn (string $state): string => match ($state) {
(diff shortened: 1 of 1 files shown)


packages/admin/src/Filament/Resources/ProductResource/RelationManagers/CustomerGroupRelationManager.php
```diff
@@ -84,9 +84,11 @@ public function getDefaultTable(Table $table): Table
                 ->color(fn (string $state): string => match ($state) {
```
wychoong: if the only way for this to work have to be loose check, instead of using `match` which is strict === checking, just use `$state ? 'success' : 'warning'`?

ChaDonSom: addressed in commit https://github.com/lunarphp/lunar/pull/2330/commits/af91802254b9e648fbcfa3633ed857ffe9b65358

Apparently I was wrong about something: I thought string "0" would evaluate truthy in PHP! When I cast it to (bool) "0" though, in Tinkerwell, it shows false.

I'm not sure whether I should be putting $state as `bool $state` in the arrow function's argument, though.

glennjacobs (approved):

PR #2346, opened 2025-11-12
Add meta to product options and values

glennjacobs:
This adds meta to Product Options & Product Option Values to support customisation, e.g. assigning hex colours.

It also updates the model annotations and corrects the `name` cast on ProductOptionValue (this might be considered a breaking change...)

## TODO

- [ ] Add docs PR, including upgrade guide entry

### DIFF
--- a/packages/core/src/Models/ProductOption.php
+++ b/packages/core/src/Models/ProductOption.php
@@ -18,10 +18,12 @@ use Spatie\MediaLibrary\HasMedia as SpatieHasMedia;
 
 /**
  * @property int $id
- * @property \Illuminate\Support\Collection $name
- * @property \Illuminate\Support\Collection $label
+ * @property AsArrayObject $name
+ * @property ?AsArrayObject $label
  * @property int $position
  * @property ?string $handle
+ * @property bool $shared
+ * @property ?AsArrayObject $meta
  * @property ?\Illuminate\Support\Carbon $created_at
  * @property ?\Illuminate\Support\Carbon $updated_at
  */
@@ -43,6 +45,7 @@ class ProductOption extends BaseModel implements Contracts\ProductOption, Spatie
         'name' => AsArrayObject::class,
         'label' => AsArrayObject::class,
         'shared' => 'boolean',
+        'meta' => AsArrayObject::class,
     ];
 
     /**
--- a/packages/core/src/Models/ProductOptionValue.php
+++ b/packages/core/src/Models/ProductOptionValue.php
@@ -2,7 +2,7 @@
 
 namespace Lunar\Models;
 
-use Illuminate\Database\Eloquent\Casts\AsCollection;
+use Illuminate\Database\Eloquent\Casts\AsArrayObject;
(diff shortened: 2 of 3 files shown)


packages/core/src/Models/ProductOptionValue.php
```diff
@@ -16,8 +16,9 @@
 /**
  * @property int $id
  * @property int $product_option_id
- * @property string $name
+ * @property AsArrayObject $name
```
wychoong, line 19: do you think it makes sense to add a `handle` field?

packages/core/src/Models/ProductOption.php
```diff
@@ -43,6 +45,7 @@ class ProductOption extends BaseModel implements Contracts\ProductOption, Spatie
         'name' => AsArrayObject::class,
         'label' => AsArrayObject::class,
         'shared' => 'boolean',
+        'meta' => AsArrayObject::class,
```
wychoong, line 48: isn't using the `attribute_data` would be a better way to extend, more inline to how Product fields are extended currently?

packages/core/src/Models/ProductOptionValue.php
```diff
@@ -34,7 +35,8 @@ class ProductOptionValue extends BaseModel implements Contracts\ProductOptionVal
      * @var array
      */
     protected $casts = [
-        'name' => AsCollection::class,
+        'name' => AsArrayObject::class,
```
wychoong, line 38: likely not a breaking change since AsCollection only allow `name->en` while AsArrayObject allows `name->en` or `name['en']`

packages/core/src/Models/ProductOptionValue.php
wychoong: also if I remember correctly, there was a discussion whether to standardized `name` field vs name in `attribute_data` like Products.

I think my question is out of the scope as this current change is just phpdoc, not any structural change...

packages/core/src/Models/ProductOption.php
glennjacobs: I don't believe so. Attribute data would mean all values would have the same fields. I don't want to add a hex colour for a size value for example, so then we'd have to add Product Option Types - it starts to get quite messy. I don't think Lunar should be too opinionated here. We should leave it to add-on developers using the meta field.

packages/core/src/Models/ProductOptionValue.php
glennjacobs: > do you think it makes sense to add a `handle` field?

There isn't a handle currently. Adding one would be outside the scope of this PR.

glennjacobs: > also if I remember correctly, there was a discussion whether to standardized `name` field vs name in `attribute_data` like Products.
> 
> I think my question is out of the scope as this current change is just phpdoc, not any structural change...

The intention in v2 is to use dedicated name fields like this one, rather than in `attribute_data`. Using a dedicated field makes it a lot easier to search, etc.

alecritson (approved):

PR #2349, opened 2025-11-14
Add cart_session config option to allow for the prevention of the cart deletion on user logout

charlielangridge:
Added 'delete_on_logout' config option within cart_session.php config file to allow the prevention of the cart deletion when a user logs out

### DIFF
--- a/packages/core/src/Listeners/CartSessionAuthListener.php
+++ b/packages/core/src/Listeners/CartSessionAuthListener.php
@@ -61,6 +61,6 @@ class CartSessionAuthListener
             return;
         }
 
-        CartSession::forget(config('lunar.cart_session.delete_on_logout', true));
+        CartSession::forget();
     }
 }
--- a/packages/core/src/Listeners/CartSessionAuthListener.php
+++ b/packages/core/src/Listeners/CartSessionAuthListener.php
@@ -61,6 +61,6 @@ class CartSessionAuthListener
             return;
         }
 
-        CartSession::forget();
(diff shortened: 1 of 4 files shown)


packages/core/src/Listeners/CartSessionAuthListener.php
```diff
@@ -61,6 +61,6 @@ public function logout(Logout $event)
             return;
         }
 
-        CartSession::forget();
```
alecritson, line 64: I feel like this should be on the `CartSessionManager` otherwise this config option has no effect outside of the listener.

If we make `$delete` an optional argument then if it's not passed we can check what to do by default using the config.

glennjacobs: Why would it be needed outside of the listener?

alecritson: > Why would it be needed outside of the listener?

If on my storefront I want to call `CartSession::forget()` anywhere, it won't use the config.

glennjacobs: OK, let's change this to `delete_on_forget`

alecritson:
@charlielangridge We're going to change this slightly to `delete_on_forget` and handle the config value on the `forget` call of the `CartSessionManager` I tried to update your PR but was unable to due to permissions, if you're happy to make the changes then feel free, otherwise close the PR so I know to take over :)

alecritson (approved):

PR #2353, opened 2025-11-20
Fix table prefix duplication in multi-level model inheritance

somegooser:
When extending a model multiple times (e.g., CustomProduct extends Product extends BaseModel), the table prefix was being applied multiple times, resulting in table names like "lunar_lunar_products" instead of "lunar_products".

**Problem:**
The `HasModelExtending::getTable()` method instantiated parent classes recursively with `(new $parentClass)->table`, causing the BaseModel constructor to apply the prefix at each inheritance level.

**Solution:**
- Added `resolveRootModelClass()` to find the root Lunar model in the inheritance chain
- Added `resolveBaseTableName()` to get the base table name using reflection without triggering the constructor
- Modified `getTable()` to resolve the root model and apply the prefix only once

**Test Coverage:**
Added test case for multi-level inheritance to ensure prefix is not duplicated across multiple levels of model extension.

### DIFF
--- a/packages/core/src/Base/Traits/HasModelExtending.php
+++ b/packages/core/src/Base/Traits/HasModelExtending.php
@@ -9,6 +9,7 @@ use Illuminate\Support\Arr;
 use Illuminate\Support\Str;
 use Lunar\Base\BaseModel;
 use Lunar\Facades\ModelManifest;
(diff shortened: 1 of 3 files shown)


ahmetakdas:
Hi.. any update about this?

PR #2357, opened 2025-11-26
Allow product association types to be customised

### LINKED ISSUE #2337: Allow for custom product association types
### Discussed in https://github.com/lunarphp/lunar/discussions/2163

<div type='discussions-op-text'>

<sup>Originally posted by **Platypuscoder** April 11, 2025</sup>
Currently there are 3 types of associations in the lunar admin tool (alternate, cross-sell and up-sell). For instance, we sell t-shirts small, medium and large but having an add-on would be for example a 'gift box' that would increase the total price of the t shirt.</div>

alecritson:
PR to allow product association types to be customised.

- Deprecates the use of constants like `UP_SELL` on the model
- Introduces a new enum to handle association types
- Allow the enum to be replaced in config by developers.

Closes #2337

### DIFF
--- a/packages/core/src/Jobs/Products/Associations/Associate.php
+++ b/packages/core/src/Jobs/Products/Associations/Associate.php
@@ -69,7 +69,7 @@ class Associate implements ShouldQueue
                 $this->targets->map(function ($model) {
                     return [
                         'product_target_id' => $model->id,
-                        'type' => $this->type->value,
+                        'type' => $this->type?->value,
                     ];
                 })
             );
--- a/packages/core/src/Jobs/Products/Associations/Associate.php
+++ b/packages/core/src/Jobs/Products/Associations/Associate.php
@@ -37,12 +37,12 @@ class Associate implements ShouldQueue
     /**
      * The product association type.
      */
-    protected ?ProvidesProductAssociationType $type = null;
+    protected ProvidesProductAssociationType $type;
 
     /**
      * Create a new job instance.
      */
-    public function __construct(ProductContract $product, mixed $targets, ?ProvidesProductAssociationType $type = null)
+    public function __construct(ProductContract $product, mixed $targets, ProvidesProductAssociationType $type)
     {
         if (is_array($targets)) {
             $targets = collect($targets);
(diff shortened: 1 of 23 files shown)


glennjacobs:
@alecritson generally looks good, but I think we need to consider translations.

alecritson:
> @alecritson generally looks good, but I think we need to consider translations.

Added some translations

packages/core/src/Models/ProductAssociation.php
```diff
@@ -77,30 +88,45 @@ public function target(): BelongsTo
      */
     public function scopeCrossSell(Builder $query): Builder
     {
-        return $query->type(self::CROSS_SELL);
+        return $query->type(ProductAssociationEnum::CROSS_SELL->value);
```
wychoong:
```suggestion
        return $query->type(ProductAssociationEnum::CROSS_SELL);
```

packages/core/src/Models/Product.php
wychoong, line 84: can this use?
```php
protected function casts():array{
    return [
        'type' => config('lunar.products.association_types_enum', \Lunar\Base\Enums\ProductAssociation::class),
        ...
    ];
}
```

packages/core/src/Jobs/Products/Associations/Associate.php
```diff
@@ -73,7 +69,7 @@ public function handle()
                 $this->targets->map(function ($model) {
                     return [
                         'product_target_id' => $model->id,
-                        'type' => $this->type,
+                        'type' => $this->type->value,
```
wychoong: the propertu $type is nullable but its not cater
```suggestion
                        'type' => $this->type?->value,
```
but can it be null?

wychoong: maybe mix up with Dissociate job

packages/core/src/Models/Product.php
```diff
@@ -144,15 +145,15 @@ public function inverseAssociations(): HasMany
         return $this->hasMany(ProductAssociation::modelClass(), 'product_target_id');
     }
 
-    public function associate(mixed $product, string $type): void
+    public function associate(mixed $product, ProvidesProductAssociationType $type): void
     {
         Associate::dispatch($this, $product, $type);
     }
```
wychoong: i think this method and the job should backward compatible with `string`
and handle enum->value
also the constant is only deprecated so string should still be supported.

glennjacobs (approved):

PR #2368, opened 2025-12-09
Add ability to skip Lunar event forwarding for specific events

somegooser:
When extending Lunar models, events are fired for both the custom class and the Lunar base class. This can cause issues when you want to override event behavior in your custom model.

This change adds a $skipLunarEventForwarding property that allows developers to specify which events should NOT be forwarded to the Lunar base model.

Usage:

class Product extends LunarProduct
{
    protected array $skipLunarEventForwarding = ["created", "updated"];
}

This will fire the event only for the custom class, not for Lunar\Models\Product.


somegooser:
I dont know why the tests fails.. did not change any other files.

alecritson:
@somegooser I appreciate this PR is old, we're slowly getting through a backlog of tasks. If you're able to add a test for this that would be appreciated.

glennjacobs:
Thanks for the PR. After review we don't think this is the right shape for Lunar:

- The Lunar→base event forwarding is load-bearing — Lunar's own observers (Searchable, etc.) listen on the base contract class. A skip list silently breaks those for any user who reaches for it, with no warning.
- The use cases described here are already covered by Eloquent's existing tools: observers returning false to halt, saveQuietly() / withoutEvents(), or simply registering listeners on the extended class instead of both.
- We'd want this kind of extension point to be a named method (e.g. shouldForwardLunarEvent($event): bool), documented on the "Extending Models" page, with tests — and ideally driven by a concrete problem that can't be solved today.

If you can share the original scenario that motivated this, happy to reopen the discussion. Closing for now.

PR #2382, opened 2025-12-14
Fix weight-based shipping rate breakpoints evaluation

### LINKED ISSUE #2381: Weight-based Shipping Rates ignore pricing breakpoints due to inconsistent weight unit handling
- Lunar version: 1.2.1
- Laravel Version: 12.42.0
- PHP Version: 8.3.28
- Database Driver & Version: MySQL 8.0.37

### Expected Behaviour:
When a Shipping Method is configured with **Charge By: Weight**, the correct pricing break (shipping tier) should be selected based on the cart’s total weight. As the cart weight increases and crosses defined breakpoints, the corresponding Shipping Breakdown price should be applied.

### Actual Behaviour:
The pricing breaks are ignored or applied inconsistently. In many cases, the system always returns the same base shipping price regardless of the cart’s total weight, as if the breakpoints are not being evaluated correctly.

### Steps To Reproduce:
1. Create a Shipping Method and set **Charge By** to **Weight**
2. Assign the Shipping Method to a Zone
3. Create a Shipping Rate with multiple pricing breaks (Shipping Breakdowns) based on weight (e.g. 5, 10)
4. Add products with variant weights set 2kg
5.  Increase the cart total weight so that it crosses defined breakpoints  (add 4 quantity from the variant (4 * 2 = 8))
6. Observe that the returned shipping price does not change according to the configured tiers

Huncsuga:
**Description**

> - This PR fixes an issue where weight-based shipping rates ignore or inconsistently apply pricing breakpoints due to mismatched weight units during comparison.
> 
> - When a Shipping Method is configured with Charge By: Weight, the cart’s total weight and the pricing break min_quantity values are not normalized to the same unit. As a result, defined Shipping Breakdowns (tiers) are skipped and the base price is often returned regardless of the actual cart weight.
> 
> - This change ensures that both values are compared using a consistent weight unit, allowing pricing breakpoints to be evaluated correctly.

**Related Issue**

> Fixes #2381 

**Type**

> Bug fix

**What was changed**

> - Normalized weight unit handling when evaluating shipping rate pricing breaks
> 
> - Removed incorrect scaling (`×100`) applied when saving weight-based pricing break values
> 
> - Ensured pricing break thresholds are stored and evaluated using the same unit as cart weight calculations

**Expected Outcome**

> - Weight-based shipping tiers are applied correctly as cart weight increases
> 
> - Pricing breaks behave consistently regardless of the product’s configured weight unit (g / kg / lbs)

**How to test**

> 1. Create a Shipping Method with **Charge By: Weight**
> 2. Assign the Shipping Method to a Zone
> 3. Create a Shipping Rate with multiple weight-based pricing breaks (e.g. 5, 10)
> 4. Create two product variants:
>    - Variant A: weight = 2 kg
>    - Variant B: weight = 1500 g
> 5. Add the following items to the cart:
>    - 1 × Variant A (2 kg)
>    - 3 × Variant B (3 × 1500 g = 4.5 kg)
> 6. Verify that the shipping rate correctly evaluates the total cart weight (6.5 kg)
> 7. Confirm that the appropriate pricing tier is applied based on the combined weight


packages/table-rate-shipping/src/Drivers/ShippingMethods/ShipBy.php
```diff
@@ -71,7 +72,17 @@ public function resolve(ShippingOptionRequest $shippingOptionRequest): ?Shipping
...
             $tier = $cart->lines->sum(function ($line) {
-                return $line->purchasable->weight_value * $line->quantity;
+                try {
+                    $unitWeightKg = Converter::from("weight.{$line->purchasable->weight_unit}")
+                        ->to('weight.kg')
+                        ->value($line->purchasable->weight_value)
+                        ->convert()
+                        ->getValue();
+                } catch (\Exception $e) {
+                    return 0;
```
alecritson: This silently excludes the line's weight from the total, potentially qualifying the cart for a cheaper shipping tier. A fallback like `$line->purchasable->weight_unit ?: 'kg'` would be safer than silently zeroing the weight.

packages/table-rate-shipping/src/Filament/Resources/ShippingZoneResource/Pages/ManageShippingRates.php
```diff
@@ -244,7 +262,7 @@ function ($price) use ($chargeBy, $currencies) {
                 if ($chargeBy == 'cart_total') {
                     $price['min_quantity'] = (int) ($price['min_quantity'] * $currency->factor);
                 } else {
-                    $price['min_quantity'] = (int) ($price['min_quantity'] * 100);
+                    $price['min_quantity'] = (int) $price['min_quantity'];
```
alecritson, line 268: Any existing installation with weight-based shipping rates stored as multiplied integers (e.g., 500 meaning 5 kg) will have breakpoints that never match after upgrade, since 5.0 >= 500 is false.

A state class would be needed to rescale existing min_quantity values would be needed

alecritson (commented):
Thanks @Huncsuga Sorry for the lengthy delay in getting to look at this. Only a couple of comments if you're happy to pick this back up.

Also one concern is now we're not multiplying the value by 100 we need to ensure the pricing manager still works on the table-rate-shipping add-on. We'd need some tests to ensure everything this changes is covered.

Huncsuga:
> Thanks @Huncsuga Sorry for the lengthy delay in getting to look at this. Only a couple of comments if you're happy to pick this back up.
> 
> Also one concern is now we're not multiplying the value by 100 we need to ensure the pricing manager still works on the table-rate-shipping add-on. We'd need some tests to ensure everything this changes is covered.

No problem, I’ve added tests and fixed the PR comments.

packages/table-rate-shipping/src/Filament/Resources/ShippingZoneResource/Pages/ManageShippingRates.php
Huncsuga: Fixed

packages/table-rate-shipping/src/Drivers/ShippingMethods/ShipBy.php
Huncsuga: @alecritson Fixed. I also added a fallback so if the unit is not weight-based, it still calculates in kg. Ideally it should always be weight, but this extra safeguard shouldn’t cause issues.

alecritson (approved):

PR #2384, opened 2025-12-14
Fix/default weight unit

### LINKED ISSUE #2383: Product variant weight_unit defaults to mm instead of a valid weight unit
- Lunar version: 1.2.1
- Laravel Version: 12.42.0
- PHP Version: 8.3.28
- Database Driver & Version: MySQL 8.0.37

### Expected Behaviour:
When creating a product and its related variants, the `weight_unit` field should default to a valid **weight measurement unit**, such as `kg`, `g`, or `lbs`.

### Actual Behaviour:
When a product variant is created, the `weight_unit` field is automatically set to `mm` by default in the database.  
`mm` is a **length unit**, not a weight unit, which leads to incorrect or inconsistent behavior when weight-based logic (e.g. shipping rates, weight calculations) is applied.

### Steps To Reproduce:
1. Create a new product in the Lunar admin panel..
2. Check the database record for the created product variant.
3. Observe that `weight_unit` is set to `mm` by default.

Huncsuga:
**Description**

> - This PR fixes an issue where newly created product variants default their weight_unit to mm, which is a length unit and not a valid weight measurement.
> 
> - When a product is created the weight unit, the system automatically assigns mm as the default value in the database. This leads to incorrect behavior in features that rely on weight-based logic, such as shipping calculations.
> 
> - This change ensures that product variants use a valid weight unit by default.

**Related Issue**

> - Fixes #2383

**Type**

> - Bug fix

**What was changed**

> - Corrected the default value of weight_unit for product variants
> 
> - Ensured the default unit represents a valid weight measurement (kg)

**Expected Outcome**

> - Newly created product variants no longer default to an invalid weight unit
> 
> - Weight-based calculations (e.g. shipping) behave consistently
> 
> - Data stored in the database correctly reflects weight semantics

**How to test**

> - Create a new product in the Lunar admin panel
> 
> - Inspect the weight_unit value in the database
> 
> - Verify that the unit defaults to a valid weight unit (e.g. kg) and not mm


packages/core/database/migrations/2025_12_14_215636_update_weight_unit_default_on_product_variants_table.php
```diff
@@ -0,0 +1,28 @@
...
+
+return new class extends Migration
+{
+    public function up(): void
+    {
+        Schema::table($this->prefix . 'product_variants', function (Blueprint $table) {
+            $table->string('weight_unit')->default('kg')->nullable()->change();
+        });
+
+        DB::table($this->prefix . 'product_variants')
```
alecritson: We shouldn't be doing any data changes within migrations, instead use a state update class. See the `database/state` folder for how these are structured :)

Huncsuga: @alecritson Thanks for the PR comment! I’ve updated it.

alecritson (approved):

PR #2393, opened 2025-12-30
Fixed error when adding shipping/billing address to order without existing address

### LINKED ISSUE #2392: Error when adding shipping/billing address to order without existing address
- Lunar version: 1.2.1
- Laravel Version: 11.47.0
- PHP Version: 8.4.15
- Database Driver & Version: MySQL 8.0

### Expected Behaviour:
When an order doesn't have an existing shipping or billing address, clicking the "Edit" button should open a modal with an empty form, allowing the staff user to add a new address to the order. Upon saving, a new address record should be created and associated with the order. The timeline should include the new address details in the activity log, just as it does when updating an existing address.

### Actual Behaviour:
When attempting to edit a shipping or billing address on an order that doesn't have one yet, the system throws an error trying to call methods on a null relationship, preventing the address from being added.

### Steps To Reproduce:
1. Create an order without setting a shipping or billing address
2. Navigate to the order in the admin panel as a staff user
3. Click the "Edit" button on the shipping or billing address section
4. Fatal 500 error: `Call to a member function toArray() on null vendor/lunarphp/lunar/src/Filament/Resources/OrderResource/Concerns/DisplaysOrderAddresses.php:210`

This appears to be occurring due to the [`fillForm`](https://github.com/lunarphp/lunar/blob/1.2.1/packages/admin/src/Filament/Resources/OrderResource/Concerns/DisplaysOrderAddresses.php#L209) not handling null addresses for both shipping and billing. Additionally, the [`action`](https://github.com/lunarphp/lunar/blob/1.2.1/packages/admin/src/Filament/Resources/OrderResource/Concerns/DisplaysOrderAddresses.php#L217) does not handle creating new addresses and assumes that its always editing an existing one

NB: In my use case, I have a custom cart validation for some scenarios where a billing and/or shipping addresses aren't required on an order. The default Lunar cart validation does prevent you from adding an order without a valid billing address, and does require a shipping address when the product is shippable.

jamiesage123:
This pull request fixes #2392.

When attempting to edit a shipping or billing address on an order that doesn't have one yet, the system throws an error trying to call methods on a null relationship, preventing the address from being added.

This pull request changes the `DisplaysOrderAddresses.php` to handle creating a new address when one doesn't exist, on top of the existing logic allowing editing, while keeping the activity log for both scenarios.

### DIFF
--- a/packages/admin/src/Filament/Resources/OrderResource/Concerns/DisplaysOrderAddresses.php
+++ b/packages/admin/src/Filament/Resources/OrderResource/Concerns/DisplaysOrderAddresses.php
@@ -207,8 +207,8 @@ trait DisplaysOrderAddresses
             ->label(__('lunarpanel::order.action.edit_address.label'))
             ->button()
             ->fillForm(fn ($record) => match ($type) {
-                'shipping' => $record->shippingAddress->toArray(),
-                'billing' => $record->billingAddress->toArray(),
+                'shipping' => $record->shippingAddress?->toArray(),
+                'billing' => $record->billingAddress?->toArray(),
                 default => []
             })
             ->form(function () {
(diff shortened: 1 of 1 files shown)


alecritson:
Thanks @jamiesage123 but it seems this is a duplicate of #2375 , if you think changes you have made here should be included on that PR feel free to comment.

PR #2399, opened 2026-01-29
Fix cart retrieval to prefer unmerged carts (and clarify active scope grouping)

### LINKED ISSUE #2397: Potential bug in scopeActive and CartSessionManager login logic
I've encountered an issue where a merged cart can be returned during login, and scopeActive might leak carts from other users due to missing query grouping.

- Lunar version: 1.2.1
- Laravel Version: 12.48.1
- PHP Version: 8.4.15
- Database MySQL 8.4.2

### Expected Behaviour:
When a user logs in (auth_policy=merge), the guest cart should be merged into the user's active cart and the session should resolve the unmerged "target" cart on subsequent requests/logins. Merged source carts (merged_id != null) should not be returned as the current cart.

### Actual Behaviour:
After merging carts (a source cart gets merged_id set), subsequent login (without creating a new guest cart) can resolve the merged source cart as the "current" cart, resulting in missing lines/items in the frontend/session.

This seems to happen because:
1) Some call sites resolve the user's cart using `->active()->first()` without excluding merged carts (`->unmerged()`), so a merged source cart can be returned.
2) `Cart::scopeActive()` uses `whereDoesntHave()->orWhereHas()` without grouping, which can cause the OR clause to escape other constraints (e.g. relationship constraints like user_id).

### Steps To Reproduce:
1. Ensure config:
   - `lunar.cart.auth_policy = 'merge'`
   - `lunar.cart_session.auto_create = false`
2. Log in as a user and create/ensure a user cart exists (add an item).
3. Log out (ensure cart is not deleted on logout; e.g. CartSession::forget(false)).
4. As guest, add an item (creates a guest cart in session).
5. Log in (merge happens; source cart gets merged_id pointing to target cart).
6. Log out and log in again WITHOUT adding new guest items.
7. Observe that the resolved "current" cart can be the merged source cart (merged_id != null), rather than the target unmerged cart.

### Notes / Proposed Fix:

- Ensure session/user cart resolution excludes merged carts, in CartSessionManager
```php
$cartId = $user->carts()
    ->unmerged()
    ->active()
    ->latest('id')
    ->value('id');
```

- Group OR conditions in `Cart::scopeActive()`:
```php
public function scopeActive(Builder $query): Builder
{
    return $query->where(function ($q) {
        $q->whereDoesntHave('orders')
          ->orWhereHas('orders', function ($sub) {
              $sub->whereNull('placed_at');
          });
    });
}
```

vencious:
### Summary
Fixes an issue where an authenticated user could have a previously-merged cart selected during cart resolution, leading to unexpected “missing” items after login/logout cycles. Also makes the `Cart::active()` scope grouping explicit for readability/intent.

### Why is this needed?
When a user has carts that were merged into a newer cart, selecting `->active()->first()` may return a merged/source cart depending on ordering and query shape. We now prefer the latest unmerged active cart.

### Changes
- In `CartSessionManager`, prefer the latest **unmerged** active cart when resolving an authenticated user's cart.
- Make the `Cart::active()` scope grouping explicit (readability/intent).
- Add regression tests covering merged cart selection and user-context active cart querying.

Fixes #2397


alecritson:
Thanks @vencious Can you quickly update the tests so the comments are in English otherwise we're unsure of what's happening :)

vencious:
Sorry about that — I’ve updated and pushed: removed the non-English comments and clarified the test intent via naming/assertions.

ryanmitchell:
This is great, thanks for figuring out the issue. We've been trying to figure out the reason for this happening on our Lunar sites.

vencious:
> This is great, thanks for figuring out the issue. We've been trying to figure out the reason for this happening on our Lunar sites.

Glad to hear that!

alecritson (approved):

PR #2406, opened 2026-02-04
Supported TaxZone in priceIncTax methods.

MacTavish-69:
Added optional TaxZone parameter to ``priceIncTax``, ``comparePriceIncTax``, and ``getPriceableTaxRate`` methods to allow tax calculations based on specific tax zones.

# Details
Lunar internally worked on default tax and default tax zone basis only. With this PR, we can provide our own determined tax zone (typically IP -> Country -> Tax Zone) and provide it to the method to show us tax inclusive prices.

Unless custom Tax zone is provided, the method continues to behave as normal and previously.

``Documentation Update Link will be added after generation of this PR.``


packages/core/src/Models/Price.php
```diff
@@ -92,41 +92,49 @@ public function priceExTax(): \Lunar\DataTypes\Price
...
 
     /**
      * Return the total tax rate amount within the predefined tax zone for the related priceable
      */
-    protected function getPriceableTaxRate(): int|float
+    protected function getPriceableTaxRate(?TaxZone $taxZone = null): int|float
     {
-        return Blink::once('price_tax_rate_'.$this->priceable->getTaxClass()->id, function () {
-            $taxZone = TaxZone::where('default', '=', 1)->first();
+        /* This function is called multiple times somewhere from the package and since we just added an optional
```
glennjacobs, line 127: @alecritson not sure we want to leave this comment in the code

MacTavish-69: Ignore this PR in favor of #2411 , because this functionality is present in the newer PR as well.

This PR/Code was our earlier solution to show Tax Incl. Prices but it wouldn't work on cart items. It would only be applicable on Models which had price attached to them.

PR #2408, opened 2026-02-16
Add and enforce shipping:manage permission

ryanmitchell:
At the moment there is no enforcement of a permission for the table rate shipping resources in the panel, this PR adds that.

### DIFF
--- a/packages/table-rate-shipping/database/migrations/2026_02_16_100000_add_can_manage_permission.php
+++ b/packages/table-rate-shipping/database/migrations/2026_02_16_100000_add_can_manage_permission.php
@@ -1,0 +1,37 @@
+<?php
+
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+use Lunar\Admin\Support\Facades\LunarPanel;
+use Lunar\Base\Migration;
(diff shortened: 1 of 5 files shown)


packages/table-rate-shipping/database/migrations/2026_02_16_100000_add_can_manage_permission.php
```diff
@@ -0,0 +1,40 @@
+<?php
+
+use Illuminate\Support\Facades\Schema;
+use Lunar\Admin\Support\Facades\LunarPanel;
+use Lunar\Base\Migration;
+use Spatie\Permission\Models\Permission;
+
+class AddCanManagePermission extends Migration
```
alecritson: Can we use anonymous classes so it's consistent across the package :)

```php
return new class extends Migration
{
```

alecritson:
Thanks @ryanmitchell only a couple of small bits, the other being the name of the migration.

```
2026_02_16_100000_add_can_manage_permission.php
```

Can this be something like `_add_can_manage_shipping_permission` so it's clear what the intent is.

ryanmitchell:
@alecritson both changes made

alecritson (approved):

PR #2411, opened 2026-02-27
Updated Price Model TaxIncl methods & Cart Item TaxIncl properties to actually provide Tax inclusive values.

MacTavish-69:
# Lunar Current Behavior
1. Taxes are only accounted for when shipping address is set.
    > For most stores, this is at the time of checkout. Logically, if tax is applicable on end user, the prices should be shown tax inclusive. Often, this is a legal requirement.
2. Lunar internally relies on Default Tax Class & Default Tax Zone.
   > When Lunar Internally just uses Default Tax class & Zone, this system causes incorrect behavior when you've some taxed zone and some zones which are NOT taxed. In this case, you'll set the non taxed zone as Default and the incTax methods will result in same non-taxed amount.
3. Because of 1 & 2, the taxIncl methods often don't actually provide the intended taxIncl prices.

# Fixes in PR
The PR approaches this problem at cart level and Price Model level.

## Cart Level
We add TaxZone property to Cart model. The typical flow is to find the user country based on IP in middleware, determine the tax zone and set it on the cart. With this, The items added to cart, will correctly give tax incl price when taxInc methods are called.

## Price Model Level
When we have Product/Purchasable instead of Cart Line (Item is added to cart), we get the price from prices relation (prices attached to purchasable) and utilize ``priceIncTax``, ``comparePriceIncTax`` methods. An optional TaxZone parameter is added so you can get the tax inclusive price as per the zone of your choice. This is useful when you need it for one time for any reason.
**Note:** You can totally ignore it if you need the prices to be as per the zone set on cart, the PR will fallback to the zone set on cart.

# Previous PR: #2406
The previous PR focused strictly on Product/Purchasable display prices. It introduced a taxzone parameter to specific methods, allowing developers to:
- Identify the user's taxable region manually.
- Pass that taxzone into price methods to display the correct tax-inclusive amounts.
- While effective for simple product displays, it did not address the underlying Cart logic.

This PR is a superset of #2406. It retains the functionality mentioned above but expands it by adding a taxZone property directly to the Cart.

**IMPORTANT:**
Since this PR includes all changes from [#2406](https://github.com/lunarphp/lunar/pull/2406), please reject/close that PR if you choose to merge this one.


packages/core/src/Models/Contracts/Price.php
```diff
@@ -4,6 +4,7 @@
 
 use Illuminate\Database\Eloquent\Relations\BelongsTo;
 use Illuminate\Database\Eloquent\Relations\MorphTo;
+use Lunar\Models\TaxZone;
```
alecritson: We should be using the contract here.

packages/core/src/Models/Cart.php
```diff
@@ -617,4 +644,29 @@ public function getEstimatedShipping(array $params, bool $setOverride = false):
...
+     * Set the tax zone override for this cart.
+     *
+     * When set, all tax calculations will use this zone instead of resolving one from the shipping address.
+     * Pass null to clear the override and fall back to the address-derived (or default) zone.
+     *
+     * The zone ID is mirrored into the cart's `meta` JSON column so the choice survives across requests.
+     * Call `->save()` afterwards to persist it to the database; the `booted()` retrieved-event listener
+     * will then restore the zone automatically on every subsequent page load.
+     */
+    public function setTaxZone(?TaxZone $taxZone): Cart
```
alecritson: Although it would cause a breaking change, this should be added to the `Cart` contract.

packages/core/src/Models/Cart.php
```diff
@@ -617,4 +644,29 @@ public function getEstimatedShipping(array $params, bool $setOverride = false):
...
+    {
+        $this->taxZone = $taxZone;
+        $meta = $this->meta ?? new \ArrayObject;
```
alecritson: The meta column is cast with `AsArrayObject::class`, so `$this->meta` is already an `ArrayObject` instance when the model exists. Creating new `ArrayObject` as a fallback is only needed for a brand-new unsaved cart where meta was never set. 

The issue is that `AsArrayObject` returns the same object instance on repeated access (it is not cloned), so assigning `$this->meta = $meta` re-assigns the same reference and may not trigger Eloquent's dirty tracking, meaning a subsequent `->save()` call might not persist the meta change. The pattern should use `$this->forceFill(['meta' => ...])` to ensure dirty state is marked.

packages/core/src/Models/Cart.php
```diff
@@ -41,6 +41,8 @@
 use Lunar\Exceptions\FingerprintMismatchException;
 use Lunar\Facades\DB;
 use Lunar\Facades\ShippingManifest;
+use Lunar\Models\TaxZone;
```
alecritson: Ensure we reference the contracts and not concrete classes.

packages/core/src/Pipelines/Cart/CalculateLines.php
```diff
@@ -18,6 +19,14 @@ class CalculateLines
     public function handle(CartContract $cart, Closure $next): mixed
     {
         /** @var Cart $cart */
+
+        // Cache the TaxZone so that price model taxInc methods accounts for it in computation.
+        if ($cart->taxZone) {
+            Blink::put('lunar_cart_tax_zone', $cart->taxZone);
```
alecritson: If application code calculates two carts in the same request (e.g. comparing shipping options across accounts, or a queue job processing multiple carts in the same process), the second cart's `CalculateLines` pass will overwrite the Blink key set by the first, meaning `Price::getPriceableTaxRate()` will silently use the wrong zone for calls made after the key is overwritten.

Recommendation: Key the Blink entry by cart ID: 'lunar_cart_tax_zone_'.$cart->id, and have `Price::getPriceableTaxRate()` accept the cart ID as a lookup hint, or pass the zone directly rather than routing it through Blink at all.

MacTavish-69:
@alecritson I've seen your review. I'm quite caught up in these days. I'll try to get this on priority. I'll try to make amendments as per the review over the weekend

packages/core/src/Models/Contracts/Price.php
MacTavish-69: Done. See f7a46fc

packages/core/src/Models/Cart.php
MacTavish-69: Done. See 742e0942cd0c4968a2b6283093ac64e0b41ff5cb

packages/core/src/Models/Cart.php
MacTavish-69: Done. See dbf31de9622b2b2ed4f9b70c117067790644717d

packages/core/src/Models/Cart.php
MacTavish-69: Done. See 36d7416272f1ae7ff92c9d6095515561751229f0

packages/core/src/Pipelines/Cart/CalculateLines.php
MacTavish-69: Done. See 068df61118656a461a9341f2eab2e9db96b1dde4

MacTavish-69:
> @alecritson I've seen your review. I'm quite caught up in these days. I'll try to get this on priority. I'll try to make amendments as per the review over the weekend

Hello @alecritson , The review has been accommodated for. Sorry for longer delay as I got busy and then carried away.

I hope it's ready for merging now.

glennjacobs:
Using cart meta is wrong. I'll do an update to implement this properly.

alecritson (approved):

PR #2423, opened 2026-03-24
Allow shipping tax rate to be determined by a callable

ryanmitchell:
Adds support for a callable config value for `shipping_rate_tax_calculation`, allowing applications to provide custom logic to determine which tax class should be applied to shipping rates.

  Previously the config only accepted `'default'` (system default tax class) or `'highest'` (highest tax rate among cart lines). Any value other than `'default'` would fall through to the highest-rate resolver, with no way to supply arbitrary logic.

  ## Changes

  - **`ShippingRate::getShippingOption()`** — when `shipping_rate_tax_calculation` is set to a callable (closure or invokable class), it is called with the `Cart` instance and the return value is used as the resolved `TaxClass`

  ## Usage

  ```php
  // config/shipping-tables.php
  'shipping_rate_tax_calculation' => function (\Lunar\Models\Cart $cart): \Lunar\Models\TaxClass {
      // custom logic — e.g. return a fixed tax class for B2B customers
      return $cart->user?->isBusinessCustomer()
          ? TaxClass::where('name', 'B2B')->first()
          : TaxClass::getDefault();
  },
```
  An invokable class is also supported:

  ```
  'shipping_rate_tax_calculation' => [\App\Shipping\CustomTaxClassResolver, 'handle'],
  ```

### DIFF
--- a/packages/table-rate-shipping/config/shipping-tables.php
+++ b/packages/table-rate-shipping/config/shipping-tables.php
@@ -6,7 +6,8 @@ return [
     /*
      * What method should we use for a shipping rate tax calculation?
      * Options are 'default' for the system-wide default tax rate,
-     * or 'highest' to select the highest tax rate in the cart
+     * 'highest' to select the highest tax rate in the cart
+     * or add a callable to use your own logic (eg => [MyTaxRateCalculator::class, 'calculate'])
      */
     'shipping_rate_tax_calculation' => 'default',
 ];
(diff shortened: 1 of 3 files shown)


alecritson:
Thanks @ryanmitchell I think we should avoid closures in config file as this will stop serialization, also maybe instead (or on top) of having to provide a handle `[MyTaxRateCalculator::class, 'calculate']` we make use of invokable classes?

Finally I think if we can simply or tidy up the nested if statements that would be ideal.

ryanmitchell:
@alecritson nested ifs are removed.

It already supports invokeable classes (eg 'shipping_rate_tax_calculation' => MyInvokableClass::class).

alecritson (approved):

PR #2433, opened 2026-04-07
Table rate shipping register ShippingDiscount discount type

ryanmitchell:
This PR adds a ShippingDiscount discount type which is registered by the Table Rate Shipping package.

When selected, this discount type allow you to select a shipping method and a price to offer it at.

Some examples of what this allows:                                                                                                                                                                                                                        
  - Free shipping on any method: leave Shipping method blank, set price = 0
  - Free shipping on Standard only: select Standard Shipping, price = 0                                                                                                                                                                     
  - Express for £5 when normally £15: select Express Shipping, price = 5.00
  - If the cart's selected shipping method doesn't match the configured one, the discount silently skips


alecritson:
Thanks @ryanmitchell Overall I think it's a sound idea, the only points of confusion I had when testing was it's not immediately clear what values I should be putting in, for example I just added £2.50 expecting it to take that value off the price of shipping but it set it to £2.50...

Also should this discount also allow the user to set whether the discount is just free shipping, or would you simply input 0 into the price field?

ryanmitchell:
Yes it's the price you sell the shipping at, not the discount price, I think you get more control that way. And yes to get free shipping you just enter 0.

alecritson:
> Yes it's the price you sell the shipping at, not the discount price, I think you get more control that way. And yes to get free shipping you just enter 0.

Gotcha, I think for me then the confusing part is the heading being "Shipping discount" which suggests you're adding a discount, not a set price so maybe some other instructions to explain the intent. Also can we make the price columns in a grid of 3

image

ryanmitchell:
I've changed the label to "Shipping Price", which is maybe clearer? If not let me know and I'll rethink.

Currencies is now a group of 3 columns.

alecritson:
I think it's better although I might be throwing a bit of a spanner in the scope of this PR.

- What if I want to apply the discount to multiple shipping methods with different prices?
- Should we allow a percentage based value? Some shipping methods also might have different rates depending on shipping zones.

ryanmitchell:
Sure, easily changed: 

CleanShot 2026-04-09 at 08 55 47@2x

alecritson:
Seems to work okay, the only thing i noticed (and not related to your work) was in `Base/Validation/CouponValidator.php` we have a `whereIn` which will stop this discount from validating, but I can't think why it would be needed..so maybe it can be removed (or add your discount type to it)

ryanmitchell:
> Seems to work okay, the only thing i noticed (and not related to your work) was in `Base/Validation/CouponValidator.php` we have a `whereIn` which will stop this discount from validating, but I can't think why it would be needed..so maybe it can be removed (or add your discount type to it)

Great, I've removed that (we have to work around it on our sites so delighted to take it away!)

glennjacobs (approved):

PR #2434, opened 2026-04-07
Update cut off to be an availability schedule

ryanmitchell:
This PR updates the cut off in Shipping Methods to be an availability schedule:

CleanShot 2026-04-07 at 09 09 54@2x

This provides much finer control over when the method is available to customers.

### DIFF
--- a/packages/table-rate-shipping/src/Models/ShippingMethod.php
+++ b/packages/table-rate-shipping/src/Models/ShippingMethod.php
@@ -3,6 +3,7 @@
 namespace Lunar\Shipping\Models;
 
 use Carbon\Carbon;
+use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Database\Eloquent\Casts\AsArrayObject;
 use Illuminate\Database\Eloquent\Factories\HasFactory;
 use Illuminate\Database\Eloquent\Relations\BelongsToMany;
@@ -64,7 +65,7 @@ class ShippingMethod extends BaseModel implements Contracts\ShippingMethod
      * Each day entry has: enabled (bool), from (time string "H:i"), to (time string "H:i").
      * When no schedule is configured the method is always available.
      */
-    public function scopeAvailableAt(\Illuminate\Database\Eloquent\Builder $query, Carbon $now): \Illuminate\Database\Eloquent\Builder
+    public function scopeAvailableAt(Builder $query, Carbon $now): Builder
     {
         return $query->where(function ($query) use ($now) {
             $dayKey = (string) $now->isoWeekday();
--- a/packages/table-rate-shipping/src/Models/ShippingMethod.php
+++ b/packages/table-rate-shipping/src/Models/ShippingMethod.php
@@ -61,45 +61,36 @@ class ShippingMethod extends BaseModel implements Contracts\ShippingMethod
      * Determine whether this shipping method is available at the given moment.
      *
      * The schedule is stored in data.schedule keyed by ISO weekday (1=Monday … 7=Sunday).
-     * Each day entry has: enabled (bool), from (time string), to (time string).
-     * Falls back to the legacy cutoff column when no schedule is configured.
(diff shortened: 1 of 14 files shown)


alecritson:
Thanks @ryanmitchell I think if the plan is to not use `cutoff` anymore then I feel we should have a migration to remove the column and use a state updater to migrate cutoff times to the schedule, I guess if a cutoff time is set then a schedule should be added for each day with the cutoff time set on `until`.

Also how would you then query shipping methods by their schedule? For example it's currently quite trivial to get all shipping methods which are due to cut off within a certain timeframe. 

There also doesn't appear to be any sort of validation to stop a user from setting from as `18:00` and to as `17:00` which wouldn't make sense.

ryanmitchell:
> Thanks @ryanmitchell I think if the plan is to not use cutoff anymore then I feel we should have a migration to remove the column and use a state updater to migrate cutoff times to the schedule, I guess if a cutoff time is set then a schedule should be added for each day with the cutoff time set on until.

No problem - I can work on that.


> Also how would you then query shipping methods by their schedule? For example it's currently quite trivial to get all shipping methods which are due to cut off within a certain timeframe.

```
$query
	->whereTime('schedule.{now()->dayOfWeek}.from', '>=', now())
	->whereTime('schedule.{now()->dayOfWeek}.until', '<=', now());
```

could add it as a scope if you want? I have added isAvailableAt() on the model itself.

> There also doesn't appear to be any sort of validation to stop a user from setting from as 18:00 and to as 17:00 which wouldn't make sense.

Good spot, i'll work on that too.

packages/table-rate-shipping/src/Models/ShippingMethod.php
```diff
@@ -56,6 +57,42 @@ public function shippingRates(): HasMany
...
     }
 
+    /**
+     * Determine whether this shipping method is available at the given moment.
+     *
+     * The schedule is stored in data.schedule keyed by ISO weekday (1=Monday … 7=Sunday).
+     * Each day entry has: enabled (bool), from (time string "H:i"), to (time string "H:i").
+     * When no schedule is configured the method is always available.
+     */
+    public function isAvailableAt(?Carbon $now = null): bool
```
alecritson: Not too sure on the naming, calling the method without any arguments i.e. `isAvailableAt()` doesn't make sense, I think maybe just `isAvailable` works well without arguments and also works with parameters `isAvailable(now()->addDay())` for example.

alecritson:
> ```
> $query
> 	->whereTime('schedule.{now()->dayOfWeek}.from', '>=', now())
> 	->whereTime('schedule.{now()->dayOfWeek}.until', '<=', now());
> ```
> 
> could add it as a scope if you want? I have added isAvailableAt() on the model itself.

Yeah I think you'd have to be strict in passing the Carbon instance through the scope though, we can't have `now()` calls here I don't think. If the scope is called within a loop or a delayed job or similar then `now` will end up being relative to the time of the method call.

packages/table-rate-shipping/src/Models/ShippingMethod.php
ryanmitchell: changed to isAvailable

ryanmitchell:
> Yeah I think you'd have to be strict in passing the Carbon instance through the scope though, we can't have `now()` calls here I don't think. If the scope is called within a loop or a delayed job or similar then `now` will end up being relative to the time of the method call.

query scope added

alecritson:
I had a little look at this in the panel, something feels a bit off with it's position and look. Do you think this would be better suited within the availability page? that would also mean it doesn't have to be hidden on load as I think this might cause a UX issue since it's not clear if you had a schedule set on initial load.

ryanmitchell:
I can drop it in on the availability page like this?

CleanShot 2026-04-08 at 14 16 14@2x

glennjacobs (approved):

PR #2435, opened 2026-04-07
Allow a min and max weight for shipping methods

ryanmitchell:
This PR updates shipping methods to have a min/max weight option:

CleanShot 2026-04-07 at 10 26 43@2x


packages/table-rate-shipping/src/Resolvers/ShippingRateResolver.php
```diff
@@ -65,6 +69,27 @@ public function cart(CartContract $cart): self
...
+        // method's weight_unit later without iterating cart lines again.
+        $this->cartWeightKg = (float) $this->cart->lines->sum(function ($line) {
+            $value = (float) ($line->purchasable->weight_value ?? 0);
+            $unit = $line->purchasable->weight_unit ?? 'kg';
+
+            if ($value === 0.0) {
+                return 0.0;
+            }
+
+            if ($unit !== 'kg') {
```
alecritson: I wonder if the converter would be okay with converting kg to kg as in my head it would just be 1:1. This way we could get rid of the if statements potentially and just have something like:

```php
return (float) Converter::from("weight.{$unit}")
                    ->to('weight.kg')
                    ->value($value * $this->quantity)
                    ->convert()
                    ->getValue();
```

We probably wouldn't even need to return early either if there is no value since `0 * N` is always going to be 0.

ryanmitchell: Done!

packages/table-rate-shipping/src/Filament/Resources/ShippingMethodResource.php
```diff
@@ -119,6 +122,30 @@ public static function getCutoffFormComponent(): Component
...
+            Forms\Components\Select::make('weight_unit')
+                ->label(__('lunarpanel.shipping::shippingmethod.form.weight_unit.label'))
+                ->options($weightUnits)
+                ->placeholder(__('lunarpanel.shipping::shippingmethod.form.weight_unit.placeholder')),
+            Forms\Components\TextInput::make('min_weight')
+                ->label(__('lunarpanel.shipping::shippingmethod.form.min_weight.label'))
+                ->numeric()
+                ->minValue(0)
+                ->required(fn (Get $get) => filled($get('weight_unit'))),
+            Forms\Components\TextInput::make('max_weight')
```
alecritson, line 178: Looks like we need some validation to ensure you can't set the min weight higher than the max weight.

ryanmitchell: Yep done

glennjacobs (approved):

PR #2436, opened 2026-04-07
Update to L13

alecritson:
Adds L13 support and drops EOL L11


packages/core/src/Base/BaseModel.php
```diff
@@ -22,4 +22,22 @@ public function __construct(array $attributes = [])
             $this->setConnection($connection);
         }
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function resolveCollectionFromAttribute()
```
alecritson, line 30: Laravel 13's `resolveCollectionFromAttribute()` walks up the parent chain and tries new `(get_parent_class(static::class))` This fails because BaseModel is abstract. The override stops the traversal at `BaseModel` by returning null instead of trying to instantiate a parent class.

packages/core/src/Console/Commands/MigrateGetCandy.php
```diff
@@ -30,7 +30,7 @@ class MigrateGetCandy extends Command
     public function handle()
     {
         $tableNames = collect(
-            DB::connection()->getDoctrineSchemaManager()->listTableNames()
+            Schema::getTableListing()
```
alecritson, line 33: We might even want to strip out all the GetCandy migration code at this point. I can't imagine anyone will be using it.

composer.json
```diff
@@ -12,8 +12,8 @@
         "awcodes/filament-badgeable-column": "^v3.0.0-beta.1",
         "awcodes/shout": "^v3.0.0-beta.2",
         "barryvdh/laravel-dompdf": "^3.0",
-        "cartalyst/converter": "^9.0|^10",
-        "doctrine/dbal": "^3.6|^4.0",
+        "cartalyst/converter": "dev-l13-compatibility as 10.0.0",
```
alecritson: https://github.com/cartalyst/converter/pull/35

alecritson: We've moved away from this package so no longer relevant.

packages/core/src/Console/Commands/MigrateGetCandy.php
glennjacobs: I think v2 will achieve that, so maybe leave as-is on v1?

glennjacobs:
I wonder if we can get away from these beta releases now?

```json
 "awcodes/filament-badgeable-column": "^v3.0.0-beta.1",
 "awcodes/shout": "^v3.0.0-beta.2",
```

glennjacobs:
Should we bump the Pint version to `1.29.0`?

glennjacobs (dismissed):

glennjacobs (approved):

PR #2443, opened 2026-04-11
Add returns/RMA foundation hooks to Lunar v1.x

tazzytazzy:
Preliminary release of adding to the hook surface needed to unblock reusable returns/RMA integrations in Lunar v1.x.

Let me know what you think. Probably few tweaks here and there might be needed to help align with your goals. Let me know and I'll work on it.

- add OrderPaid plus canonical refund orchestration via RefundTransaction
- add pre-refund authorization DTOs/contracts and refund lifecycle events
- expand PaymentRefund with refund transaction context and allocation passthrough
- add OrderLineCapabilitiesInterface with the default conservative resolver
- route admin refund actions through the orchestration path
- document the new refund and order hook surface in the Lunar docs

Test coverage:
- refund allowed, denied, driver-failed, and thrown-exception paths
- RefundRequested ordering before the driver call
- RefundCompleted and RefundFailed dispatch behavior
- allocation passthrough and refund transaction hydration
- default and override order-line capability behavior, including unplaced lines
- admin header and bulk refund flows
- OrderPaid coverage for Offline, Stripe, Opayo, and Paypal
- Paypal expanded PaymentRefund coverage

Verified with:
- vendor/bin/pint --dirty --format agent
- vendor/bin/pest tests/core/Unit/Actions/Orders/RefundTransactionTest.php tests/core/Unit/Base/OrderLineCapabilitiesTest.php tests/core/Unit/PaymentTypes/OfflinePaymentTypeTest.php tests/stripe/Unit/StripePaymentTypeTest.php tests/opayo/Feature/OpayoPaymentTypeTest.php tests/admin/Feature/Filament/Resources/OrderResource/Pages/ManageOrderTest.php tests/paypal/Unit/PaypalPaymentTypeTest.php

Result:
- 34 tests passed
- 207 assertions


tazzytazzy:
I'm working on resolving these failures.

glennjacobs:
I've been pondering on this PR for a while. The main concern is that we want to introduce inventory, fulfilments and returns (etc) to v2, and I can't be sure right now that the capabilities interface is really what we want.

It might be worth a discussion on Discord.

glennjacobs:
Hi @tazzytazzy, first off thank you for this. It's a genuinely thorough piece of work, the test coverage across all the payment drivers especially, and I can see the care that's gone into the refund orchestration and the capabilities resolver.

I'm going to close this, but I want to be clear it's a direction decision rather than a reflection on the work itself.
We're not introducing order fulfilments, inventory, or the returns/RMA surface into v1. These are concepts we're deliberately holding for v2, where orders move to a state-machine model for payment and fulfilment status, and where the underlying primitives (fulfilments as first-class entities, a stock movement ledger, return-capable movements) will look quite different from what we'd be able to bolt onto 1.x. I don't want to ship a hook surface into v1 that we'd then have to reshape or deprecate a few months later, as that's a worse experience for everyone building against it.

On RMA specifically, the plan is for the workflow to live in add-on packages rather than core, with core only owning the minimal primitives those packages target. So even in v2 a lot of what's here would sit in a package rather than core itself.

Thanks again for the effort and for being so responsive on the cleanup. Sorry it's landing as a close.

tazzytazzy:
Just to be clear, this doesn't add RMA, fulfillments, etc, this just adds the hooks to add these features. If you have a hook for this for V2, can you backport that hook so that I can use it?

PR #2447, opened 2026-04-13
Fix dropdown lookup state normalization for attribute edit forms

tazzytazzy:
## Summary

This fixes dropdown attribute lookup hydration when editing attributes in the admin panel. This should resolve PR #2443 too.

Dropdown `lookups` are stored as an array of `{ label, value }` pairs, but the edit form expects keyed state for the `KeyValue` component. That mismatch caused lookup state to hydrate incorrectly in attribute edit flows.

## What changed

- add configuration normalization support for field types
- normalize dropdown lookup configuration for edit-form state
- normalize attribute relation-manager configuration state during hydration
- add focused regression coverage for:
  - dropdown lookup normalization
  - editing dropdown attribute lookups in the relation manager

## Why

The admin attribute edit flow needs dropdown lookup state in keyed form like:

```php
[
    'lookups' => [
        'Foo' => 'bar',
    ],
]
```
but persisted configuration is stored like:

```php
[
    'lookups' => [
        ['label' => 'Foo', 'value' => 'bar'],
    ],
]
```
Without normalization, the edit form mounts with the wrong shape and the lookup field does not hydrate correctly.

## Tests
- tests/admin/Feature/Filament/Resources/AttributeGroupResource/RelationManagers/AttributesRelationManagerTest.php
- tests/admin/Unit/Support/Forms/AttributeConverters/DropdownConverterTest.php

### DIFF
--- a/packages/admin/src/Filament/Resources/AttributeGroupResource/RelationManagers/AttributesRelationManager.php
+++ b/packages/admin/src/Filament/Resources/AttributeGroupResource/RelationManagers/AttributesRelationManager.php
@@ -127,12 +127,6 @@ class AttributesRelationManager extends BaseRelationManager
                     ->schema(function (Get $get) {
                         return AttributeData::getConfigurationFields($get('type'));
                     })
-                    ->afterStateHydrated(function (Grid $component, $state, Get $get): void {
-                        $component->state(AttributeData::mutateConfigurationForForm(
-                            $get('type'),
-                            is_array($state) ? $state : [],
-                        ));
-                    })
                     ->key('configuration')
                     ->statePath('configuration'),
             ]);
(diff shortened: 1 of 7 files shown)


tazzytazzy:
I slimmed down the changes in #2443 to be just the PR. Some reason, i added fixes for other things that should have been seperated PRs. I'll start creating those now.

alecritson:
Thanks @tazzytazzy 

I noticed on the attribute editing screen that if you try and edit a dropdown attribute the options appear as empty

image

Not sure if related to this PR specifically however I feel like it might be useful to investigate as part of this PR.

tazzytazzy:
Yea. I had a patch to that, but then when i cleaned up my other PR, i think it was lost. I'll poke around and try to dig it up again.  The issue when away from my side. I'll dig around my past local commits.

tazzytazzy:
Found it. Tested by adding, removing, trying to add duplicate rows, removing rows in the middle, etc.

I think a future feature would be nice to re-order the values. :-)

Also updated the tests, I'm hoping the updated test is valid.

Tests:    1 skipped, 769 passed (2815 assertions)
Duration: 193.48s

alecritson (approved):

Issue #2450, opened 2026-04-18
File attribute images are not reloaded after save in admin panel


Huncsuga:
- Lunar version: 1.5.0-beta.2
- Laravel Version: 12.56.0
- PHP Version: 8.4.20
- Database Driver & Version: MySQL 8.4

### Expected Behaviour:
Uploaded images should persist after saving and remain visible in the file attribute field within the admin panel.

### Actual Behaviour:
Images are successfully uploaded and stored on disk, but after clicking save, they disappear from the admin panel and the file field appears empty, as if no files were uploaded.

### Steps To Reproduce:
1. Create a Lunar\FieldTypes\File attribute (e.g. for a collection)
2. Set the type to image/*, min files: 1, max files: 5
3. Go to the collection edit page and upload 2–3 images
4. Save the changes

After saving, the uploaded images are no longer visible in the file field, although they are correctly stored on disk.

glennjacobs:
This should be fixed

PR #2451, opened 2026-04-18
Fix/file attribute images not persisting after save

Huncsuga:
The issue was caused by overriding getKey(), which broke Filament 4’s schema-based component resolution, preventing previously uploaded files from being reloaded in the FileUpload field. The fix restores proper key generation by using Filament’s native key() method, preserving the component hierarchy while still ensuring unique keys. This resolves the issue reported in #2450
.


packages/admin/src/Support/Forms/Components/Attributes.php
```diff
@@ -108,6 +104,7 @@ protected function setUp(): void
     {
         parent::setUp();
 
+        $this->key('attributeData');
```
alecritson, line 107: This will override the `using` method which sets a model specific key. Without then when Product and Product Variant attributes are on the same page it will cause issues.

glennjacobs (approved):

PR #2454, opened 2026-04-20
Optimize loading in search indexers

wychoong:
default loading is retrieving all fields of all variants
this changes only select only `sku` as only using `sku` in the indexer

could potentially breaking change if userland extended ProductIndexer and accessing other fields from $model->variants

TODO:
- [x] customer indexer 
- [x] order indexer
- [x] product option indexer
- [x] document the changes in changelog https://github.com/lunarphp/docs/pull/38


wychoong:
@glennjacobs appreciate your input on this before working on it

glennjacobs:
> @glennjacobs appreciate your input on this before working on it

As we encourage devs to make their own indexers, do you think this is important?

wychoong:
> > @glennjacobs appreciate your input on this before working on it
> 
> As we encourage devs to make their own indexers, do you think this is important?

I think is the out of the box experience. 
In my use case, I have a project is to sync data from external system, where a product can up to 500 variants, and on top of that it needs to merge a few products into single product in lunar. So when the catalog is huge and fetching a lot of data from relationships, the scout command just died halfway. 

For now I replaced with own indexer, with the minor change as per this PR, because the default already good enough. 

For the normal use case probably doesn’t feel it.

glennjacobs:
The breaking change is my concern here. I'm torn :-) I guess v1.5 is a big update, so what's one more breaking change?!

wychoong:
> The breaking change is my concern here. I'm torn :-) I guess v1.5 is a big update, so what's one more breaking change?!

I think it should be easy upgrade as long as documented. 

Optimized for those using default. Those making their own would be a minor change in their custom indexer.

glennjacobs:
@wychoong I'm happy for you to finish up this PR

wychoong:
> @wychoong I'm happy for you to finish up this PR

done, with the changelog updated https://github.com/lunarphp/docs/pull/38

glennjacobs (approved):

PR #2458, opened 2026-04-25
Fix Price object cast error when editing tier and customer group prices

krlmrr:
## Summary

Fixes a regression in 1.5.0-beta.2 / beta.3 where opening the edit modal for a tier price or customer group price throws:

> ErrorException: Object of class Lunar\DataTypes\Price could not be converted to float
> at vendor/filament/schemas/src/Components/StateCasts/NumberStateCast.php:28

This is the same ordering bug as #2441 — Filament v4's `NumberStateCast` (registered by `->numeric()`) runs `floatval()` on the raw state during hydration, before `formatStateUsing` is given a chance to unwrap the value. The two relation managers that bind a `Lunar\DataTypes\Price` object straight to a numeric `TextInput` were not covered by that fix.

PR #2441 solved the equivalent problem for `FieldType` attribute values by overriding `hydrateState`. For prices, the values arrive through plain record attributes, so `EditAction::mutateRecordDataUsing()` runs early enough to convert the `Price` object into its decimal value before child component hydration — which means the `formatStateUsing` calls can be removed entirely.

## Changes

- `PriceRelationManager` and `CustomerGroupPricingRelationManager`: add `mutateRecordDataUsing` on the `EditAction` to convert `Price` objects to their decimal value via a small `unwrapPriceData` helper.
- Drop the now-redundant `formatStateUsing` callbacks on `price` / `compare_price` text inputs.
- Add Pest feature tests that mount the edit action against a stored price and verify the form receives the expected decimal values (these tests fail on `1.x` without the fix with the original exception).

## Test plan

- [x] `vendor/bin/pest --testsuite=admin --group=resource.product`
- [x] `vendor/bin/pest --testsuite=admin --filter='customer group price|tier price'`
- [x] `vendor/bin/pint --format agent` on touched files
- [x] Confirmed the new tests fail on `1.x` without the fix (reproducing the original `Object of class Lunar\DataTypes\Price could not be converted to float` exception)

### DIFF
--- a/packages/admin/src/Filament/Resources/ProductResource/RelationManagers/CustomerGroupPricingRelationManager.php
+++ b/packages/admin/src/Filament/Resources/ProductResource/RelationManagers/CustomerGroupPricingRelationManager.php
@@ -18,6 +18,7 @@ use Illuminate\Database\Eloquent\Model;
 use Illuminate\Validation\Rules\Unique;
 use Lunar\Admin\Events\ProductPricingUpdated;
 use Lunar\Admin\Support\RelationManagers\BaseRelationManager;
+use Lunar\DataTypes\Price as PriceDataType;
 use Lunar\Facades\DB;
 use Lunar\Models\Currency;
 use Lunar\Models\CustomerGroup;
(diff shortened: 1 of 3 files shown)


glennjacobs (approved):

glennjacobs:
Thanks for the fix @krlmrr

PR #2461, opened 2026-05-05
Support collection rewards in BuyXGetY discounts

ryanmitchell:
This PR updates BuyXGetY discounts to support collections as a reward type, allowing more flexibility for editors.

It also fixes some regressions in BuyXGetY from the 1.5 update, specifically:
- Brings collection conditions back in the UI, as it wasn't being displayed
- Ensures collection conditions save to discountables


glennjacobs:
Nice work, this fills an obvious gap. A few things I'd like to see before this goes in:

## N+1 on the rewards filter

BuyXGetY.php:124-128 calls `$line->purchasable->product->collections->pluck('id')` per line inside the closure, which lazy-loads collections for each product on every cart calculation. Can we hoist the reward collection ids out of the loop and test membership against a pre-built set? Same pattern applies to the conditions filter on line 77, though that one's pre-existing.

## Null guard in processAutomaticRewards

BuyXGetY.php:227 — discountableRewards->random()->discountable can return null (orphaned morph row, or a soft-deleted Collection via NodeTrait). When that happens the instanceof LunarCollection check is false and we fall into the else branch, which then calls ->variants on null. A quick if `(! $selectedRewardItem) { $remainingRewardQty--; continue; }` before the instanceof would cover it.

## Required field on the condition select

CollectionConditionRelationManager.php:50 — the Select for discountable_id isn't marked `->required()`. The sibling ProductConditionRelationManager uses MorphToSelect which enforces it, but a plain Select won't.

## Nits (take or leave):
- ProductRewardRelationManager.php:88 — `titleAttribute('id')` for the Collection type looks like a copy-paste, 'name' would be more consistent with the others (the callbacks override it anyway).
- `?->` on `$record->discountable->attr('name')` in both relation managers as belt-and-braces.

Tests look good 👍

ryanmitchell:
Thanks for the review, all changes made as suggested.

glennjacobs (approved):

PR #2467, opened 2026-05-18
Convert Stripe amounts for zero-decimal currencies

### LINKED ISSUE #2272: Stripe payment amount incorrect for zero-decimal currency HUF
- Lunar version: 1.0
- Laravel Version: 11.45
- PHP Version: 8.4
- Database Driver & Version: MYSQL 8.0.41

### Expected Behaviour:

units without dividing or multiplying unnecessarily.
For example, if the cart total is 11480 HUF, the Stripe PaymentIntent should be created with amount: 11480.

### Actual Behaviour:

Currently, the amount appears to be treated as if HUF had decimals, causing Stripe to interpret the amount incorrectly.
Example:
	•	Cart total: 11480 HUF
	•	Stripe PaymentIntent created with amount: 11480 → Stripe treats as 114.80 HUF instead of 11480 HUF.

### Steps To Reproduce:

	1.	Create a cart in Lunar with currency HUF and total 11480 HUF.
	2.	Go to checkout using the Lunar Stripe integration.
	3.	Inspect the created Stripe PaymentIntent in the dashboard or via webhook logs.
	4.	Observe that Stripe interprets the amount as if it had two decimals, leading to wrong charged amount or error (“Amount must be at least 175.00 Ft”).

	•	According to Stripe docs: https://docs.stripe.com/currencies, HUF is a zero-decimal currency and should be passed in minor units without further multiplication/division.
	•	Possible fix: Handle HUF in StripeManager::buildIntent() and related amount conversion methods, ensuring amount is passed correctly for zero-decimal currencies.

Proposed Fix

Add amount normalization hooks to the Stripe manager layer (and to the charge‑store path), then use them consistently from buildIntent() and syncIntent() as well as when persisting charges to transactions.

1) Manager: per‑currency toStripeAmount() + usage in buildIntent() & syncIntent()

```
// App\Services\Stripe\StripeManager.php (example override of vendor manager)
namespace App\Services\Stripe;

use Lunar\Stripe\Managers\StripeManager as BaseStripeManager;
use Lunar\Models\Contracts\Cart as CartContract;
use Stripe\PaymentIntent;

class StripeManager extends BaseStripeManager
{
    /**
     * Normalize amount for Stripe per currency.
     * (Example implementation based on our current workaround.)
     */
    protected function toStripeAmount(int $value, string $currency): int
    {
        $currency = strtolower($currency);

        // Example: project-specific handling (workaround)
        // NOTE: If Lunar guarantees minor units already, this can be passthrough.
        if (in_array($currency, ['huf', 'twd'])) {
            return $value * 100;
        }

        return $value;
    }

    protected function buildIntent(int $value, string $currencyCode, array $opts = []): PaymentIntent
    {
        $amount = $this->toStripeAmount($value, $currencyCode);

        $params = [
            'amount' => $amount,
            'currency' => strtolower($currencyCode),
            'automatic_payment_methods' => ['enabled' => true],
            'capture_method' => config('lunar.stripe.policy', 'automatic'),
        ];

        return PaymentIntent::create([
            ...$params,
            ...$opts,
        ]);
    }

    public function syncIntent(CartContract $cart): void
    {
        $intentId = $this->getCartIntentId($cart);
        if (! $intentId) return;

        $cart = $cart->calculate();

        $this->getClient()->paymentIntents->update(
            $intentId,
            [
                'amount'   => $this->toStripeAmount($cart->total->value, $cart->currency->code),
                'currency' => strtolower($cart->currency->code),
            ]
        );
    }
}
```

2) Charge → Transaction: per‑currency fromStripeAmount()

```
// App\Services\Stripe\StoreCharges.php (example)
namespace App\Services\Stripe;

use Illuminate\Support\Collection;
use Lunar\Models\Contracts\Order as OrderContract;
use Lunar\Models\Order;
use Lunar\Models\Transaction;

class StoreCharges
{
    /**
     * Convert Stripe charge amount back to app units per currency.
     * (Example implementation matching the manager above.)
     */
    protected function fromStripeAmount(int $amount, string $currency): int
    {
        $currency = strtolower($currency);

        // Example: reverse of the workaround above
        if (in_array($currency, ['huf', 'twd'])) {
            return intdiv($amount, 100);
        }

        return $amount;
    }

    public function store(OrderContract $order, Collection $charges)
    {
        /** @var Order $order */
        if ($charges->isEmpty()) {
            return $order;
        }

        $transactions = $order->transactions()->get();

        foreach ($charges as $charge) {
            $timestamp = now()->createFromTimestamp($charge->created);

            $transaction = $transactions->first(
                fn ($t) => $t->reference == $charge->id
            ) ?: new Transaction;

            $type = $charge->refunded ? 'refund' : ($charge->captured ? 'capture' : 'intent');

            $paymentType    = collect($charge->payment_method_details)->keys()->first();
            $paymentDetails = collect($charge->payment_method_details)->first();

            $lastFour = $paymentDetails['last4']   ?? null;
            $cardType = $paymentDetails['brand']   ?? $paymentType;
            $meta     = (array) ($paymentDetails['checks'] ?? []);

            $transaction->fill([
                'order_id'    => $order->id,
                'success'     => (bool) ! $charge->failure_code,
                'type'        => $type,
                'driver'      => 'stripe',
                'amount'      => $this->fromStripeAmount($charge->amount, $order->currency_code),
                'reference'   => $charge->id,
                'status'      => $charge->status,
                'notes'       => $charge->failure_message ?: $charge->description,
                'card_type'   => $cardType,
                'last_four'   => $lastFour,
                'captured_at' => $charge->amount_captured ? $timestamp : null,
                'meta'        => $meta,
            ]);

            $transaction->save();
        }

        return $order;
    }
}
```

AppServiceProvider:
```
$this->app->singleton(BaseStripeManager::class, HufStripeManager::class);
$this->app->bind(BaseStoreCharges::class, CustomStoreCharges::class);
```

glennjacobs:
## Summary

- Adds `StripeManager::toStripeAmount()` which converts a Lunar price value to the amount Stripe expects for the given currency.
- Uses it from `createIntent()` and `syncIntent()` so payment intents are created and synced with the correct amount.
- Adds unit-tests covering standard 2‑decimal, zero-decimal (JPY/KRW/VND), special "x100" zero-decimal (HUF/TWD/UGX), 3‑decimal (BHD), and the cases where Lunar's `Currency::decimal_places` is misconfigured.

## Why

Lunar stores prices as integers in its own smallest unit, derived from `Currency::decimal_places`. Stripe expects amounts in its own currency sub-unit, which differs for two groups of currencies:

- **Zero-decimal currencies** (JPY, KRW, VND, …) — the amount _is_ the whole-unit value.
- **Special "x100" zero-decimal currencies** (HUF, TWD, UGX) — Stripe still requires the amount to be a multiple of 100, even though the currency has no sub-unit. So 11,480 HUF must be sent as `1148000`.

Previously Lunar passed `$cart->total->value` straight to Stripe with no conversion, so a cart of 11,480 HUF (stored as `11480`) was sent as `11480` and Stripe interpreted it as **114.80 HUF**.

The new conversion routes through the major currency unit so it's correct regardless of how the merchant has configured `Currency::decimal_places` for the currency.

See <https://docs.stripe.com/currencies>.

Fixes #2272.

## Test plan

- [x] `vendor/bin/pest --testsuite=stripe` — 42 passing
- [ ] Smoke-test against a Stripe test account with a HUF cart

### DIFF
--- a/packages/stripe/src/Managers/StripeManager.php
+++ b/packages/stripe/src/Managers/StripeManager.php
@@ -5,6 +5,7 @@ namespace Lunar\Stripe\Managers;
 use Illuminate\Support\Collection;
 use Lunar\Models\Cart;
 use Lunar\Models\Contracts\Cart as CartContract;
+use Lunar\Models\Contracts\Currency as CurrencyContract;
 use Lunar\Stripe\Enums\CancellationReason;
 use Stripe\Charge;
 use Stripe\Exception\ApiErrorException;
@@ -85,7 +86,7 @@ class StripeManager
         }
 
         $paymentIntent = $this->buildIntent(
-            $cart->total->value,
+            static::toStripeAmount($cart->total->value, $cart->currency),
             $cart->currency->code,
             $opts
         );
@@ -154,7 +155,7 @@ class StripeManager
 
         $this->getClient()->paymentIntents->update(
             $intentId,
-            ['amount' => $cart->total->value]
+            ['amount' => static::toStripeAmount($cart->total->value, $cart->currency)]
         );
     }
 
@@ -214,6 +215,35 @@ class StripeManager
     public function getCharge(string $chargeId): Charge
     {
         return $this->getClient()->charges->retrieve($chargeId);
+    }
+
+    /**
+     * Convert a Lunar price value to the amount expected by Stripe.
+     *
+     * Lunar stores prices as integers in the smallest unit configured by the
+     * Currency's decimal_places. Stripe expects amounts in its own currency
+     * sub-unit, which differs from Lunar's storage for zero-decimal currencies
+     * (e.g. JPY) and the "x100" zero-decimal special cases (HUF, TWD, UGX).
+     *
+     * @see https://docs.stripe.com/currencies
+     */
+    public static function toStripeAmount(int $value, CurrencyContract $currency): int
+    {
+        $code = strtolower($currency->code);
+
+        $stripeDecimals = match (true) {
+            in_array($code, ['huf', 'twd', 'ugx'], true) => 2,
+            in_array($code, ['bhd', 'jod', 'kwd', 'omr', 'tnd'], true) => 3,
+            in_array($code, [
+                'bif', 'clp', 'djf', 'gnf', 'jpy', 'kmf', 'krw',
+                'mga', 'pyg', 'rwf', 'vnd', 'vuv', 'xaf', 'xof', 'xpf',
+            ], true) => 0,
+            default => 2,
+        };
+
+        $majorAmount = $value / (10 ** $currency->decimal_places);
+
+        return (int) round($majorAmount * (10 ** $stripeDecimals));
     }
 
     /**
--- a/packages/stripe/src/Managers/StripeManager.php
+++ b/packages/stripe/src/Managers/StripeManager.php
@@ -220,30 +220,23 @@ class StripeManager
     /**
      * Convert a Lunar price value to the amount expected by Stripe.
      *
-     * Lunar stores prices as integers in the smallest unit configured by the
-     * Currency's decimal_places. Stripe expects amounts in its own currency
-     * sub-unit, which differs from Lunar's storage for zero-decimal currencies
-     * (e.g. JPY) and the "x100" zero-decimal special cases (HUF, TWD, UGX).
+     * For most currencies Stripe expects the amount in the same sub-unit
+     * Lunar already stores it in (controlled by `Currency::decimal_places`).
+     * HUF, TWD and UGX are the exception: although they are ISO zero-decimal
+     * currencies, Stripe requires amounts to be sent as if they had two
+     * decimal places.
      *
      * @see https://docs.stripe.com/currencies
      */
     public static function toStripeAmount(int $value, CurrencyContract $currency): int
     {
-        $code = strtolower($currency->code);
-
-        $stripeDecimals = match (true) {
-            in_array($code, ['huf', 'twd', 'ugx'], true) => 2,
-            in_array($code, ['bhd', 'jod', 'kwd', 'omr', 'tnd'], true) => 3,
-            in_array($code, [
-                'bif', 'clp', 'djf', 'gnf', 'jpy', 'kmf', 'krw',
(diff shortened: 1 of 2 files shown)


packages/stripe/src/Managers/StripeManager.php
```diff
@@ -216,6 +217,28 @@ public function getCharge(string $chargeId): Charge
...
+     * Lunar already stores it in (controlled by `Currency::decimal_places`).
+     * HUF, TWD and UGX are the exception: although they are ISO zero-decimal
+     * currencies, Stripe requires amounts to be sent as if they had two
+     * decimal places.
+     *
+     * @see https://docs.stripe.com/currencies
+     */
+    public static function toStripeAmount(int $value, CurrencyContract $currency): int
+    {
+        if (! in_array(strtolower($currency->code), ['huf', 'twd', 'ugx'], true)) {
```
alecritson, line 233: Is there a better way to detect this rather than having a hard coded array? What conditions of the currency cause the problem and could that be used for the logic instead?

glennjacobs: As far as I'm aware, this is a Stripe specific thing, so can't be modelled any other way.

glennjacobs: Stripe's currency docs (https://docs.stripe.com/currencies) define three buckets:

1. Standard (2 decimals) — USD, EUR, GBP, etc. Amounts in cents/pence.
2. Zero-decimal — JPY, KRW, VND, etc. Amounts in the base unit (¥1000 = 1000).
3. Three-decimal — BHD, KWD, etc. Amounts in 1/1000, must end in 0.
4. Special cases: HUF, TWD, UGX — ISO 4217 says these are zero-decimal, but Stripe requires them as if they had 2 decimals (×100, must end in 00).

https://docs.stripe.com/currencies#special-cases

The "special cases" exist purely because of a Stripe API design choice — they explicitly call this out as a quirk in their docs. There's no intrinsic property of the Hungarian Forint that distinguishes it from the Japanese Yen mathematically; it's an arbitrary historical inconsistency in Stripe's API.

alecritson (approved):

PR #2468, opened 2026-05-18
Handle soft-deleted purchasables in cart lines

### LINKED ISSUE #2001: Soft delete on product variant causes an error on cart line
Hello,

- Lunar version: 1.0.0-beta.4
- Laravel Version: 11.29.0
- PHP Version: 8.3.12
- Database Driver & Version: MariaDB 10.6.19

### Expected Behaviour:

When soft deleting a product variant added to a cart, the line should be ignored when calculating the cart.

### Actual Behaviour:

An error happens:

> Lunar\Managers\PricingManager::for(): Argument #1 ($purchasable) must be of type Lunar\Base\Purchasable, null given, called in …/vendor/lunarphp/core/src/Pipelines/CartLine/GetUnitPrice.php on line 37

### Steps To Reproduce:

- Add a product variant to a cart
- Soft delete the product variant
- Re-Authenticate with the user
- Listener `Lunar\Listeners\CartSessionAuthListener` triggers a calculating for the cart
- The error happens

### LINKED ISSUE #2367: Delete variants while cart is not empty
Lunar version: #.#.#
Laravel Version: #.#.#
PHP Version: #.#.#
Database Driver & Version:
Expected Behaviour:
If a product variant is deleted while it is still present in a cart, the cart should either gracefully ignore/remove the cart line referencing the deleted variant, or display a user-friendly error message.

Actual Behaviour:
When a product variant is deleted but still referenced in a cart, an error occurs (such as a fatal exception or crash) when viewing or manipulating the cart.

Steps To Reproduce:
Add a product with a variant to the cart.
Delete the variant from the admin or directly in the database.
Try to view or update the cart.

glennjacobs:
## Summary

- `CartLine::purchasable()` now uses `morphTo()->withTrashed()` so a soft-deleted variant still resolves through the relation and the cart can be rendered/recalculated without crashing.
- `ValidateCartForOrderCreation` now fails with a clear `carts.line_unavailable` error if any line points at a trashed (or otherwise missing) purchasable, so checkout cannot proceed.
- New translation key `lunar::exceptions.carts.line_unavailable` (en).
- Tests covering both behaviours.

## Why

Variants still use the `SoftDeletes` trait and are soft-deleted in two admin flows:

- `ProductOptionsWidget` when an admin removes/trims variants while editing product options (`packages/admin/src/Filament/Resources/ProductResource/Widgets/ProductOptionsWidget.php:381,446`).
- `ProductObserver::deleted` cascades a soft-delete to the product's variants (`packages/core/src/Observers/ProductObserver.php:33`).

With the old `morphTo()` (no trashed scope), any pipeline that reads `$line->purchasable` got `null` and threw `TypeError`s (e.g. `GetUnitPrice::for()`), so an existing customer cart blew up the moment an admin trimmed/removed the corresponding variant.

Including trashed models in the relation keeps prices/totals computing; the new validator rule then blocks checkout cleanly rather than silently letting an order be placed for a removed variant.

Fixes #2001.
Fixes #2367.

## Test plan

- [x] `vendor/bin/pest --testsuite=core` — 505 passing
- [ ] Manual: add variant to cart → admin trims variant → cart renders, checkout shows "line unavailable" error


alecritson (dismissed):

glennjacobs:
@alecritson not sure I'm happy with this one now. It assumes a purchasable will have soft deletes, which they may not. I'm going to have a re-think.

alecritson (dismissed):

packages/admin/src/Filament/Resources/ProductResource.php
```diff
@@ -129,12 +130,65 @@ public static function getDefaultForm(Schema $schema): Schema
                     ->content(
                         __('lunarpanel::product.status.availability.customer_groups')
                     )->type('warning')->hidden(function (Model $record) {
+                        if ($record?->status != 'published') {
```
alecritson: Can probably remove this if statement and have the logic also on the return

packages/admin/src/Filament/Resources/ProductResource.php
```diff
@@ -129,12 +130,65 @@ public static function getDefaultForm(Schema $schema): Schema
...
+                            return true;
+                        }
+
                         return $record->customerGroups()->where('enabled', true)->count();
                     }),
+                Shout::make('product-no-default-customer-group')
+                    ->content(
+                        __('lunarpanel::product.status.availability.no_default_customer_group')
+                    )->type('warning')->hidden(function (Model $record) {
+                        if ($record?->status != 'published') {
```
alecritson: Same here, multiple `if` statements could be cleaned up.

packages/admin/src/Filament/Resources/ProductResource.php
```diff
@@ -129,12 +130,65 @@ public static function getDefaultForm(Schema $schema): Schema
...
+                            return true;
+                        }
+
+                        return (bool) CustomerGroup::getDefault();
+                    }),
+                Shout::make('product-hidden-from-guests')
+                    ->content(
+                        __('lunarpanel::product.status.availability.hidden_from_guests')
+                    )->type('warning')->hidden(function (Model $record) {
+                        if ($record?->status != 'published') {
```
alecritson: Here also, `if` statements could be cleaned up.

packages/admin/src/Filament/Resources/ProductResource.php
```diff
@@ -129,12 +130,65 @@ public static function getDefaultForm(Schema $schema): Schema
...
+                            return true;
+                        }
+
+                        $default = CustomerGroup::getDefault();
+
+                        if (! $default) {
+                            return true;
+                        }
+
+                        return $record->customerGroups()
```
alecritson: There is already a scope on the `HasCustomerGroups` trait we might be able to use here?

packages/admin/src/Filament/Resources/ProductResource.php
```diff
@@ -129,12 +130,65 @@ public static function getDefaultForm(Schema $schema): Schema
...
+                            ->where(function ($query) {
+                                $query->whereNull('ends_at')->orWhere('ends_at', '>=', now());
+                            })
+                            ->exists();
+                    }),
                 Shout::make('product-channels')
                     ->content(
                         __('lunarpanel::product.status.availability.channels')
                     )->type('warning')->hidden(function (Model $record) {
+                        if ($record?->status != 'published') {
```
alecritson: Simplify `if` statements on the return.

packages/core/src/Validation/CartLine/CartLineAvailability.php
```diff
@@ -0,0 +1,92 @@
...
+        $groups = $this->resolveCustomerGroups($cart);
+
+        if (! $channel || $groups->isEmpty()) {
+            return $this->pass();
+        }
+
+        $pivotTable = (new Product)->customerGroups()->getTable();
+        $now = now();
+
+        $available = Product::query()
```
alecritson: It feels like this is doing a lot and should either be extracted or attempt to use the customer group scope available if possible.

packages/admin/src/Filament/Resources/ProductResource.php
```diff
@@ -129,12 +130,65 @@ public static function getDefaultForm(Schema $schema): Schema
...
+                        }
+
+                        $default = CustomerGroup::getDefault();
+
+                        if (! $default) {
+                            return true;
+                        }
+
+                        return $record->customerGroups()
+                            ->where('customer_group_id', $default->id)
```
alecritson, line 172: You can pass the customer group through the `customerGroup` parameter of the scope.

packages/admin/src/Filament/Resources/ProductResource.php
```diff
@@ -129,12 +130,65 @@ public static function getDefaultForm(Schema $schema): Schema
...
+                            ->where(function ($query) {
+                                $query->where('enabled', true)->orWhere('visible', true);
```
alecritson, line 174: The scope does this by default.

packages/admin/src/Filament/Resources/ProductResource.php
```diff
@@ -129,12 +130,65 @@ public static function getDefaultForm(Schema $schema): Schema
...
+                            })
+                            ->where(function ($query) {
```
alecritson, line 176: The customer group scope accepts `startsAt` and `endsAt` parameters.

packages/admin/src/Filament/Resources/ProductResource.php
alecritson, line 171: You should be able to use the `$record->customerGroup(...)` scope.

alecritson (approved):

PR #2470, opened 2026-05-18
Apply discounts on guest carts when max_uses_per_user is set

### LINKED ISSUE #2442: Coupon Applies but there is no effect on prices/calculations
- Lunar version: 1.3.0
- Laravel Version: 12.56.0
- PHP Version: 8.3.0
- Database Driver & Version: MYSQL 8.2.0

### Expected Behaviour:
Discount applies, the calculations adjust as per discount config.

### Actual Behaviour:
Discount is applied/registered on the cart but the calculations remain unchanged.

### Steps To Reproduce:
Create a Discount, In Discount configuration, Set the ``Max uses per user`` to 1 or greater than 1.
Use the discount in your site/store.
The discount will be applied on the cart but the calculations will remain unchanged.

### Workaround:
Do not set  ``Max uses per user``. Leave it empty

Image
Image

glennjacobs:
## Summary

- `AbstractDiscountType::checkDiscountConditions()` now only runs the per-user usage check when the cart actually has a user. Guest carts with `max_uses_per_user` set no longer silently fail the condition.
- Added three tests covering guest cart, under-limit user, at-limit user.

## Why

The old condition was:

```php
if ($validMaxUses && $this->discount->max_uses_per_user) {
    $validMaxUses = $cart->user && ($this->usesByUser($cart->user) < $this->discount->max_uses_per_user);
}
```

If the cart had no user attached (guest checkout, or anonymous cart pre-login), `$cart->user` was `null`, so `$validMaxUses` became `false` and the discount was filtered out — but only at calculation time, after the coupon code had already been accepted onto the cart. From the customer's point of view the coupon "worked" but their total never changed. This is exactly what #2442 reported.

A per-user limit only makes sense when a user is known, so the new condition skips the check entirely for guest carts and preserves whatever `$validMaxUses` was already set to:

```php
if ($validMaxUses && $this->discount->max_uses_per_user && $cart->user) {
    $validMaxUses = $this->usesByUser($cart->user) < $this->discount->max_uses_per_user;
}
```

Logged-in users still hit the limit correctly — the third test asserts that an attached usage row blocks reuse.

Fixes #2442.

## Test plan

- [ ] Manual: discount with `max_uses_per_user = 1` on a guest cart → discount applies to total
- [ ] Manual: same discount on a logged-in user who has already used it → total unchanged

### DIFF
--- a/packages/core/src/DiscountTypes/AbstractDiscountType.php
+++ b/packages/core/src/DiscountTypes/AbstractDiscountType.php
@@ -84,8 +84,8 @@ abstract class AbstractDiscountType implements DiscountTypeInterface
 
         $validMaxUses = $this->discount->max_uses ? $this->discount->uses < $this->discount->max_uses : true;
 
-        if ($validMaxUses && $this->discount->max_uses_per_user) {
-            $validMaxUses = $cart->user && ($this->usesByUser($cart->user) < $this->discount->max_uses_per_user);
+        if ($validMaxUses && $this->discount->max_uses_per_user && $cart->user) {
+            $validMaxUses = $this->usesByUser($cart->user) < $this->discount->max_uses_per_user;
         }
 
         return $validCoupon && $validMinSpend && $validMaxUses;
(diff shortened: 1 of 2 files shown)


ryanmitchell:
Does this mean that the discount can be used indefinitely by a non-logged in user? 
If so would it not be better to reject the discount code if there is no logged in user.

PR #2471, opened 2026-05-18
Honour the discount stop flag in the apply loop

### LINKED ISSUE #2116: "Stop other discounts applying after this one" Feature Not Halting Further Discounts
- Lunar version: 1.0.0-beta.6
- Laravel Version: 11.43.2
- PHP Version: 8.3.15

### Expected Behaviour:
When multiple discounts are applied without a coupon code, then a discount has "Stop other discounts applying after this one" set to true, then no further discounts should be applied after it.

For example:

**Discount A:** 10% off, low priority, "Stop other discounts applying after this one" = true.
**Discount B:** 20% off, high priority, "Stop other discounts applying after this one" = true.

Only **Discount B** should be applied since it has higher priority and is set to stop further discounts.

### Actual Behaviour:
Even though "Stop other discounts applying after this one" is true for both discounts, the system still applies both. 

### Steps To Reproduce:
1. Create a discount with 10% off, set priority to low, and enable "Stop other discounts applying after this one". Do not set a max uses limit or coupon code.

2. Create another discount with 20% off, set priority to high, and also enable "Stop other discounts applying after this one". Do not set a max uses limit or coupon code.

3. Add a product to the cart.

4. Observe that both discounts are applied instead of only the higher-priority.

glennjacobs:
## Summary

- `DiscountManager::apply()` now checks `$discount->stop` after each discount runs, and breaks out of the loop when a stopping discount was actually applied.
- The "actually applied" check is done by comparing `$cart->discounts->count()` before and after the call, so a `stop=true` discount whose conditions fail (mismatched coupon, customer scope, etc.) does **not** halt the chain.
- Added three tests in `DiscountManagerTest.php`:
  - `stop=true` on a higher-priority discount halts a lower-priority one
  - `stop=true` on a discount whose conditions fail does not halt the next one
  - `stop=false` on the higher-priority discount lets the lower-priority one apply

## Why

The Discount model has had a `stop` boolean and an admin toggle labelled "Stop other discounts applying after this one" since 1.0, but the manager's `apply()` loop iterated every active discount unconditionally — the flag had no runtime effect, and multiple high-priority discounts all stacked even when one was explicitly marked as exclusive.

Issue #2116 reports this exact behaviour: configure two discounts both with `stop=true` and the system applies both instead of stopping at the higher-priority one.

The fix breaks the loop once a stopping discount has applied. Verified by `git stash`ing the change — the new "halts further discounts after a discount applies" test fails without it and passes with it.

Fixes #2116.

## Test plan

- [ ] Manual: configure two discounts as in the issue, see only the higher-priority one apply

### DIFF
--- a/packages/admin/resources/lang/en/discount.php
+++ b/packages/admin/resources/lang/en/discount.php
@@ -29,21 +29,11 @@ return [
         ],
         'priority' => [
             'label' => 'Priority',
-            'helper_text' => 'Discounts with higher priority will be applied first.',
-            'options' => [
-                'low' => [
-                    'label' => 'Low',
-                ],
-                'medium' => [
-                    'label' => 'Medium',
-                ],
-                'high' => [
-                    'label' => 'High',
-                ],
-            ],
+            'helper_text' => 'A higher number means higher priority. Discounts with a higher priority are applied first. Must be between 1 and 100.',
         ],
         'stop' => [
             'label' => 'Stop other discounts applying after this one',
+            'helper_text' => 'When this discount applies, any discount with a lower priority will be skipped. Give discounts different priorities to control the order they apply in.',
         ],
         'coupon' => [
             'label' => 'Coupon',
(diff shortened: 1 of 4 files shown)


alecritson (dismissed):

glennjacobs:
@alecritson made this change to make it clear how it works.

image

glennjacobs:
@ryanmitchell as you are Mr Discounts, are you happy with this? :)

ryanmitchell:
Yes this works well.

PR #2475, opened 2026-05-18
Harden telemetry rate-limiting against missing cache and failed HTTP

### LINKED ISSUE #2410: 429 too many attempts error when cache driver not set
This is due to the Lunar telemetry constantly firing, rather than being limited to once a day.

### LINKED ISSUE #2465: Telemetry runs on request termination and crashes in stancl/tenancy central context by querying Product::count() on the wrong database
## Summary

When using `lunarphp/core` in a multi-tenant application powered by `stancl/tenancy`, Lunar telemetry runs on request termination and attempts to query `Lunar\Models\Product::count()` using the current/default database connection.

In a typical central/tenant database setup, Lunar tables exist only in tenant databases, not in the central database. As a result, telemetry triggers a SQL exception on central requests.

## Packages / Versions

We are using:

- `stancl/tenancy`: `^3.10.0`
- `lunarphp/core`: `>=1.3 <1.5`

*Also observed in a Laravel 11 app with Octane/FrankenPHP.*

## What happens

Lunar registers this terminating callback in `LunarServiceProvider`:

```php
$this->app->terminating(function () {
    if (! app()->runningInConsole()) {
        Telemetry::run();
    }
});
```

During `Telemetry::run()`, Lunar eventually calls:

```php
Lunar\Models\Product::count()
```

In our setup, this happens after the tenant context has already reverted to the central application context, so the query runs against the central database instead of the tenant database.

Because Lunar tables are only migrated in tenant databases, the central DB does not have `market_products` (or `lunar_products`), and the request ends with an exception.

## Error

```sql
SQLSTATE[42S02]: Base table or view not found: 1146 Table 'app.market_products' doesn't exist
(Connection: mysql, SQL: select count(*) as aggregate from `market_products` where `market_products`.`deleted_at` is null)
```

## Relevant Stack Trace

```text
Lunar\Models\Product::__callStatic
Lunar\Base\TelemetryInsights->productCount()
Lunar\Base\TelemetryService->getInsightsPayload()
Lunar\Base\TelemetryService->run()
Lunar\LunarServiceProvider->{closure}
Illuminate\Foundation\Http\Kernel->terminate()
Laravel\Octane\ApplicationGateway->terminate()
```

## Expected behavior

Telemetry should not break requests in applications where:
- Lunar tables are only present in tenant databases.
- The current request is running in a central/non-tenant context.
- Or the current connection does not have Lunar tables.

## Actual behavior

Telemetry runs unconditionally on HTTP request termination and performs product counts against the wrong connection, causing SQL errors.

## How to reproduce

1. Install `lunarphp/core`.
2. Use `stancl/tenancy` with separate central and tenant databases.
3. Run Lunar migrations only for tenant databases.
4. Trigger a central HTTP request that touches tenant logic during the request lifecycle.
5. Let the request finish.
6. On termination, Lunar telemetry runs and queries `Product::count()` on the central DB.
7. The request/log ends with a SQL error because the central DB does not contain Lunar tables.

## Proposed fix

Any of the following would solve the issue safely:

### Option 1: Make telemetry opt-in / configurable
Allow disabling telemetry entirely via config/env, e.g.:

```php
'lunar.telemetry.enabled' => false
```

### Option 2: Skip telemetry when not in a tenant-aware / valid Lunar DB context
Before running telemetry, check whether the current connection is suitable for Lunar queries. For example:
- Ensure the current request is in the intended tenant context.
- Or ensure the Lunar products table exists before calling `Product::count()`.

### Option 3: Fail silently
Wrap telemetry execution in a try/catch so telemetry never breaks application requests. Something like:

```php
$this->app->terminating(function () {
    if (app()->runningInConsole()) {
        return;
    }

    try {
        Telemetry::run();
    } catch (\Throwable $e) {
        // swallow/log silently, but do not break the request
    }
});
```

glennjacobs:
## Summary

- `TelemetryService::run()` now records the daily attempt **before** the HTTP call, so a failed/slow request still counts and the next request termination won't hammer the endpoint again.
- Any exception thrown by the HTTP client (timeout, DNS failure, connection refused) **or by `getInsightsPayload()`** is swallowed — telemetry must never bubble out of a request-termination callback.
- `TelemetryService::shouldRun()` returns `false` when the cache store is a `NullStore` (or otherwise non-persistent), since rate-limiting cannot work without persistent storage.

## Why

The old flow was `Cache::remember(key, ttl, fn)` keyed on the HTTP response: success returned `now()` and the value was cached for 24h; failure returned `null` and `null` got cached. On the next request `shouldRun()` saw a falsy cache value and re-fired the request immediately — eventually hitting `stats.lunarphp.io`'s 429 rate limit, which is what reporter #2410 ran into.

The same problem affects any setup where the cache store doesn't persist between requests (the `null` driver, an unmigrated database cache table, etc.) — the marker never sticks so telemetry fires on every request termination.

The new `try { ... } catch (\Throwable $e)` also covers the `Product::count()` failure that #2465 reports under `stancl/tenancy` — when the terminating callback fires in the central context after the tenant has been switched out, the insights query runs against the wrong connection and throws. That exception now no longer bubbles out of the terminating callback, and because the cache marker is set before the try, the failing run still counts towards the daily limit.

Recording the attempt up-front makes throttling robust regardless of HTTP outcome, and explicitly detecting `NullStore` keeps misconfigured apps from drowning their own logs (and ours).

Fixes #2410.
Fixes #2465.

## Test plan

- [ ] Manual: set `CACHE_DRIVER=null` → `Telemetry::shouldRun()` returns false, no HTTP request fires
- [ ] Manual (multi-tenant): hit a central-context request that previously crashed in `Telemetry::run()` → request completes without the SQL exception bubbling out of the terminating callback

### DIFF
--- a/packages/core/src/Base/TelemetryService.php
+++ b/packages/core/src/Base/TelemetryService.php
@@ -2,7 +2,7 @@
 
 namespace Lunar\Base;
 
-use Carbon\Carbon;
+use Illuminate\Cache\NullStore;
 use Illuminate\Support\Facades\Cache;
 use Illuminate\Support\Facades\Http;
 
@@ -38,6 +38,10 @@ class TelemetryService implements TelemetryServiceInterface
             return false;
         }
 
+        if (Cache::getStore() instanceof NullStore) {
+            return false;
+        }
+
         $lastAttempt = Cache::get($this->getCacheKey());
 
         return ! $lastAttempt || ! now()->parse($lastAttempt)->isToday();
@@ -67,15 +71,17 @@ class TelemetryService implements TelemetryServiceInterface
             return;
         }
 
-        Cache::forget($this->getCacheKey());
+        // Record the attempt up-front so a failed or slow HTTP request still
+        // counts against the daily limit instead of retrying on every hit.
+        Cache::put($this->getCacheKey(), now()->toIso8601String(), now()->endOfDay());
 
-        Cache::remember($this->getCacheKey(), 86400, function (): ?Carbon {
-            $response = Http::withHeader('Accept', 'application/json')
+        try {
+            Http::withHeader('Accept', 'application/json')
                 ->timeout(3)
                 ->retry(3, 100)
                 ->post($this->getInsightsUrl(), $this->getInsightsPayload());
-
-            return $response->successful() ? now() : null;
-        });
+        } catch (\Throwable $e) {
+            // Telemetry must never affect the surrounding request.
+        }
     }
 }
--- a/packages/core/src/Base/TelemetryService.php
+++ b/packages/core/src/Base/TelemetryService.php
@@ -34,11 +34,7 @@ class TelemetryService implements TelemetryServiceInterface
 
     public function shouldRun(): bool
     {
-        if (! $this->shouldRun) {
-            return false;
-        }
-
-        if (Cache::getStore() instanceof NullStore) {
+        if (! $this->shouldRun || Cache::getStore() instanceof NullStore) {
             return false;
         }
 
@@ -73,7 +69,11 @@ class TelemetryService implements TelemetryServiceInterface
 
         // Record the attempt up-front so a failed or slow HTTP request still
         // counts against the daily limit instead of retrying on every hit.
-        Cache::put($this->getCacheKey(), now()->toIso8601String(), now()->endOfDay());
+        Cache::put(
+            $this->getCacheKey(),
+            now()->toIso8601String(),
+            now()->endOfDay(),
+        );
 
         try {
             Http::withHeader('Accept', 'application/json')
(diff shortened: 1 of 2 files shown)


packages/core/src/Base/TelemetryService.php
```diff
@@ -38,6 +38,10 @@ public function shouldRun(): bool
             return false;
         }
 
+        if (Cache::getStore() instanceof NullStore) {
```
alecritson: This could be put into the `if` statement above, using a `||`, to keep things cleaner.

glennjacobs: Did a little tidy-up

alecritson (approved):

PR #2476, opened 2026-05-18
Reset the shipping breakdown on every cart calculate

### LINKED ISSUE #2199: Shipping Option Change adds up the costs
- Lunar version: 1.x-dev
- Laravel Version: 12.15.0
- PHP Version: 8.3.17

### Expected Behaviour:
In a simple Shipping Option form, when you change the selected option, it overrides the last selected and applies the new price to the total.

### Actual Behaviour:
When you set a new shipping option via ``CartSession::setShippingOption($option)`` it saves the value correctly but add the price to the current total, instead of replace the current shipping total it adds up.

This error just last for the current request, after  a refresh the totals are ok.

### Steps To Reproduce:
Call ``CartSession::setShippingOption()`` twice in a row and see the car total,  shipping total and shipping break down, 

Looking the [code](https://github.com/lunarphp/lunar/blob/e15acf62981f27408ab2e353b81d9d8d5fc8ea8a/packages/core/src/Pipelines/Cart/ApplyShipping.php#L43) I see there is something happening with shipping override and shipping breakdowns.

In the docs and the code it's not clear how ``setShippingOption`` should work, it's a ``set`` it shouldn't add the shipping option but reset it.

glennjacobs:
## Summary

- `ApplyShipping` now constructs a fresh `ShippingBreakdown` on each run rather than mutating the cart's previously-cached one.
- The previous `if (\$cart->shippingOptionOverride) { \$shippingBreakdown->items = collect(); }` clear is removed — it was a partial workaround for the same issue.
- Adds a regression test that swaps the cart's shipping option mid-request and asserts both the breakdown item count and shipping subtotal reflect only the latest option.

## Why

`ApplyShipping` was seeded from `\$cart->shippingBreakdown ?: new ShippingBreakdown` and then `put()`'d the resolved option into the existing items collection. Calling `CartSession::setShippingOption()` (or otherwise recalculating the cart) within the same request added the new option to the items collection alongside the previous one — `\$shippingBreakdown->items->sum('price.value')` then summed both. The page refresh fix that reporters noticed was because `shippingBreakdown` is runtime-only and disappears between requests.

The breakdown is purely a derived view of the single currently-selected option, so starting fresh every time is both correct and simpler.

Fixes #2199.

## Test plan

- [x] `vendor/bin/pest --testsuite=core` — 504 passing
- [ ] Manual: call `CartSession::setShippingOption()` twice in a row with different options → cart total reflects only the latest

### DIFF
--- a/packages/core/config/cart.php
+++ b/packages/core/config/cart.php
@@ -12,6 +12,7 @@ use Lunar\Pipelines\Cart\ApplyDiscounts;
 use Lunar\Pipelines\Cart\ApplyShipping;
 use Lunar\Pipelines\Cart\Calculate;
 use Lunar\Pipelines\Cart\CalculateLines;
+use Lunar\Pipelines\Cart\CalculateShippingSubTotal;
 use Lunar\Pipelines\Cart\CalculateTax;
 use Lunar\Pipelines\CartLine\GetUnitPrice;
 use Lunar\Pipelines\CartPrune\PruneAfter;
@@ -64,6 +65,7 @@ return [
         'cart' => [
             CalculateLines::class,
             ApplyShipping::class,
+            CalculateShippingSubTotal::class,
             ApplyDiscounts::class,
             CalculateTax::class,
             Calculate::class,
--- a/packages/core/src/Pipelines/Cart/ApplyShipping.php
+++ b/packages/core/src/Pipelines/Cart/ApplyShipping.php
@@ -20,7 +20,6 @@ final class ApplyShipping
     public function handle(CartContract $cart, Closure $next): mixed
     {
         /** @var Cart $cart */
-        $shippingSubTotal = 0;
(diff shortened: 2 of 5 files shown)


alecritson:
In this case, what happens if a developer adds a custom pipeline class that needs to add additional shipping costs/breakdowns before they are calculated?

glennjacobs:
> In this case, what happens if a developer adds a custom pipeline class that needs to add additional shipping costs/breakdowns before they are calculated?

The post-ApplyShipping slot is the supported extension point for additional shipping costs; that still works. The pre-ApplyShipping slot was never reliable (override path already cleared it) and is the right thing to break in service of fixing the duplicate-option bug. If they want a documented hook for pre-resolution adjustments, that's a separate feature, not something to revert this fix for.

alecritson (approved):

PR #2477, opened 2026-05-18
Normalise tag values via a model mutator

### LINKED ISSUE #2401: Tag values not converted to uppercase on save
- Lunar version: 1.2.1
- Laravel Version:  11.47.0
- PHP Version: 8.4.16
- Database Driver & Version: sqlite

### Expected Behaviour:
When creating a new tag from Filament tag resource for it to be converted to uppercase.

### Actual Behaviour:
Value remains lowercase.

### Steps To Reproduce:
Log into Lunar admin panel, got to tags, add new tag in lowercase, save.

Image

Image

### Notes

This is not an issue when adding tags from the edit product page. Seems the TagFactory.php runs in that case but not when creating a new tag from the tag resource.

Image

glennjacobs:
## Summary

- Adds a `value` mutator on the `Tag` model that uppercases on set (null preserved).
- Adds a `Tag` model test covering both casing normalisation and null handling.

## Why

Tag values added via the product edit page were uppercased by `SyncTags::handle()`, but tags created from the standalone `TagResource` went straight in unchanged. Two entry points, two different stored forms for the same logical tag — cross-referencing them stopped working.

Pushing the rule onto the model is the smallest centralisation that fixes both entry points. `SyncTags`' explicit `Str::upper()` becomes a no-op (left in place — defensive and behaviourally equivalent), and any future entry point that creates `Tag` rows automatically inherits the same casing rule.

Matching the existing **uppercase** convention rather than switching to lowercase, because:

1. Existing production data is already uppercase (from `SyncTags`).
2. Switching to lowercase would split tag values across two cases and break tag matching until a data migration runs.

If lowercase is preferred long-term, that's better tackled as a deliberate data migration + mutator flip in a separate PR.

Fixes #2401.

## Test plan

- [x] `vendor/bin/pest --testsuite=core` — 505 passing
- [ ] Manual: create `Foo` in `Tags` resource → DB stores `FOO`; create another product tag `foo` from product page → matches the same row

### DIFF
--- a/packages/admin/src/Support/Forms/Components/Tags.php
+++ b/packages/admin/src/Support/Forms/Components/Tags.php
@@ -42,7 +42,7 @@ class Tags extends TagsInput
 
     public function syncTags(?Model $record, array $state): void
     {
-        // value fields is utf8mb4_unicode_ci prevent no case sensitive
+        // Tag::value has an uppercasing mutator, but we still upper here so the whereIn lookup below matches existing rows under case-sensitive collations.
         $state = collect($state)->map(function (string $value) {
             return Str::upper($value);
         })->toArray();
--- a/packages/core/database/factories/TagFactory.php
+++ b/packages/core/database/factories/TagFactory.php
@@ -2,7 +2,6 @@
 
 namespace Lunar\Database\Factories;
 
-use Illuminate\Support\Str;
 use Lunar\Models\Tag;
 
 class TagFactory extends BaseFactory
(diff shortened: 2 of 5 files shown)


alecritson:
Are there currently any places where we are doing this manually we can remove?

glennjacobs:
> Are there currently any places where we are doing this manually we can remove?

Pushed a few tweaks based upon this feedback.

alecritson (approved):

PR #2491, opened 2026-05-24
Add skill for laravel boost

punyflash:
This PR adds skill for LunarPHP that may be used in Laravel Boost:

```sh
php artisan boost:add-skill lunarphp/lunar
```


glennjacobs:
Thanks @punyflash 

There's some opinionated bits in this, e.g. Laravel Sail for example, which it probably shouldn't have.

wychoong:
only have a few suggestions
- installation guide seems redundant as the skill wont be available before install
- break the skills into references like how laravel skills are, so can go in-depth in future
- might be great to incorporate https://docs.lunarphp.com/1.x/guides guides, how to avoid duplicate effort in package & docs

glennjacobs:
Thanks for your patience on this @punyflash — it sat far longer than it should have, sorry.

I've pushed changes straight to the branch rather than send you round another review cycle, since the two things that needed fixing weren't really review comments so much as information you had no way of having. Here's the reasoning.

**Location.** The skill now lives at `packages/core/resources/boost/skills/lunar/`. Boost has two separate ways of picking up skills: `SkillComposer::getThirdPartySkills()` scans `vendor/<pkg>/resources/boost/skills/` and installs from there automatically on `boost:install` and `boost:update`, while `boost:add-skill` fetches from a repo's default branch over the GitHub API. The second one is what a root-level `skills/` directory limits you to, and it isn't really viable for us — monorepo-builder only splits `packages/*` into the published packages, so a root directory never reaches anyone's `vendor/`. From `resources/boost/skills/` it ships inside `lunarphp/core`, versions with the release the user actually has installed, and needs no action from them at all.

**Size.** This is the part you couldn't have known: our docs are on Mintlify, which already publishes `llms.txt`, per-page markdown, an MCP server at `docs.lunarphp.com/mcp`, and — it turns out — a generated skill of its own at `mintlify://skills/lunar`. So the reference files were a second copy of material that already exists and is already kept in sync, and they'd have drifted from it within a release or two. The skill now routes to those sources instead and carries only the things worth having in context before the first attempt.

**What stayed.** Your `description` frontmatter, near enough intact — the activation wording was the right call and is doing the most important job in the file. And the pitfalls list, which was the genuinely valuable contribution here and is now most of the skill's body. You're credited as co-author on the commit.

Two things worth flagging from checking the content against `1.x`:

- The custom discount type example wouldn't have run — `DiscountTypeInterface::apply()` takes and returns `Lunar\Models\Contracts\Cart`, so typing the concrete model is a signature mismatch. That trap is now written up as a gotcha in its own right, since it applies to anything extending Lunar's interfaces.
- `search-docs` doesn't cover Lunar. `boost.laravel.com/api/docs` returns no results for `lunarphp/*`, so that line has been replaced with pointers to our MCP server.

For what it's worth, the "root collection must exist before children" pitfall was also wrong, but I only caught that because I'd carried it into the rewrite myself and then went back through every entry against source. `Collection::create()` with no parent becomes a root on its own.

Merging shortly. Thanks for kicking this off — the trigger wording and the pitfalls are the parts that were hard to get right, and you got both.

PR #2503, opened 2026-05-29
fix(attributes): fix hydrateState signature for Filament 4.11

bpotmalnik:
Filament 4.11 added `$shouldApplyStateCasts` and `$appliedStateCastPaths` parameters to `Component::hydrateState()`. Updates the override in `Attributes` to match, fixing a fatal error when viewing any resource that uses the Attributes component.

Introduced in: https://github.com/filamentphp/filament/pull/19911

### DIFF
--- a/packages/admin/src/Support/Forms/Components/Attributes.php
+++ b/packages/admin/src/Support/Forms/Components/Attributes.php
@@ -71,13 +71,13 @@ class Attributes extends Group
         $this->rawState($data);
     }
 
-    public function hydrateState(?array &$hydratedDefaultState, bool $shouldCallHydrationHooks = true): void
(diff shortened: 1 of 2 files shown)


bpotmalnik:
I believe it might be breaking change on their end for all 3rd party that overwrites this method. On the other hand it is architecturally internal method so I decided to just fix it! 🔧

glennjacobs:
> I believe it might be breaking change on their end for all 3rd party that overwrites this method. On the other hand it is architecturally internal method so I decided to just fix it! 🔧

Yeah, I don't think this is part of their public API as such. It's a constant battle we face. Thanks for the fix.

glennjacobs (approved):

PR #2504, opened 2026-05-29
fix: trim whitespace from string attributes on BaseModel

### LINKED ISSUE #1664: Input values aren't being trimmed or validated when saved on resource
- Lunar version: 1.xx

### Expected Behaviour:

Columns should be trimmed where appropriate.

### Actual Behaviour:

One instance found is the `handle` column on various tables can be saved as `       size`

### Steps To Reproduce:

Go to a resource with a `handle` column, add spaces, save, spaces should persist.

Looks like the `TrimStrings` middleware (even if added to the panel) is not being completed as the Livewire class `Livewire\Mechanisms\HandleRequests\HandleRequests` is causing it to be skipped.

bpotmalnik:
Laravel's TrimStrings middleware is intentionally skipped for Livewire requests (Livewire\Mechanisms\HandleRequests\HandleRequests calls TrimStrings::skipWhen). This means leading/trailing whitespace entered via Livewire forms (e.g. the admin panel) is never stripped before reaching the model, allowing values like "  size  " to be persisted.

Override setAttribute() on BaseModel to trim all incoming string values before delegating to the parent. This covers every model in Lunar (handle, sku, slug, coupon, code, reference, postcode, etc.)

Fixes #1664

### DIFF
--- a/packages/core/src/Base/BaseModel.php
+++ b/packages/core/src/Base/BaseModel.php
@@ -24,6 +24,11 @@ abstract class BaseModel extends Model
         }
     }
 
+    public function setAttribute($key, $value): mixed
+    {
+        return parent::setAttribute($key, is_string($value) ? trim($value) : $value);
+    }
+
     /**
      * {@inheritdoc}
      */
--- a/tests/core/Unit/Base/BaseModelTest.php
+++ b/tests/core/Unit/Base/BaseModelTest.php
@@ -5,6 +5,7 @@ use Illuminate\Support\Facades\Route;
 use Lunar\Base\BaseModel;
 use Lunar\Base\Traits\HasModelExtending;
 use Lunar\Models\Collection as ModelsCollection;
+use Lunar\Models\CustomerGroup;
 use Lunar\Models\Product;
 use Lunar\Models\Url;
 use Lunar\Tests\Core\TestCase;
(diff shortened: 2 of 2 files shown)


bpotmalnik:
I was considering if it is breaking change or not, but it would break things only if people would assert things agains hardcoded strings (which likelihood is pretty low) so I think it can got to v1.

wychoong:
I don’t think this is a suitable fix
- it’s too heavy handed (a Model::update([…]) will be affected too)
- Not the common livewire/filament form expectation
- Probably more suitable to implement Input::configureUsing( ->dehydrateStateUsing ) in your app to trim the fields

bpotmalnik:
@wychoong I hear you. I still think that reinstating [default Laravel behavior](https://laravel.com/docs/13.x/validation#a-note-on-optional-fields) would be handy for users (developers) using the package. 

As it is not indeed default filament behavior, but this package is kind of special case where lunar is in control of both core and admin panel itself (filament) which can drive handy behaviors.

glennjacobs:
Thank you @bpotmalnik for this — the diagnosis was spot on (Livewire deliberately skips `TrimStrings` via `skipRequestPayloadTamperingMiddleware`, so admin input never gets trimmed), and the fix worked well.

After weighing it up, we've decided to go with the form-layer approach instead, in #2642. Two things tipped the decision:

1. Filament v4 now ships a first-class API for exactly this (`->trim()` + `configureUsing()`, added in filamentphp/filament#16706 and documented as the equivalent of the `TrimStrings` middleware), which is essentially what @wychoong suggested on this thread.
2. Lunar 2.x is moving the admin to Inertia.js, where the standard `TrimStrings` middleware applies normally — so the gap this fixes is specific to the Livewire/Filament admin, and a `BaseModel::setAttribute()` override in core would outlive the problem it solves while also affecting every programmatic write path (jobs, imports, seeders).

Really appreciate the contribution and the discussion it prompted — this is what got the issue over the line. Closing in favour of #2642.

PR #2505, opened 2026-05-29
fix(attributes): expose and apply default_value in the admin panel

### LINKED ISSUE #1958: lunar_attributes.default_value is not respected
- Lunar version: 1.0.0-beta.2
- Laravel Version: 11.22.0
- PHP Version: 8.3.11

### Expected Behaviour:
In the `lunar_attributes` table there's a column `default_value`, which should be applied to new entities (e.g. products). This column should be included in the edit attribute modal as well.

### Actual Behaviour:
Neither the default value is applied to new entities, nor it could be edited in the admin panel.

### Steps To Reproduce:
1. In database, manually edit the value of `lunar_attributes`.`default_value` of any existing row. In my case it's a number attribute which `attribute_type` = 'product' and `type` = 'Lunar\FieldTypes\Number'.
2. Create a product in the admin panel

bpotmalnik:
Fixes: https://github.com/lunarphp/lunar/issues/1958


glennjacobs:
Thanks @bpotmalnik — good feature, and the seeding location in `AttributeData` was right. While testing it against the other field types we found the plain-string default couldn't work everywhere: Filament's state casts coerce `null` before the default branch runs (Toggle to `false`, KeyValue to `[]`), so a default on those types was silently discarded — and unvalidated input could corrupt data ('abc' on a Number attribute persists as `0` on every save; a File default persists a path that was never uploaded).

We've pushed a follow-up commit to your branch (and rebased it onto current `1.x`) that makes the feature type-aware:

- Field types opt in via `BaseFieldType::canHaveDefaultValue()` — enabled for Text, TranslatedText, Number and Dropdown; custom field types can opt in the same way.
- Per-type validation via `getDefaultValueValidationRules()`: `numeric` for Number, and Dropdown defaults must match a configured lookup value.
- The attribute form hides the field for unsupported types, and `AttributeData` skips seeding for them, so a legacy `default_value` on an unsupported type stays inert instead of misbehaving.

Your original tests are untouched and still pass, with five new ones covering the validation and visibility behaviour.

glennjacobs (approved):

PR #2506, opened 2026-05-29
fix(order-lines): resolve purchasable relationship for ShippingOption  lines

### LINKED ISSUE #2294: ShippingOption error when accessing purchasable relationship of OrderLine
- Lunar version: 1.0.0-beta20
- Laravel Version: 11.45.0
- PHP Version: 8.3.25

### Expected Behaviour:

When accessing the purchasable relationship of an OrderLine that has purchasable_type of ShippingOption, it should return the ShippingOption object without errors.

### Actual Behaviour:

When trying to access the purchasable relationship on an OrderLine object with a purchasable_type of Lunar\DataTypes\ShippingOption results in the following error:

`Too few arguments to function Lunar\DataTypes\ShippingOption::__construct(), 0 passed in /vendor/laravel/framework/src/Illuminate/Database/Eloquent/Concerns/HasRelationships.php on line 793 and at least 5 expected`

### Steps To Reproduce:

Ensure an OrderLine with a purchasable_type of Lunar\DataTypes\ShippingOption exists
Attempt to access the purchasable relationship on the OrderLine object (i.e. $orderLine->purchasable)
Observe the error above

This issue was mentioned before #1211. It was closed but the fix applied to it only just ignores the shipping option from the dashboard, it does not fix the bug. #1217

bpotmalnik:
Fixes #2294

Accessing $orderLine->purchasable on a shipping line threw a fatal error
  because Eloquent tried to instantiate ShippingOption as a model with no
  arguments, but it is a plain DTO with required constructor parameters.

I've added a getPurchasableAttribute() accessor that detects purchasable_type of
  ShippingOption and reconstructs the DTO from the snapshotted order line
  data (description, identifier, unit_price, option, meta) instead of
  querying the database. All other purchasable types fall through to the
  normal morphTo relationship unchanged.

### DIFF
--- a/packages/core/src/Models/OrderLine.php
+++ b/packages/core/src/Models/OrderLine.php
@@ -14,6 +14,8 @@ use Lunar\Base\Casts\TaxBreakdown;
 use Lunar\Base\Traits\HasMacros;
 use Lunar\Base\Traits\LogsActivity;
 use Lunar\Database\Factories\OrderLineFactory;
+use Lunar\DataTypes\ShippingOption;
+use Lunar\Models\TaxClass;
 
 /**
  * @property int $id
@@ -79,6 +81,23 @@ class OrderLine extends BaseModel implements Contracts\OrderLine
     public function order(): BelongsTo
     {
         return $this->belongsTo(Order::modelClass());
+    }
+
+    public function getPurchasableAttribute(): mixed
+    {
+        if ($this->purchasable_type === ShippingOption::class) {
+            return new ShippingOption(
+                name: (string) $this->description,
+                description: (string) $this->description,
+                identifier: (string) $this->identifier,
+                price: $this->unit_price,
+                taxClass: TaxClass::getDefault(),
+                option: $this->option,
+                meta: $this->meta ? (array) $this->meta : null,
+            );
+        }
+
+        return $this->getRelationValue('purchasable');
     }
 
     public function purchasable(): MorphTo
--- a/tests/core/Unit/Models/OrderLineTest.php
+++ b/tests/core/Unit/Models/OrderLineTest.php
@@ -159,3 +159,35 @@ test('non eloquent models can be added to an order', function () {
         ->and($orderLine->unit_price->unitDecimal)
         ->toEqual(6.5);
 });
+
+test('can access purchasable on a shipping order line', function () {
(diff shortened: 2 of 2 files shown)


packages/core/src/Models/OrderLine.php
```diff
@@ -81,6 +83,23 @@ public function order(): BelongsTo
...
 
+    public function getPurchasableAttribute(): mixed
+    {
+        if ($this->purchasable_type === ShippingOption::class) {
+            return new ShippingOption(
+                name: (string) $this->description,
+                description: (string) $this->description,
+                identifier: (string) $this->identifier,
+                price: $this->unit_price,
+                taxClass: TaxClass::getDefault(),
```
bpotmalnik, line 94: I've made it use default as `taxClass` property is not nullable and there is no tax_class_id stored on order_lines (as every detail is captured on `$orderLine->tax_breakdown`).

bpotmalnik: I was thinking about DX and how to inform dev about that this taxClass might be not accurate representation of what is in tax_breakdown, but I did not find right place for comment. 

I think also rather than returning `ShippingOption`, it could return `ShippingLineSnapshot` that could be more accurate object (as it is already not [real model class stored](https://github.com/lunarphp/lunar/blob/1.x/packages/core/src/Pipelines/Order/Creation/CreateShippingLine.php#L36-L37) in db.

glennjacobs:
Thanks @bpotmalnik — the diagnosis here was exactly right: `ShippingOption` is a non-Eloquent DTO, so the morph Eloquent tries to instantiate on `$orderLine->purchasable` can never resolve, and the accessor you built was a reasonable way to paper over that.

We've gone with the alternative approach in #2643 (now merged): rather than reconstructing the DTO on read, the purchasable morph becomes nullable and shipping lines stop carrying the fake morph entirely — matching the architecture 2.x already adopted in #2538, so code written against 1.x behaves identically after an upgrade. That also covers the eager-load path (`with('lines.purchasable')`), which an accessor can't intercept, and avoids the `TaxClass::getDefault()` substitution you'd flagged yourself on this thread.

Your test scenarios were ported into #2643. Appreciate the report and the PR — closing as superseded.

PR #2509, opened 2026-06-01
Ensure automatic discounts can be fulfilled

ryanmitchell:
the processAutomaticDiscounts method in BuyXGetY doesnt current consider whether the discount item can be fulfilled, so errors can be thrown in user land when trying to add an item to cart that has no available stock.

this PR updates the logic to ensure we only choose items that can be fulfilled.

### DIFF
--- a/packages/core/src/DiscountTypes/BuyXGetY.php
+++ b/packages/core/src/DiscountTypes/BuyXGetY.php
@@ -249,18 +249,37 @@ class BuyXGetY extends AbstractDiscountType
 
         // we have lines to add
         if ($remainingRewardQty > 0) {
-            while ($remainingRewardQty > 0) {
-                $selectedRewardItem = $this->discount->discountableRewards->random()->discountable;
+            $fulfillableRewards = $this->discount->discountableRewards->filter(function ($discountableReward) {
+                $rewardItem = $discountableReward->discountable;
 
-                if (! $selectedRewardItem) {
-                    $remainingRewardQty--;
-
-                    continue;
+                if (! $rewardItem) {
+                    return false;
                 }
 
+                if ($rewardItem instanceof LunarCollection) {
+                    return $rewardItem->products()
+                        ->with('variants')
+                        ->get()
+                        ->some(fn ($p) => $p->variants->first()?->canBeFulfilledAtQuantity(1));
+                }
+
+                return (bool) $rewardItem->variants->first()?->canBeFulfilledAtQuantity(1);
+            });
+
+            if ($fulfillableRewards->isEmpty()) {
+                return [$affectedLines, $discountTotal];
+            }
+
+            while ($remainingRewardQty > 0) {
+                $selectedRewardItem = $fulfillableRewards->random()->discountable;
+
                 if ($selectedRewardItem instanceof LunarCollection) {
-                    $product = $selectedRewardItem->products()->inRandomOrder()->first();
-                    $purchasable = $product?->variants()->first();
+                    $product = $selectedRewardItem->products()
+                        ->inRandomOrder()
+                        ->with('variants')
+                        ->get()
+                        ->first(fn ($p) => $p->variants->first()?->canBeFulfilledAtQuantity(1));
+                    $purchasable = $product?->variants->first();
                     $selectedRewardItem = $product;
                 } elseif ($selectedRewardItem instanceof Purchasable) {
                     $purchasable = $selectedRewardItem;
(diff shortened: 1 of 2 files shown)


glennjacobs (changes requested):
Thanks Ryan — the fix is the right idea in the right place: automatically-added reward lines bypass the `CartLineStock` validator, so checking fulfillability at selection time is exactly where this belongs, and the four test scenarios are thorough.

A few things need addressing before this can land, mostly stemming from the branch pre-dating your own #2624/#2631/#2632 merges (as those PRs predicted, a rebase is needed — GitHub already reports conflicts, so nothing can merge silently):

**1. The new filter fatals on variant-typed rewards**
Both the fulfillability filter and the selection assume a reward discountable is a `Product` or `Collection` (`$rewardItem->variants->first()`), but a `ProductVariant` reward is legitimate — #2631 just fixed this exact crash in the old code. `ProductVariant` has no `variants` relation, so `null->first()` throws, and the variant-reward test now on `1.x` would fail. The rebase alone won't cover it: the new filter needs its own `instanceof Purchasable` branch mirroring the one #2631 added to the selection.

**2. Collection hydration cost inside `Cart::calculate()`**
The filter loads `products()->with('variants')->get()` — the entire collection — per collection reward, and the selection loop re-hydrates the whole collection once per remaining reward unit (previously a single `inRandomOrder()->first()` query). This runs on essentially every cart mutation in a storefront, so for large collections it's O(collection × reward_qty) per recalculation. Could the fulfillable check be pushed into the query (or at least memoised across the filter/selection steps and loop iterations)?

**3. `->or` isn't a Pest API**
`expect($cart->freeItems)->toBeNull()->or->toBeEmpty()` doesn't do what it reads as: `or` resolves to a higher-order expectation on a (nonexistent) `or` property, so the `toBeEmpty()` trivially passes against `null` — and a failed `toBeNull()` throws before it anyway. The tests pass today because `freeItems` genuinely is null, but the fallback is illusory — best to assert `toBeNull()` alone.

**Smaller notes**
- The qty-1 guard checks unchanging in-memory stock per iteration, so `reward_qty` 3 against a stock-1 variant still allocates 3 units. Still a clear improvement over no check — flagging in case it's cheap to track cumulative allocation while you're in here, especially if #2639 (which touches the same loop) lands first.
- The two back-to-back guards with identical outcomes (`! $purchasable` / `! canBeFulfilledAtQuantity`) can collapse into one.

Given #2639 also reworks this loop, it may be easiest to land that first and rebase this on top. Happy to re-review quickly once it's updated.

ryanmitchell:
Would probably have preferred this to merge before the others as it would have been easier to merge these changes into the others :)

I dont mind refactoring now, but could do with guidance from you on merge order before I spend the time on it. If you're planning on merging 2639 first then I'd need to wait.

glennjacobs:
#2639 has been merged now.

ryanmitchell:
Thanks, I've updated this now and applied the changes you suggested. It still checks for quantity 1, but if you want that I can change it.

glennjacobs:
Thanks Ryan, that covers it — I checked the amends over:

- The `instanceof Purchasable` branch fixes the variant-reward crash. Confirmed by running a variant-typed reward cart against `235f3f6d` (still throws at the filter) and against `6fc10eb8` (adds the line correctly). The existing "Test Variant Reward Discount" case covers that path, so it can't silently regress.
- Caching the fulfillable products per collection removes the per-iteration re-hydration.
- `expect($cart->freeItems)->toBeNull()` now asserts something real, since `freeItems` stays null until a reward is pushed.
- Rebase onto #2639 is clean and the merged fixes are all intact. CI green.

One thing left, which is the cumulative-allocation note from last time — now that #2639 has landed it shows up more sharply. A variant with `purchasable = 'in_stock'` and `stock = 1`, on a discount with `reward_qty = 3`, produces a single reward line with **quantity 3**: the guard asks `canBeFulfilledAtQuantity(1)` on every iteration and never accounts for what this run has already allocated. Before #2639 that was three lines of one, which is just as over-allocated but less visible; now it's one line whose quantity is triple the stock.

Not a regression — before this PR the reward was added regardless of stock — but it's the same bug class the PR is about, and the loop already tracks what it has added, so it's a couple of lines:

```php
$allocated = $addedRewardLines[$rewardKey]->quantity ?? 0;

if (! $purchasable || ! $purchasable->canBeFulfilledAtQuantity($allocated + 1)) {
```

(`$rewardKey` would need computing just above the guard rather than just below it.) Worth a test alongside the four you have: `reward_qty` 3 against a stock-1 variant should add one unit, not three.

Happy to merge once that's in — everything else here is good to go.

ryanmitchell:
No problem, easily added as you say. Test coverage there now too.

glennjacobs (approved):

PR #2513, opened 2026-06-04
Order fulfilments, statuses & open/closed lifecycle + Notify customer (specs 0022, 0024, 0025, 0035)

glennjacobs:
Implements [`specs/0022-order-fulfilments.md`](specs/0022-order-fulfilments.md) and, on top of it, [`specs/0024-shipping-carriers.md`](specs/0024-shipping-carriers.md), [`specs/0025-order-cancellation.md`](specs/0025-order-cancellation.md), [`specs/0029-entry-point-conventions.md`](specs/0029-entry-point-conventions.md), [`specs/0030-fulfillable-order-lines.md`](specs/0030-fulfillable-order-lines.md), [`specs/0031-fulfilment-methods.md`](specs/0031-fulfilment-methods.md), [`specs/0032-auto-close-settled-orders.md`](specs/0032-auto-close-settled-orders.md) [`specs/0033-reduce-config-surface.md`](specs/0033-reduce-config-surface.md) and [`specs/0035-notify-customer-action.md`](specs/0035-notify-customer-action.md). It introduces line-item **fulfilments** and, in the process, **reshapes the order lifecycle**: the hand-driven headline `Order::$status` is gone, replaced by two derived rollups plus a Shopify-style open/closed archive.

> The design evolved during review — see the amendment banner at the top of spec 0022. Most notably, the originally-proposed *derived headline `OrderState`* (with `OnHold`/`Cancelled`/`Refunded` overrides) was **dropped entirely** in favour of the open/closed model below.

## Order lifecycle

- **No headline `OrderState`.** An order now carries only:
  - `payment_status` — derived rollup of `transactions` (Shopify `financial_status` vocabulary).
  - `fulfilment_status` — derived rollup over the order's physical lines and its fulfilments.
  - **Open / Closed** — a nullable `closed_at` archive flag (inbox-zero). `CloseOrder` / `ReopenOrder` + `OrderClosed` / `OrderReopened`.
- The `OrderState` machine, `OrderStateConfig`, `OrderStateCategory`, `computeOrderStatus`/`overrideStates`, `OrderStatusUpdated`, and the status-tied actions (`MarkOrderAsComplete`/`Shipped`, `UpdateOrderStatus`) are **removed**. `RecomputeOrderStatus` now only recomputes the two rollups.
- Order list defaults to the **Open** work queue (the old 6-month perf filter is gone).
- **Auto-close (spec 0032, opt-in).** The `OrderSettings` contract (`autoClosesSettledOrders()`) — off by default — closes an order via `CloseOrder` the moment it is fully **paid and fulfilled**, so a settled order drops out of the Open queue without a click (Shopify's "automatically archive"). A `CloseSettledOrder` listener on the two status-changed events; close-only (a later return/refund never reopens). Additive — no schema, no breaking change.

## Fulfilments (per-parcel)

- **`Fulfilment` + `FulfilmentLine`** models/contracts/factories; new baseline tables. A guarded `FulfilmentState` lifecycle: `pending → in-progress → shipped`, plus `returned`, `cancelled`.
- **Shopify-style split-down:** an order gets **one** fulfilment covering all physical lines on placement; the merchant **splits / merges** from there (no manual "create"). Inline, Shopify-like split & merge UX.
- **Reversible at each step** (each a deliberate action, not a casual transition):
  - `InProgress ↔ Pending`
  - `Shipped → Pending` — "Cancel fulfilment" un-ships (clears `shipped_at` + tracking, items back to unfulfilled, re-shippable). Lives as a danger ⋮ action, not in the status menu.
  - `Returned → Shipped` — "Undo return" restores the shipment (keeps `shipped_at` + tracking).
- **On hold** ([Shopify-style](https://shopify.dev)) — a `held_at` + `hold_reason` + `hold_note` flag, orthogonal to state; blocks shipping until released; reasons from the `HoldReasons` manifest.
- **Locations** — a `Location` model (warehouses/stores) with admin CRUD; every fulfilment belongs to a location; merge/move refuse to cross locations; fulfilments can be moved between locations.
- **Fulfillability is a stamped fact, not a type-string match (spec 0030)** — `order_lines.requires_shipping` is set from `Purchasable::isShippable()` at order creation, and the fulfilment system (initial parcel, rollup, eligibility, quantity validation) keys off `Order::fulfillableLines()` instead of `type === 'physical'`. A shippable custom purchasable (`'giftcard'`, `'bundle'`, …) is now fulfilled like any other line; covering a non-shippable line is rejected outright. `physicalLines()`/`digitalLines()` remain as display helpers and `type` is an open set again.
- **Concurrency-safe quantity invariant** — the "covered quantity never exceeds the line" check is a cross-row sum, so every coverage-changing write (create / split / merge / move) and both validation rules take `lockForUpdate()` on the rows they reason over, inside the action's transaction. Concurrent requests serialise on the order-line row instead of both validating against the same stale total.

## Tracking & carriers (spec 0024)

- **One-to-many tracking** per fulfilment (`fulfilment_trackings`); optional fulfilment `reference`; add / **remove** tracking.
- **Shipping carrier registry** — `ShippingCarrier` contract + `CarrierManifest` (+ `Carriers` facade) with container-registered carriers (ships Royal Mail / DPD / UPS / FedEx). Picking a carrier supplies the service options and **derives the tracking URL** from the number; validates the number format.

## Fulfilment methods (spec 0031)

Generalises `Fulfilment` from *a shipment* to a parcel whose flow is owned by a registered **fulfilment method** — so collection and digital goods, and any bespoke flow, stop being forced through ship-and-track.

- **A registered seam, batteries included.** `FulfilmentMethod` contract + `FulfilmentMethodManifest` (+ `FulfilmentMethods` facade), mirroring the carrier manifest: singleton, seeded with the core three, `register()` / `set()` / `forget()` / `all()` / `get()`. A method owns its **state graph**, which order **lines it claims**, and whether it carries **tracking**; `Fulfilment` stores its `method` as a key and resolves it like a carrier key.
- **Three core methods on the seam** — `shipping` (0022's behaviour refactored on, the catch-all, the only one that carries tracking), `collection` (click-and-collect: `pending → ready-for-collection → collected`, consuming the existing `ShippingOption::collect` flag stamped onto the shipping line), `digital` (manual provisioning: `pending → provisioned`). A consumer registers a `prescription` flow the same way — a class implementing `FulfilmentMethod`, registered against the manifest from a service provider (see **Extending** below).
- **Pluggable per-parcel states, fixed rollup categories.** Every `FulfilmentState` declares a `FulfilmentStateCategory` (`Outstanding` / `Fulfilled` / `Returned` / `Cancelled`) — the closed vocabulary the rollup, split/merge and return mechanics reason over, so an arbitrary flow stays correct by construction. `shipped_at` generalises to a category-driven handed-over timestamp (reads as collected-at / provisioned-at), stamped/cleared in `TransitionFulfilment`.
- **Per-method graphs over stock Spatie.** `DefaultFulfilmentStateConfig` feeds Spatie the *union* of every method's states/transitions; a single `MethodAwareTransition` guard restricts each parcel to its own method's graph, so `transitionTo` / `canTransitionTo` / `transitionableStates` (and the admin "Update status" menu) are method-correct without a per-row config.
- **`requires_fulfilment` completes the rollup denominator (extends 0030).** A new stamped `order_lines.requires_fulfilment` (from `Purchasable::requiresFulfilment()`, defaulting to `isShippable()`) re-keys `Order::fulfillableLines()` off physical-only — provisionable digital goods are now counted, and a digital-only order no longer resolves instantly to `Fulfilled`. `EnsureInitialFulfilment` walks methods by `priority()`, each claiming a subset of the still-unclaimed lines, so a mixed basket fans into **one parcel per method**.
- **Verbs & admin.** New `Fulfilment::fulfil()` (the no-tracking terminal — "Mark collected" / "Mark fulfilled"); `ship()` throws on a non-tracking method; merge/move gain a same-method guard. The fulfilments card gains a **method badge**, a **category-driven state badge**, a `fulfil` action, tracking UI gated on `usesTracking()`, and an `extendFulfilmentActions` hook — every action still routes through the model verbs (agent-native parity).

## Order cancellation (spec 0025)

- Cancel an order **that hasn't shipped**: `CancelOrder` (`CancelsOrder`) voids the un-shipped parcels, stamps `cancelled_at` / `cancel_reason` / `cancel_note`, **also closes** the order, and fires `OrderCancelled`. One-way.
- Status-only — **no refund and no restock** (deferred to the refunds / inventory specs). Cancel reasons from the `CancelReasons` manifest; `SendOrderCancelledNotifications` gated by a "notify the customer" toggle.
- "Cancelled" takes display precedence (`Order::lifecycleStatus()`) over Open/Closed.

## Entry points & public API (spec 0029)

A conventions spec landed mid-PR after the manager/facade question came up in review:

- **Three layers, one job each**: actions are the swap seam (unchanged, per spec 0016); operations whose **first parameter is the model** are exposed as verb methods on that model *and its contract*, one-line-delegating to the bound action contract (the `$cart->createOrder()` shape); managers + facades are reserved for things that genuinely need managing — session state, fluent context, driver resolution, registries.
- The consumer API is therefore model-centric: `$order->createFulfilment($lines)` / `cancel()` / `close()` / `reopen()` / `capture()` / `refund()`, and `$fulfilment->ship()` / `split()` / `merge()` / `moveLinesTo()` / `cancel()` / `markReturned()` / `transition()` / `hold()` / `release()` / `changeLocation()` / `addTracking()`, plus `$tracking->remove()`. Rebinding an action contract still takes effect through the verb.
- The `FulfilmentManager` + `Fulfilments` facade introduced earlier on this branch were a stateless router over model-first calls — exactly what the convention rules out — and were **removed before merge** (never released, so no upgrade impact). `Carriers` (a registry) stays. The Filament order actions and the fulfilments panel call the verbs; `app(Contract::class)->execute(...)` is gone from consumer-facing code.
- The convention is recorded in `CLAUDE.md` so future PRs are reviewed against it.

## Admin

- A **bespoke fulfilments panel** (custom Livewire component) on the order screen: cards per fulfilment, grouped lines with expandable per-line detail (extensible via hooks), inline split/merge, an **Update status** menu (forward-only), and a ⋮ group for Split / Merge / Change location / Add tracking / Hold / Release / Undo return / **Cancel fulfilment**. Ship modal with a carrier-aware tracking form.
- Order header actions: Capture, Refund, **Close / Reopen**, **Cancel order** (reason + staff note + notify toggle). Order table & global search show payment/fulfilment status + Open/Closed/Cancelled.
- **Location** resource (CRUD).
- **Timeline** now renders fulfilment changes meaningfully (state transitions + holds/releases) instead of a blank/generic entry.
- **Fix:** the order page no longer 500s when an address has no country (`country?->name` in the address infolist).

## Notifications

All customer notifications live in **one** `OrderNotifications` registry (a container manifest, not config) — each entry declares the states that auto-fire it (`on`), whether it's manually sendable (`manual`), and its `scope` (`Order` or `Fulfilment`, which sets the payload). `Order` is `Notifiable` out of the box, so a consumer just writes a notification class and registers it. The branded lifecycle bodies don't ship yet (drafted as spec 0036); the registry ships one general-purpose default (`OrderUpdate`, manual-only). This PR makes every path live and gated.

- **Order-level status notifications** fire when a derived status changes — `SendOrderPaymentStatusNotifications` (`paid` / `refunded` / …), `SendOrderFulfilmentStatusNotifications` (`fulfilled` / …), and `SendOrderCancelledNotifications` (`cancelled`). They resolve `OrderNotifications::triggeredBy($name, Order)` and construct the notification with the **`Order`**.
- **Per-fulfilment notifications** fire on a fulfilment state change — the *"your parcel has shipped, here's the tracking"* path (spec 0034). The new `SendFulfilmentStatusNotifications` listener consumes `FulfilmentStatusUpdated` and resolves through the method-aware `FulfilmentStateConfig::notificationsFor($state)` seam, so a custom fulfilment method maps **its own** states to notifications. The notification is constructed with the **`Fulfilment`** (so it can read tracking + lines).
- **Per-operation `notify` gate (spec 0034).** `$fulfilment->ship()` / `fulfil()` / `markReturned()` / `transition()` take a trailing `bool $notify = true`; the intent rides the events (`FulfilmentStatusUpdated` / `OrderFulfilmentStatusUpdated`), so unticking it suppresses **both** the per-fulfilment email *and* the order-rollup email that the same operation triggers (e.g. shipping the final parcel, which also flips the order to `fulfilled`). The admin status modals show a **"Notify customer"** toggle, but only when a notification is registered for the destination state — its presence is the cue that the action will email the customer.
- **Mail routing.** `Order::routeNotificationForMail()` sends to the **billing** contact by default. A notification implementing `RoutesToOrderContact` targets the **shipping** contact instead (with a fallback to the other contact when the chosen one has no email); one implementing `ResolvesOrderMailRoute` names its own recipient entirely (an account email, an ops inbox, …).
- **Interactive send (spec 0035 — now implemented, see below).** `$order->notifyCustomer(...)` restores v1's ad-hoc "compose & send": a variant registry, custom message, recipient choice, and timeline audit.

**Implement a notification.** Write a standard Laravel notification, type-hint the payload its layer hands it (`Fulfilment` for a fulfilment-state key, `Order` for a status key), and register it under the matching key:

```php
use Illuminate\Notifications\Notification;
use Illuminate\Notifications\Messages\MailMessage;
use Lunar\Core\Contracts\RoutesToOrderContact;
use Lunar\Core\Models\Contracts\Fulfilment;

class OrderShipped extends Notification implements RoutesToOrderContact
{
    public function __construct(protected Fulfilment $fulfilment) {}

    public function via(object $notifiable): array
    {
        return ['mail'];
    }

    public function orderContactType(): string
    {
        return 'shipping';   // deliver to the shipping contact, not the billing default
    }

    public function toMail(object $notifiable): MailMessage
    {
        $tracking = $this->fulfilment->trackings->first();

        return (new MailMessage)
            ->subject('Your order has shipped')
            ->line('Good news — your order is on its way.')
            ->when($tracking, fn ($mail) => $mail->action('Track your parcel', $tracking->url));
    }
}
```

```php
use Lunar\Core\Enums\NotificationScope;
use Lunar\Core\Facades\OrderNotifications;

// in a service provider — auto-fire on the 'shipped' fulfilment state (receives the Fulfilment)
OrderNotifications::register('shipped', \App\Notifications\OrderShipped::class, on: ['shipped'], scope: NotificationScope::Fulfilment);
```

Now `$fulfilment->ship($tracking)` (or the admin **Ship** action) emails the customer; `$fulfilment->ship($tracking, notify: false)` — or unticking **Notify customer** — makes the same change silently.

## Interactive customer notifications (spec 0035)

Restores v1's lost ad-hoc customer email as an **order-level** action — compose & send on demand — with the send and audit trail in **core** so an API caller logs the same way the admin does.

- **`$order->notifyCustomer($notification, $message, $recipients)`** (core `NotifyCustomer` action / `NotifiesCustomer` contract): delivers a chosen notification to each recipient, logs an `email-notification` activity per recipient, and dispatches `OrderCustomerNotified`. Recipients default to the order's billing + shipping contacts (deduped); an explicit list overrides.
- **One `OrderNotifications` registry** — a manifest + facade mirroring `CancelReasons` / `HoldReasons`, holding *all* customer notifications (manual + automatic) rather than a separate manual catalogue. Each entry carries `on` (auto-trigger states), `manual` (sendable from the admin — so an automatic notification can be **resent**), and `scope`. The action lists `OrderNotifications::sendable(Order)`. Ships a default general-purpose **`OrderUpdate`** (manual-only, order-scoped, implements `AcceptsCustomerMessage`) so it works out of the box; swap it by re-registering `order-update`.
- **Re-homed audit renderer** — the orphaned `email-notification` activity renderer has a producer again; the timeline records which variant, to whom, and the message (no rendered-HTML storage — a privacy / retention call).
- **Filament `NotifyCustomerAction`** on the order header: variant select, optional message, recipient checkboxes + an ad-hoc address. Hidden when no order-scoped notification is sendable. The ad-hoc address is required when the order has no contact to fall back on, and any domain failure surfaces as a notification rather than a 500.

```php
use Lunar\Core\Facades\OrderNotifications;

// add your own admin-sendable variants, or swap the default
OrderNotifications::register('delay-apology', \App\Notifications\DelayApology::class, 'Delay apology');
```

## Extending

Spec 0033 moved per-store *data* and class *swaps* off `config` and onto the container, so each seam below is wired from a service provider — manifest/facade calls in `boot()`, container binds in `register()`. Short examples follow; each uses a real, current API.

**Register a shipping carrier.** Extend `Carrier` (it implements the tracking-URL substitution and number validation) and register the class:

```php
use Lunar\Core\Shipping\Carriers\Carrier;

class Evri extends Carrier
{
    public function getKey(): string { return 'evri'; }
    public function getName(): string { return 'Evri'; }
    public function getServices(): array { return ['Standard', 'Next Day']; }

    protected function trackingUrlTemplate(): ?string
    {
        return 'https://www.evri.com/track/parcel/{tracking_number}';
    }
}
```
```php
use Lunar\Core\Facades\Carriers;

Carriers::register(Evri::class);   // add one …
Carriers::forget('ups', 'fedex');  // … drop ones you don't offer …
Carriers::set([Evri::class]);      // … or replace the whole set
```

**Register a fulfilment method.** A method owns its state graph, declares which lines it `claim()`s, and whether it `usesTracking()`. The lowest `priority()` claims first; `shipping` (30) is the catch-all:

```php
use Illuminate\Support\Collection;
use Lunar\Core\Contracts\FulfilmentMethod;
use Lunar\Core\Models\Order;
use Lunar\Core\States\Fulfilment\{Pending, Cancelled};

class Prescription implements FulfilmentMethod
{
    public function getKey(): string { return 'prescription'; }
    public function getLabel(): string { return __('Prescription'); }

    public function states(): array
    {
        return [Pending::class, Verified::class, Dispensed::class, Cancelled::class];
    }

    public function transitions(): array
    {
        return [
            Pending::class   => [Verified::class, Cancelled::class],
            Verified::class  => [Dispensed::class, Pending::class],
            Dispensed::class => [Pending::class],
            Cancelled::class => [],
        ];
    }

    public function defaultState(): string { return Pending::class; }
    public function fulfilledState(): string { return Dispensed::class; }

    public function claim(Order $order, Collection $unclaimed): Collection
    {
        // your predicate — claim a subset of the still-unclaimed lines
        return $unclaimed->filter(fn ($line) => $line->purchasable?->is_prescription)->values();
    }

    public function priority(): int { return 15; }
    public function usesTracking(): bool { return false; }
}
```
```php
use Lunar\Core\Facades\FulfilmentMethods;

FulfilmentMethods::register(Prescription::class);  // also: set([...]) / forget('collection')
```

A custom state only declares its `category()` — the fixed rollup vocabulary (`Outstanding` / `Fulfilled` / `Returned` / `Cancelled`) is what keeps an arbitrary flow correct (here `Verified` would be `Outstanding`):

```php
use Lunar\Core\Enums\FulfilmentStateCategory;
use Lunar\Core\States\Fulfilment\FulfilmentState;

class Dispensed extends FulfilmentState
{
    public static string $name = 'dispensed';

    public function label(): string { return __('Dispensed'); }

    public function category(): FulfilmentStateCategory
    {
        return FulfilmentStateCategory::Fulfilled;  // counts as handed-over in the rollup
    }
}
```

**Customise hold / cancel reasons** — plain `key => label` sets behind two facades:

```php
use Lunar\Core\Facades\{HoldReasons, CancelReasons};

HoldReasons::add('weather', 'Severe weather delay');  // keep defaults, add one
CancelReasons::forget('declined');                    // drop one
CancelReasons::set(['gift' => 'Unwanted gift']);      // … or replace the set
```

**Make a custom purchasable fulfillable** — stamped onto the order line at creation. `requiresFulfilment()` defaults to `isShippable()`, so override it for goods that are provisioned but not posted:

```php
public function isShippable(): bool { return false; }        // a gift card isn't posted …
public function requiresFulfilment(): bool { return true; }  // … but a human still provisions it
```

**Enable auto-close** — bind the `OrderSettings` contract (off by default); decide globally or per order / channel:

```php
use Lunar\Core\Contracts\OrderSettings;
use Lunar\Core\Models\Contracts\Order;

$this->app->bind(OrderSettings::class, new class implements OrderSettings {
    public function autoClosesSettledOrders(Order $order): bool { return true; }
});
```

**Swap an operation's behaviour** — rebind the action contract; the model verb (`$fulfilment->ship()`) routes through it automatically:

```php
use Lunar\Core\Contracts\Actions\Fulfilment\ShipsFulfilment;

$this->app->bind(ShipsFulfilment::class, \App\Actions\ShipViaCourierApi::class);
```

**Extend the admin fulfilments panel** — add per-line detail rows (or ⋮ actions, via `extendFulfilmentActions`) through the panel's hooks:

```php
use Lunar\Admin\Support\Facades\LunarPanel;
use Lunar\Admin\Filament\Resources\OrderResource\Pages\Components\OrderFulfilments;

LunarPanel::extensions([
    OrderFulfilments::class => new class {
        public function extendFulfilmentLineDetails(array $rows, $line): array
        {
            $rows[] = ['label' => 'SKU', 'value' => $line->orderLine?->identifier];

            return $rows;
        }
    },
]);
```

## Translations & upgrade

- All 16 locales updated (core `states` / `exceptions` / new `fulfilment` method labels; admin `order` / `components`; filament `actions` / `order`).
- `upgrade` Rector (`LunarSetList`): removed the now-targetless order-status renames; action-contract map trimmed for the removed actions. A new guarded, one-way migration backfills `order_lines.requires_fulfilment` from v1's `type = 'physical'`.

## Breaking changes (vs the original 0022 proposal / v1)

- **Headline `Order::$status` removed** — `orders.status` column dropped; orders expose `payment_status`, `fulfilment_status`, `closed_at`, `cancelled_at` instead. `OrderState*`, `OrderStateConfig`, the status actions and `OrderStatusUpdated` are gone.
- Filament order actions Cancel/Hold/Resume/MarkComplete/MarkShipped/UpdateStatus (and bulks) removed; replaced by Close/Reopen/Cancel + the per-parcel actions.
- **Spec 0031 contract additions** — `Contracts\Purchasable` gains `requiresFulfilment(): bool` (additive, defaults to `isShippable()` on the shipped base) and `States\Fulfilment\FulfilmentState` gains abstract `category()`; consumers implementing either from scratch must add them. `Order::fulfillableLines()` re-keys to the new `requires_fulfilment` column (a superset — physical orders unaffected). `DefaultFulfilmentStateConfig` now derives its catalogue from the `FulfilmentMethodManifest`.

## Tests & checks

- New tests across core (resolvers, both state machines, fulfilment operations + eligibility, **fulfilment methods — manifest, per-method graphs, line assignment, `fulfil`/`ship` verbs, rollup**, holds, transitions, tracking/carriers, cancellation, auto-close of settled orders, activity logging) and admin/filament (fulfilments panel incl. method-aware status routing + the consumer-action hook, location resource, cancel/close/reopen actions).
- ✅ core / admin / filament / search / stripe / shipping / upgrade suites green · Pint clean · PHPStan (level 0) clean.

## Not in this PR

- `specs/0023-demo-data-package.md` and `specs/0036-default-customer-notifications.md` are **drafted only** (no implementation). `specs/0037-notification-manifests.md` is **superseded** — its config→manifest migration is absorbed into the single `OrderNotifications` registry this PR ships (the payload distinction survives as the `NotificationScope` on each entry).
- The full branded lifecycle notification set (order confirmation, shipped-with-tracking, refunded, …), refund-on-cancel, and inventory/restock are deferred to future specs (spec 0036 / others; noted in `TODO.md`). This PR ships only the one general-purpose `OrderUpdate` default that the interactive action needs.


packages/core/config/fulfilment.php
```diff
@@ -0,0 +1,61 @@
...
+    | extra data-shaped methods here (key, label, states, transitions,
+    | default/fulfilled state, priority, tracking) and they are registered via
+    | the GenericFulfilmentMethod. A method that needs real behaviour — line
+    | claiming, custom logic — implements Lunar\Core\Contracts\FulfilmentMethod
+    | and registers with the FulfilmentMethodManifest from a service provider
+    | instead (container-for-behaviour, config-for-data).
+    |
+    | Example:
+    |
+    | 'slot-booking' => [
```
wychoong: suggest that lunar to move away from heavy config files and multiple ways to setup
- only register via *Manifest 
- use *State class for states (same as OrderState)
- would prefer to just instruct define a /App/Lunar/FulfiilmentMethods/InHouse.php, then register via FulfilmentMethodManifest in service provider


TL;DR: drop GenericFulfilmentMethod

packages/core/src/States/Order/Fulfilment/FulfilmentStatus.php
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Derived order-level fulfilment status — a rollup over the order's
+ * fulfillable physical lines and the quantities its `Fulfilment` records
+ * cover (see `Actions\Orders\ResolveFulfilmentStatus`).
+ *
+ * Distinct from the per-parcel `States\Fulfilment\FulfilmentState`: this is
+ * derived and unguarded (assigned directly by the resolver), where the
+ * lifecycle machine is hand-driven and guarded.
+ */
+abstract class FulfilmentStatus extends State
```
wychoong, line 19: standardized on *Status or *State?

packages/core/config/fulfilment.php
glennjacobs: Yes, as part of v2 I want to move away from reliance on configs. It doesn't support multi-tenancy or multi-store projects well.

packages/core/src/States/Fulfilment/DefaultFulfilmentStateConfig.php
```diff
@@ -41,9 +44,8 @@ public function defaultFulfilmentState(): string
 
     public function notificationsFor(FulfilmentState $state): array
     {
-        /** @var array<class-string> $notifications */
-        $notifications = (array) config('lunar.orders.notifications.'.$state::$name, []);
-
-        return $notifications;
+        // Fulfilment-scoped sends, looked up by the state name; constructed with
+        // the Fulfilment by SendFulfilmentStatusNotifications.
+        return $this->notifications->triggeredBy($state::$name, NotificationScope::Fulfilment);
```
wychoong, line 49: Where the $name from?

packages/core/src/Drivers/FulfilmentMethods/Collection.php
```diff
@@ -0,0 +1,106 @@
...
+        ];
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function transitions(): array
+    {
+        return [
+            Pending::class => [ReadyForCollection::class, Collected::class, Cancelled::class],
```
wychoong, line 54: I understand why the transitions are defined in this structure. But having it in a different structure than the spatie structure kind of introduces a different mental model

packages/core/src/States/Fulfilment/DefaultFulfilmentStateConfig.php
glennjacobs: Have a look at packages/core/src/States/Fulfilment/FulfilmentState.php

packages/core/src/Drivers/FulfilmentMethods/Collection.php
glennjacobs: Agreed it reads as a second model, but it's forced: spatie's config() is static per class hierarchy, so allowTransition() can't vary the graph per method. Transitions have to be queryable data (manifest unions them, MethodAwareTransition enforces the per-method subset), and an adjacency map is the natural shape for that. 

A fluent builder would just add a contract type wrapping the same map, not worth it. I've documented the rationale on the transitions() contract instead.

PR #2515, opened 2026-06-13
Fix/table rate shipping by weight

### LINKED ISSUE #2514: Weight-based shipping rate tiers
# Bug: Weight-based shipping rate tiers — remaining issues after PR #2382

## Summary

PR #2382 partially fixed weight-based shipping by normalizing product units to kg via the Cartalyst Converter. However, several issues remain that make weight-based shipping unreliable in production.

### 1. `weight_unit` hardcoded to `'kg'` — not read from shipping method data

`ShipBy.php` hardcodes `'weight.kg'` in the Converter call:

```php
$tier = $cart->lines->sum(
    fn ($line) => $line->purchasable->weight->to('weight.kg')->convert()->getValue() * $line->quantity
);
```

The Shipping Method form has a `weight_unit` select field (`kg`, `g`, `lbs`) that is saved to the `data` JSON column, but ShipBy **never reads it**. If a merchant selects grams as the unit, tiers would be entered in grams (e.g., min 500g) but the cart weight is always converted to kg (e.g., 0.5), making `0.5 >= 500` always false.

Similarly, `ManageShippingRates.php` hardcodes `'kg'` in the field suffix and helper text:

```php
->suffix(fn (Get $get) => static::isWeightCharge($get) ? 'kg' : null)
```

The selected `weight_unit` is never used to display the correct unit or to convert weights accordingly.

### 2. Float truncation on save — decimals are lost

`ManageShippingRates.php` saves weight tiers with:

```php
$price['min_quantity'] = (int) $price['min_quantity'];
```

Since `min_quantity` is an integer column, an admin entering `2.5` kg gets `(int) 2.5 = 2` stored. The 0.5 kg is silently lost. Tier boundaries become inaccurate.

### 3. Float `$tier` value causes truncation and precision bugs

`PricingManager::qty()` declares `public function qty(int $qty)`. At runtime, `$tier` for weight mode is a float (e.g., `2.5`). PHP truncates this to `2` when passed to `qty()`. While ShipBy applies its own filter with the original `$tier`, the truncated value in PricingManager can affect base price selection.

Additionally, IEEE 754 float arithmetic means `0.1 + 0.2 != 0.3`, which can cause intermittent comparison failures in `$tier >= $price->min_quantity`.

## Suggested Fix

Restore the ×100 storage scale (consistent with `cart_total` mode), make `weight_unit` configurable and used throughout, and use `(int) round()` to eliminate float precision issues:

- **Store**: Admin enters `2.5` → stored as `250` (×100, integer, no truncation)
- **Runtime**: `(int) round(2.5 * 100) = 250` (integer comparison, no float bugs)
- **Display**: `÷100` on hydrate → shows `2.5` again
- **Unit**: Read `$data['weight_unit'] ?? 'kg'` and convert all weights to that unit
- **Admin**: Dynamic label/suffix/helper showing the configured unit

zalito12:
# Fix: Weight-based shipping rate tiers with ×100 scale and configurable weight_unit

Fix #2514 

## Problem

Weight-based shipping in `lunarphp/table-rate-shipping` is broken due to three issues:

1. **Scale mismatch**: `min_quantity` for weight tiers is stored as a raw integer (e.g. `5` for 5 kg) but `PricingManager::qty(int $qty)` truncates floats like `2.5` to `2`, causing tier matching failures and precision bugs.
2. **Hardcoded unit**: The upstream fix (#2382) hardcoded `'weight.kg'` in the Converter call, ignoring the `weight_unit` attribute on `ProductVariant`. Products stored in grams or pounds would be incorrectly converted.
3. **No unit awareness in admin**: The admin panel hardcodes "kg" in the suffix and helper text regardless of the configured weight unit.

### Scale: ×100 — consistent with `cart_total` mode

Weight thresholds now use the same ×100 pattern as currency amounts in `cart_total` mode:

- **Admin enters**: `2.5` kg → **stored** as `250` in `min_quantity`
- **Runtime computes**: `(int) round(2.5 * 100)` = `250` → **compares** `250 >= 250` ✓

This eliminates `PricingManager::qty(int)` truncation and float precision issues entirely, since all values are integers.

### Configurable `weight_unit`

`ShipBy.php` reads `weight_unit` from the shipping method's `data` JSON column (defaulting to `'kg'`), and uses it to convert all product weights into the same unit before summing:

```php
$weightUnit = $data['weight_unit'] ?? 'kg';
$tier = (int) round($cart->lines->sum(
    fn ($line) => $line->purchasable->weight->to("weight.{$weightUnit}")->convert()->getValue() * $line->quantity
) * 100);
```

The admin displays the selected weight_unit dynamically via a new getShippingWeightUnit() helper in label, suffix, and helper text.

Testing
- All existing cart_total tests remain unchanged (no regression)
- Existing weight tests updated to use ×100 min_quantity values
- New test suite covers: configurable weight_unit, default fallback, data persistence, multi-unit conversion with ×100 scale


glennjacobs:
Thank you @zalito12 — the two problems you identified here are real and well-diagnosed: `ShipBy` ignoring the shipping method's `weight_unit`, and integer weight tiers making sub-kilogram thresholds impossible to express (with the admin silently truncating decimals).

We've fixed both in #2645, but via a different mechanism. The ×100 scale this PR restores is the convention #2382 deliberately migrated away from a few weeks before this was opened (see alecritson's review there) — reapplying it without a companion migration would misprice every weight tier stored since, so we couldn't take that route. Instead, #2645 keeps raw integer storage and evaluates tiers in the method's configured `weight_unit` (the column #2435 added): a merchant needing 2.5 kg precision configures the method in grams and stores `2500`, existing kg data stays valid, and the admin now shows the configured unit everywhere and rejects decimals with a visible error rather than truncating.

Really appreciate the report and the PR — the diagnosis is what got this fixed. Closing in favour of #2645.

PR #2521, opened 2026-06-26
Inventory fundamentals (spec 0038)

glennjacobs:
## Inventory fundamentals (spec 0038)

Per-location stock for Lunar v2 — replacing the flat `stock` column with a real quantity model, an audit ledger, and the movement seams that ship / return / cancel / restock all flow through. Implements `specs/0038-inventory-fundamentals.md`.

Feature-complete (all slices in the spec's implementation plan are done, including the v1→v2 upgrade backfill).

### What's in it

- **`StockLevel`** — per-location balance (`on_hand`, `incoming`, `committed`, `unavailable`, computed `available`).
- **`StockMovement`** — append-only ledger behind `on_hand` (the only ledgered bucket).
- **`StockReservation`** — a time-boxable checkout hold; `lunar:stock:release-expired` sweeps lapsed holds (scheduled by default).
- **Rollup on `ProductVariant`** — cached `stock_*` columns; indexed `stock_available` is the live sellable figure, so `getTotalInventory()` now respects committed/reserved/held-back stock.
- **`committed` and `reserved` are global-first** and **recomputed from the order book** (idempotent), so order-cancel-vs-fulfilment-cancel, partial shipments and un-ship/undo-return all fall out correctly. Physical movements (ship −, return +, refund-restock +) flow through one `RecordStockMovement` seam.
- **`TracksStock` capability** — the polymorphic seam the cart/checkout/lifecycle call, so a custom `Purchasable` can participate; `ProductVariant` is the only built-in implementation.
- **Admin** — minimal, swappable per-variant inventory controls (no standalone screens, since add-ons own opinionated inventory systems): read-only rollup summary headed by the default location, `AdjustStock` in the Stock card, selling-policy fields, recent movements. `LunarPanel::withoutInventoryControls()` disables them for an add-on to take over.
- `lunar:stock:reconcile` rebuilds the rollup from the ledger + open orders as a backstop.
- **v1 → v2 upgrade** — a guarded, one-way data migration backfills a `StockLevel` + `OpeningBalance` movement at the default location from each v1 variant's `stock`, seeds the rollup, then drops the column.

### Breaking changes

- **`product_variants.stock` column dropped.** The baseline migration was edited in place (v2 convention), so **existing dev databases need `php artisan migrate:fresh`** after pulling. CI / fresh installs are unaffected.
- `Contracts\Actions\Products\AdjustsStock::execute()` gains a nullable-last `?Location` param (defaults to `Location::getDefault()`) — existing 3-arg call sites stay valid, **no Rector needed**.
- Filament bridge: `ProductVariantForm::getStockComponent()` removed; the variant inventory `helper_text` lang keys renamed to `tooltip` and the dead `stock` group dropped.

### Testing

All suites pass in isolation: core 839, admin 226, filament 40, shipping 98, stripe 40, upgrade 43 · Pint, PHPStan (level 0) and the architecture test clean. (Running the whole `vendor/bin/pest` at once shows a pre-existing cross-suite Testbench migration artifact for the stripe suite — unrelated to this work; each suite is green on its own.)

### Follow-ons (tracked in `TODO.md`, separate specs)

- Location-scoped availability & routing; checkout reservation wiring.
- Selling-policy rework (sell-against-incoming, the `purchasable` → `selling_policy` column rename, deny-oversell mode).
- Low-stock thresholds; returned-goods quarantine.

> Note for reviewers: the upgrade ledger's `v2_baseline` list is missing the spec 0022 fulfilment-table migrations (`000059`–`000063`) — a pre-existing gap, out of scope here.


packages/core/src/Models/StockLevel.php
```diff
@@ -0,0 +1,77 @@
...
+     */
+    public function movements(): HasMany
+    {
+        return $this->hasMany(StockMovement::modelClass(), 'product_variant_id', 'product_variant_id')
+            ->where('location_id', $this->location_id);
+    }
+
+    public function getAvailableAttribute(): int
+    {
+        return $this->on_hand - $this->committed - $this->unavailable;
```
alecritson, line 79: Should we be casting these values if we're doing this.

glennjacobs: Good point, worth being explicit. Will sort.

PR #2529, opened 2026-06-29
Add Laravel 13 support

wychoong:
## What changed

**Dependency constraints** (`composer.json`, `packages/core/composer.json`)
- Added `^13.0` to `laravel/framework` version constraint
- Added `^7.0.0` to `kalnoy/nestedset` — v7 is the release that explicitly targets Laravel 13; v6 is retained for L11/L12 installs
- Removed `doctrine/dbal` as a direct dependency — Lunar never used it directly; the sole usage was a bridge API removed in Laravel 11
- Added `^11.0` to `orchestra/testbench` in dev dependencies

**`MigrateGetCandy` command** (`packages/core/src/Console/Commands/MigrateGetCandy.php`)
- Replaced `DB::connection()->getDoctrineSchemaManager()->listTableNames()` with `Schema::getTableListing()`, which is the Laravel 11+ native equivalent. The old call relied on the Doctrine DBAL bridge that was dropped in Laravel 11, meaning it was already broken on all currently supported versions.

**CI** (`.github/workflows/tests.yml`)
- Added `13.*` to the Laravel test matrix with `testbench: 11.*`
- Excluded PHP 8.2 + Laravel 13 (consistent with existing PHP 8.2 exclusions)

## Remaining blocker

**`cartalyst/converter` has no Laravel 13-compatible release.** The package versions lock-step with illuminate/* — v9 supports L11, v10 supports L12, and no v11 exists yet for L13. Until upstream releases a v11, Lunar cannot fully resolve on Laravel 13. This needs to be tracked and the constraint updated (`^9.0|^10|^11`) once the release lands.


glennjacobs:
@wychoong changed to draft whilst still WIP.

In v2 I've removed `cartalyst/converter`, possible we could port that I guess.

glennjacobs:
See https://github.com/lunarphp/lunar/blob/2.x/packages/core/src/Utils/MeasurementConverter.php

wychoong:
> @wychoong changed to draft whilst still WIP.
> 
> In v2 I've removed `cartalyst/converter`, possible we could port that I guess.

claude targeted to the wrong branch to merge

I wanted to target 1.2, max 1.4. 1.5 already support L13 and removed `cartalyst/converter`
so likely have to fork it publish with the original namespace instead of porting the package (no BC)

glennjacobs:
What is the purpose of this PR? We are on v1.3 now and soon to be 1.5, so I don't know why we're targeting 1.2

wychoong:
> why we're targeting 1.2

because only 1.2 can be targeted, 1.x is 1.5
ideally is 1.3

glennjacobs:
> > why we're targeting 1.2
> 
> because only 1.2 can be targeted, 1.x is 1.5 ideally is 1.3

Oh, following now.

We should look to use the `1.4` branch.

glennjacobs:
@wychoong are you still intending to update this PR?

wychoong:
> @wychoong are you still intending to update this PR?

the only open item should be `cartalyst/converter` and what do you suggest for this? fork under lunar? or backport `MeasurementConverter` from 1.5?

glennjacobs:
> > @wychoong are you still intending to update this PR?
> 
> the only open item should be `cartalyst/converter` and what do you suggest for this? fork under lunar? or backport `MeasurementConverter` from 1.5?

Certainly don't want to fork it, so probably backport.

wychoong:
# Upgrade guide

### Measurement converter replaces cartalyst/converter

Lunar now ships its own measurement converter and no longer depends on
`cartalyst/converter`.

**No action needed if** you only define measurements in
`config/lunar/shipping.php` and use the model dimension attributes
(`$variant->length->to('length.m')->convert()->getValue()`) — the API and
config format are identical.

**Update your code if you:**

1. Import the Cartalyst facade — replace:
```php
   use Cartalyst\Converter\Laravel\Facades\Converter;
   with:
   use Lunar\Facades\Converter;
```
2. Type-hint or instanceof-check the converter — replace
Cartalyst\Converter\Converter with Lunar\Utils\MeasurementConverter.
3. Register Cartalyst\Converter\Laravel\ConverterServiceProvider anywhere
(e.g. your test suite) — remove it; Lunar registers its converter
automatically.
4. Use Cartalyst's currency conversion or exchangers — these are not
supported by the new converter. Require cartalyst/converter in your
own composer.json if you still need them.

glennjacobs:
@wychoong I've retargetted to branch `1.4`, but it looks like we have a code style issue. Would you be able to sort?

wychoong:
@glennjacobs fixed

one thing to highlight is the test matrix. Php 8.2 is mentioned to be supported in installation guide (<1.5) but test is not covered.

 Probably old enough to ignore then we should remove 8.2 from matrix and cleanup the exclude block, or keep 8.2 and remove exclude block to cover 8.2 and Laravel 11,12

glennjacobs:
Thanks @wychoong, the work here is solid (the CI matrix, the converter backport, and the doctrine/dbal removal all check out, and the L13 cells genuinely pass). Unfortunately the release path is blocked outside Lunar: leandrocfe/filament-apex-charts has no stable Laravel 13 release for Filament v3 (3.2.1 caps at illuminate ^12; 4.x+ requires Filament v4), and with Filament now on v5 that tag is realistically never coming. CI only resolves because the monorepo allows dev stability, a real Laravel 13 app on stable stability can't install the result.

So we're keeping 1.4 (Filament v3) on Laravel 11/12 and directing Laravel 13 users to Lunar 1.5, which already supports it. Closing rather than merging, but genuinely appreciated, and the converter groundwork lives on in 1.5.

Issue #2524, opened 2026-06-27
Product that is Published throws TypeError when applying customer group scope


kevinjbecker:
- Lunar version: 1.5.0-beta.6
- Laravel Version: 13.17.0
- PHP Version: 8.5.7
- Database Driver & Version: PostgreSQL 18.4

### Expected Behaviour:

Product edit page shown.

### Actual Behaviour:

TypeError thrown. See error below:

TypeError
vendor/lunarphp/core/src/Base/Traits/HasCustomerGroups.php:97

Lunar\Models\Product::applyCustomerGroupScope(): Argument ##3 ($startsAt) must be of type DateTime, Carbon\CarbonImmutable given, called in /var/www/html/vendor/lunarphp/core/src/Base/Traits/HasCustomerGroups.php on line 151 

### Steps To Reproduce:

1. Create new Laravel project via `laravel/installer` composer package. 
2. Install Lunar pinned to 1.5.0-beta.6
3. Complete Lunar setup (publish configs, run migrations, create first Staff, seed locations data, etc.)
4. Create a new Product. 
5. Change Product status to Published.
6. Error is thrown every time the EditProduct page is shown.

PR #2533, opened 2026-06-29
fix(core): ensure generated time values are correct type.

kevinjbecker:
Fix for #2524. CustomerGroup query scope contained a type mismatch when default values are generated. See: `HasCustomerGroups::applyCustomerGroupScope`.

### DIFF
--- a/packages/core/src/Base/Traits/HasCustomerGroups.php
+++ b/packages/core/src/Base/Traits/HasCustomerGroups.php
@@ -141,11 +141,11 @@ trait HasCustomerGroups
         }
 
         if (! $startsAt) {
-            $startsAt = now();
+            $startsAt = now()->toDateTime();
         }
 
         if (! $endsAt) {
-            $endsAt = now()->addSecond();
+            $endsAt = now()->addSecond()->toDateTime();
         }
 
         return $this->applyCustomerGroupScope($query, $groupIds, $startsAt, $endsAt);


glennjacobs:
Thanks for this @kevinjbecker, and for the clear report in #2524 — the diagnosis was spot on.

We're going to land it as #2648 instead, which is a backport of the equivalent fix we already shipped on `2.x` in #2594. Rather than coercing the internal `now()` defaults with `->toDateTime()`, it widens the `DateTime` parameter hints to `DateTimeInterface`. Three reasons we'd rather go that way:

- It also fixes the caller-facing signature. `scopeCustomerGroup($group, CarbonImmutable::now())` still throws under this patch, because the hint on `scopeCustomerGroup()` itself is unchanged.
- `->toDateTime()` changes the runtime type passed down to `applyCustomerGroupScope()` from `Carbon` to native `DateTime`, so anyone overriding that method and calling a Carbon-only method on the argument would break at runtime instead of at load.
- The same narrow hints exist in `HasChannels` and `CanScheduleAvailability`, so `scopeChannel()`/`scheduleChannel()` would still fail for the same reason.

Keeping `1.x` and `2.x` identically shaped here also stops the branches drifting apart again.

You're credited on #2648, and the fix will be in the next `1.5.0` beta. Closing this in favour of that one — thanks again for tracking it down.

PR #2538, opened 2026-06-30
Optional order-line purchasables and de-morph shipping options (0045)

glennjacobs:
## Summary

Implements spec 0045. Makes the **order**-line `purchasable` morph optional and stops storing shipping options as polymorphic purchasables.

Shipping lines were a fake morph: `ShippingOption` is a non-Eloquent value object with no table, so `CreateShippingLine` hardcoded `purchasable_id => 1` to satisfy the mandatory morph. The line already snapshots everything it needs, so the morph carried no information. This change drops the lie and lets purchasable-less order lines (shipping, ad-hoc charges) stand on their own.

## Changes

**Slice 1 — nullable morph + relaxed observer**
- `create_order_lines_table`: `morphs('purchasable')` -> `nullableMorphs` (baseline edit, v2 pre-release).
- `OrderLineObserver::assertPurchasable()` early-returns on a null `purchasable_type`; a *set* morph must still implement `Purchasable`. `cart_lines` / `CartLineObserver` untouched.

**Slice 2 — de-morph shipping**
- `CreateShippingLine` stops setting the morph and keys its existing-line lookup on `type` + `identifier`.
- Admin `OrderItemsTable` product-link column already guards null purchasable (`$record->purchasable && ...`).
- Tests migrated: `CreateShippingLineTest`, `OrderLineTest` (new self-describing null-morph case), `CleanUpOrderLinesTest`.

**Slice 3 — upgrade package**
- New data migration nulls the fake shipping morph on existing `order_lines` (`type = 'shipping'`); guarded/idempotent, no `down()`.

## Breaking changes
- `$orderLine->purchasable` may now be `null` for shipping and ad-hoc lines — but it already returned a non-resolvable morph, so no correct consumer relied on it. Display/price reads route through the snapshot columns, unchanged.

## Verification
- Affected core tests + upgrade suite green. Pint and PHPStan pass.


alecritson:
How does this work with eager loading? Be good to see a test for it as this is what trips up the current version.

glennjacobs:
> How does this work with eager loading? Be good to see a test for it as this is what trips up the current version.

Works fine, as the field is now nullable and we don't put "fake" entries in. 

I've added a test also.

PR #2539, opened 2026-06-30
public_id (ULID) external addressing for addressable models (0046)

glennjacobs:
## Summary

Implements spec 0046. Adds a stable, non-sequential `public_id` (ULID) as the external address for Lunar models — generated on create, unique per table, exposed for APIs, webhooks, and syncs. The integer `id` stays the internal primary key and FK target; `public_id` is purely the outward-facing handle.

## Membership — default-on, exclusion-driven

The original draft carried `public_id` on a curated list of eight tables. That list wasn't derivable from a rule, so it read as arbitrary and had real gaps (`Fulfilment`/`Transaction` — which a fulfilment/payment webhook must address — were missing). The membership is now a **default with a short, principled exclusion list**:

> Every standalone model gets a `public_id`, **except** link/pivot models (no independent identity — addressed through the rows they join) and immutable-standard-code models (`Country`, `Currency`, `Language`, `State` — the ISO code is already a stable public id, and a better one than a ULID).

**40 tables carry it.** Two reasons drove the change: (1) the package cannot know which entities a given store exposes over its API, so guessing "not addressable" is an opinion it shouldn't hold — and a wrong guess forces a downstream migration on Lunar's own tables; (2) a mutable `handle`/`code` is not a safe external address (a rename breaks every stored reference), so `handle` and `public_id` coexist — readable key vs. rename-proof machine address.

An `ArchitectureTest` case enforces the rule: every `Base`-subclass model must use `HasPublicId` unless it's on the exclusion allowlist, so a new model can't silently miss the decision.

## Changes

- **`Models\Concerns\HasPublicId`** — mints a ULID at **construction** (`initializeHasPublicId`) so the id is present even when a row is persisted through an event-skipping path (`createQuietly`/`saveQuietly`/`withoutEvents`); still re-mints on `creating` when an explicit null is assigned, and resets to a fresh id on `replicating` (a copy is a distinct record). Exposes a `wherePublicId()` scope typed against the custom `Builder`. No route-binding override (`id` stays the route key).
- **Models** — trait + `@property` on 32 further models. `Staff` (extends `Authenticatable`, not `Base`) also gains Lunar's custom builder so its typed scope resolves.
- **Schema** — `ulid('public_id')->unique()` born on each baseline `create_*_table` migration; factories generate one (Asset has no factory).
- **Search** — `ProductOptionIndexer` now emits `public_id` (it's an included searchable model). `getScoutKey()` stays `id`.
- **Upgrade** — one migration over all 40 tables: add `public_id` nullable → batched backfill seeded from each row's `created_at` → tighten to `NOT NULL` with a unique index. Guarded/idempotent, no `down()`.

## No admin surfacing

`public_id` is deliberately **not** shown in the Filament panel. It's a machine address — customers never quote it and operators never act on it; the only reader is a developer, who reads it from the database. Surfacing it would add a translation key per resource and a column per table for no operator value. (An earlier revision added those; they were reverted.)

## Breaking changes

None for existing consumers — purely additive. New surface: `HasPublicId`, the `public_id` attribute + `wherePublicId()` scope on the included models, and `public_id` in the relevant search indexers.

## Verification

- Tests: `HasPublicIdTest`, `AddPublicIdToAddressableModelsTest`, and the `ArchitectureTest` membership rule.
- All seven CI suites pass in isolation: core 906, admin 230, filament 40, shipping 98, stripe 40, search 4, upgrade 51. Pint clean, PHPStan **No errors**.


packages/upgrade/database/migrations/2026_06_01_000008_add_public_id_to_addressable_models.php
```diff
@@ -0,0 +1,126 @@
...
+            $blueprint->ulid('public_id')->nullable()->after('id');
+        });
+    }
+
+    protected function backfill(string $table): void
+    {
+        DB::table($table)
+            ->whereNull('public_id')
+            ->orderBy('id')
+            ->each(function ($row) use ($table) {
```
wychoong, line 105: add `chunkById`

packages/upgrade/database/migrations/2026_06_01_000008_add_public_id_to_addressable_models.php
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    protected function addColumn(string $table): void
+    {
+        if (Schema::hasColumn($table, 'public_id')) {
+            return;
+        }
+
+        Schema::table($table, function (Blueprint $blueprint) {
+            $blueprint->ulid('public_id')->nullable()->after('id');
```
wychoong, line 96: maybe good idea to avoid using `->after` ?
Image
Latest LTS is MySQL 8.4 and MariaDB 12.3

glennjacobs: As it's the upgrade package, is it that much of a concern?

wychoong: store deployed with lunar 1.0 likely won't be using DB version < 2022, so maybe not a concern.

but should chunk the backfill

PR #2548, opened 2026-07-07
Search tweaks

alecritson:
Search engine tweaks

What's changed

SearchResults now serialises with camelCase keys and carries the active sort

- Dropped the SnakeCaseMapper from the SearchResults data object, so toArray() now emits perPage / totalPages instead of per_page / total_pages.
- Added sortField and sortDirection to the payload (populated via a new AbstractEngine::getSortParts() helper, [null, null] when no sort is requested), so storefronts can reflect the active sort in the UI.
- Added TypeScript literal types for links and sortDirection so the generated TS definitions are accurate.

TypeScript transformer attributes

- Search data objects are now marked with the #[TypeScript] attribute instead of the /** @typescript */ docblock annotation.
- Added spatie/typescript-transformer (^3.3) as an explicit dependency — the attribute classes live there and were previously not installed. Bumped spatie/laravel-data to ^4.23. Note: typescript-transformer v3 is a rewrite; the attribute namespace is unchanged, but host apps generating definitions via the v2-based laravel-typescript-transformer wrapper should check their transformer config.

Typesense engine hardening

- A multi-search request returns HTTP 200 even when the collection is missing — the failure is a per-search error code in the body. Any non-200 result is now logged and treated as an empty result set instead of blowing up on missing hits / facet_counts keys. ObjectNotFound is also caught alongside the existing connection exceptions.
- Facets are reset to a list (->values()) before serialisation so facets is a JSON array, matching the Meilisearch engine's output.
- Blank filter_by entries passed through by Scout are filtered out — an empty entry joined with real filters produced "unbalanced && operands" errors.
- vector_query (hybrid semantic search) is only added when the collection schema actually declares an auto-embed embedding field; requesting it without one 404s the whole multi-search.

Breaking

SearchResults::toArray() key casing changed (per_page → perPage, total_pages → totalPages) — any consumer reading the old snake_case keys must update.

Tests

New TypesenseEngineTest covering empty results, keyed-facet-to-list normalisation, camelCase serialisation with sort fields, and null sort behaviour.

### DIFF
--- a/WHATS-NEW.md
+++ b/WHATS-NEW.md
@@ -105,6 +105,18 @@ the same events, closing the programmatic-change gap. On the read side,
 a registry of named dependency graphs. This gives storefronts a precise cache
 key/invalidation vocabulary without hand-rolled busting.
 
+## Search
+
+Every engine (Database, Meilisearch, Typesense) returns a typed `SearchResults`
+data object with a consistent shape: camelCase keys (`perPage`, `totalPages`),
+the active sort echoed back as `sortField` / `sortDirection`, and facets always
+serialised as a plain array. The search data objects are annotated with
+`spatie/typescript-transformer` v3 attributes (`#[TypeScript]`), so storefronts
+can generate accurate TypeScript definitions for the whole results payload. The
+Typesense engine degrades gracefully — a missing collection yields an empty
+result set instead of an exception, and hybrid semantic search is only requested
+when the collection schema actually declares an embedding field.
+
 ## Pricing & attributes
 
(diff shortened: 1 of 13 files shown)


glennjacobs:
@alecritson As there's no spec for this, can we add an entry into https://github.com/lunarphp/lunar/blob/2.x/WHATS-NEW.md to assist with writing the docs down the line?

alecritson:
> WHATS-NEW.md

Added!

glennjacobs (approved):

PR #2567, opened 2026-07-17
fix: make record URL resolvers config-cache safe

### LINKED ISSUE #2564: [2.x] config:cache fails because lunar-filament record URL resolvers contain closures
  - Laravel version: `12.64.0`
  - Filament version: `5.6.8`
  - PHP version: `8.5`
  - Database driver: MySQL, although the error occurs before database access

  ### Expected Behaviour

  Laravel's production optimization commands should complete successfully:

  ```shell
  php artisan config:cache
  # or
  php artisan optimize
  ```

  ### Actual Behaviour

  `config:cache` fails because a closure is stored in the Laravel configuration repository:

  ```text
  LogicException

  Your configuration files could not be serialized because the value at
  "lunar-filament.record_urls.order" is non-serializable.
  ```

  The underlying exception is:

  ```text
  Call to undefined method Closure::__set_state()
  ```

  This also causes deployments that run `php artisan optimize` to fail before the application can become healthy.

  ### Steps To Reproduce

  2. Clear any existing configuration cache:

     ```shell
     php artisan config:clear
     ```

  3. Cache the configuration:

     ```shell
     php artisan config:cache
     ```

  4. Observe the serialization exception for:

     ```text
     lunar-filament.record_urls.order
     ```

  ### Root Cause

  `Lunar\Admin\LunarPanelProvider::registerBridgeRecordUrls()` writes closures into Laravel's configuration repository:

  ```php
  $this->app['config']->set(
      'lunar-filament.record_urls.order',
      fn ($record, array $context = []) => ManageOrder::getUrl([
          ...$context,
          'record' => $record,
      ]),
  );
  ```

  The same method also registers closures for:

  ```text
  lunar-filament.record_urls.product_variant
  lunar-filament.record_urls.collection_edit
  ```

  Laravel's `config:cache` command exports the final configuration using `var_export()`. Closures cannot be reconstructed from that exported
  representation, so the first resolver causes the command to fail.

  Relevant source:

  - https://github.com/lunarphp/admin/blob/9da7134fc7b8b27cc2f73e6cdb447d96caaa7fa5/src/LunarPanelProvider.php#L99-L112
  - https://github.com/lunarphp/filament/blob/1acfb414d026786b951471841891db4aa1e54b76/src/Support/RecordUrls.php#L14-L25

  ### Suggested Fix

  Use serializable callable arrays instead of closures, for example:

  ```php
  $this->app['config']->set(
      'lunar-filament.record_urls.order',
      [RecordUrlResolvers::class, 'order'],
  );
  ```

  Alternatively, the URL resolver callbacks could be stored in a dedicated runtime registry rather than Laravel's configuration repository.

  `Lunar\Filament\Support\RecordUrls` already checks `is_callable()`, so static callable arrays work without changing its public behavior.

  ### Application Workaround

  We worked around the issue by replacing the three closures after Lunar's service providers have booted:

  ```php
  app()->booted(function (): void {
      config([
          'lunar-filament.record_urls.order' => [
              LunarRecordUrls::class,
              'order',
          ],
          'lunar-filament.record_urls.product_variant' => [
              LunarRecordUrls::class,
              'productVariant',
          ],
          'lunar-filament.record_urls.collection_edit' => [
              LunarRecordUrls::class,
              'collectionEdit',
          ],
      ]);
  });
  ```

  `LunarRecordUrls` contains equivalent public static methods that call the existing Lunar Admin pages and resources.

  With this workaround:

  - `php artisan config:cache` succeeds.
  - All three values remain callable.
  - The generated order, product-variant, and collection URLs remain unchanged.

RaBic:
## Summary

- Replace the Lunar Admin record URL closures with serializable static callable arrays.
- Move the unchanged URL construction into `RecordUrlResolvers`.
- Add regression coverage for all three provider registrations.

## Why

`LunarPanelProvider` writes closures into Laravel configuration at boot. Laravel cannot reconstruct those values while caching configuration, so `php artisan config:cache` and `php artisan optimize` fail.

The callable arrays preserve all existing order, product-variant, and collection URL behaviour while allowing Laravel to export and reload configuration.

Fixes #2564.

## Verification

- `vendor/bin/pest tests/admin/Unit/LunarPanelProviderTest.php`
- `vendor/bin/pest --testsuite=admin` (236 passed, 793 assertions)
- `vendor/bin/pint --dirty --format agent`
- `vendor/bin/phpstan analyse packages/admin/src tests/admin/Unit/LunarPanelProviderTest.php --memory-limit=1G`

### DIFF
--- a/packages/admin/src/LunarPanelProvider.php
+++ b/packages/admin/src/LunarPanelProvider.php
@@ -9,11 +9,9 @@ use Illuminate\Support\ServiceProvider;
 use Lunar\Admin\Console\Commands\MakeLunarAdminCommand;
 use Lunar\Admin\Console\Commands\PublishAdminResourcesCommand;
 use Lunar\Admin\Events\CustomerUserEdited;
-use Lunar\Admin\Filament\Resources\CollectionResource;
-use Lunar\Admin\Filament\Resources\OrderResource\Pages\ManageOrder;
-use Lunar\Admin\Filament\Resources\ProductVariantResource;
 use Lunar\Admin\Listeners\FilamentUpgradedListener;
 use Lunar\Admin\Support\ActivityLog\Manifest as ActivityLogManifest;
+use Lunar\Admin\Support\RecordUrlResolvers;
 
 class LunarPanelProvider extends ServiceProvider
 {
@@ -98,17 +96,9 @@ class LunarPanelProvider extends ServiceProvider
      */
     protected function registerBridgeRecordUrls(): void
     {
-        $this->app['config']->set('lunar.filament.record_urls.order',
-            fn ($record, array $context = []) => ManageOrder::getUrl([...$context, 'record' => $record]),
-        );
-
-        $this->app['config']->set('lunar.filament.record_urls.product_variant',
-            fn ($record, array $context = []) => ProductVariantResource::getUrl('edit', [...$context, 'record' => $record]),
-        );
-
-        $this->app['config']->set('lunar.filament.record_urls.collection_edit',
-            fn ($record, array $context = []) => CollectionResource::getUrl('edit', [...$context, 'record' => $record]),
-        );
+        $this->app['config']->set('lunar.filament.record_urls.order', [RecordUrlResolvers::class, 'order']);
+        $this->app['config']->set('lunar.filament.record_urls.product_variant', [RecordUrlResolvers::class, 'productVariant']);
+        $this->app['config']->set('lunar.filament.record_urls.collection_edit', [RecordUrlResolvers::class, 'collectionEdit']);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


glennjacobs:
Thanks so much for digging into this and for the clean PR, @RaBic, you diagnosed the config:cache failure well before anyone else picked it up, and your fix is correct.

We're going to go with #2622 instead, only because it pushes the fix down into the RecordUrls contract itself: with class-string support in the bridge, third-party panels overriding lunar.filament.record_urls.* can also cache their config, rather than just fixing Lunar's shipped defaults. Your approach here directly informed that direction.

Closing in favour of #2622, really appreciate the contribution, and hope to see more from you!

PR #2593, opened 2026-08-04
Fix hub product variants mapping unmatched permutations to first SKU

wychoong:
## Summary
- Fixes a bug in `MapVariantsToProductOptions` where unmatched cartesian option permutations were incorrectly bound to the first product variant.
- Cause: `collect()->search()` returns `false` on no match, and PHP coerces `$variants[false]` to `$variants[0]`.
- With `fillMissing: false` (used by the Product Options / Variants hub widget), unmatched permutations are now skipped instead of inventing mislabelled rows.
- Adds regression coverage for sparse option matrices and for `fillMissing: true` leaving unmatched rows without an existing SKU.

### Reproduction
On a product with shared options and a non-cartesian variant matrix (for example Size/Colour/Length combinations that do not all exist), `/hub/products/{id}/variants` could show the first SKU under the wrong option labels.

## Test plan
- [x] Added Pest unit coverage in `MapVariantsToProductOptionsTest`
- [ ] Run `./vendor/bin/pest tests/admin/Unit/Actions/Products/MapVariantsToProductOptionsTest.php`
- [ ] Confirm `/hub/products/{id}/variants` Option column matches each variant’s pivot values for a sparse shared-option product

### DIFF
--- a/packages/admin/src/Actions/Products/MapVariantsToProductOptions.php
+++ b/packages/admin/src/Actions/Products/MapVariantsToProductOptions.php
@@ -28,7 +28,7 @@ class MapVariantsToProductOptions
                 $valueDifference = array_diff_assoc($permutation, $variant['values']);
 
                 if (! count($valueDifference)) {
-                    return $variant;
+                    return true;
                 }
 
                 $amountMatched = count($permutation) - count($valueDifference);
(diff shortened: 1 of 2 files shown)


wychoong:
Superseded by https://github.com/lunarphp/lunar/pull/2609, which includes this mapping fix plus the follow-up for new permutations missing `tax_class_id`.

PR #2595, opened 2026-08-04
fix(search): harden Typesense engine and decouple database engine from admin

alecritson:
Fixes several issues in the Typesense search engine and makes the database engine work standalone, with test coverage for both.

## TypesenseEngine

- **Per-element highlights on `string[]` fields** - Typesense highlights array fields as a `snippets` list with index-aligned nested `matched_tokens`, not a single `snippet`. The engine now emits one `SearchHitHighlight` per snippet and falls back to the existing scalar shape for other fields, guarding both `matched_tokens` and `snippet` so a missing key no longer throws.
- **Empty-query (browse) requests** - with no search term, `q` now falls back to `*` (Typesense's match-all), and the `embedding` field is stripped from `query_by` by index - along with its position-aligned entries in `query_by_weights` and `infix` - instead of a fragile `str_replace`, which broke when the field wasn't first or the weight/infix counts no longer matched.
- **`query_by_weights` / `infix` passthrough** - these options were previously dropped; they're now forwarded when set.
- **Typo fix**: `exlude_fields` → `exclude_fields`, so the embedding vector is actually excluded from hit payloads.
- **Configurable facet limit** - `max_facet_values` now reads the new `lunar.search.max_facet_values` config key (default 50, was hardcoded 50; Typesense caps at 250) - hierarchical category facets need more.
- **Active facet state** - facet values now include an `active` flag reflecting the currently applied facet filters.

## DatabaseEngine

- Uses Scout's `Model::search()` builder directly instead of the admin package's `get_search_builder()` helper, so the search package works without `lunar/admin` installed. Also respects `perPage` instead of the paginator default.
- Applies the indexer's `makeAllSearchableUsing()` eagerallback, so mapping hits no longer lazy-loads relationsper row (crashed under `Model::preventLazyLoading()`, N+1 otherwise).
- Passes the paginator's `links` to `SearchResults` - it's a required constructor parameter, so every `get()` previously threw `CannotCreateData`.

## ProductIndexer

- `toSearchableArray()` is now null-safe on `productType` and `created_at`. Scout's database engine calls it on an empty model prototype to derive the searchable columns, which crashed on `$model->productType->name`.

## Tests

- `TypesenseEngineTest`: highlight mapping (array-field,cet `active` flag, browse-mode match-all + embedding strip with aligned weights/infix lists, option passthrough, `exclude_fields` regression, `max_facet_values` config, vector-query gating, and
`stripListEntry` edge cases.
- New `DatabaseEngineTest`: results resolve via Scout without the admin helper; `perPage` respected.
- Search suite `TestCase` now registers the nestedset, boviders


glennjacobs:
@alecritson does the Meilisearch driver need updating with any of these new config entries etc?

alecritson:
> @alecritson does the Meilisearch driver need updating with any of these new config entries etc?

Apparently not:

- Active facet flag - already there
- Empty-query browse - Meilisearch natively treats an empty query as match-all; no * sentinel needed, and no hybrid embedding field is wired into that engine, so nothing to strip.
- links / perPage / sort parts - all already passed to SearchResults.
- max_facet_values - doesn't map. In Meilisearch the facet value limit (faceting.maxValuesPerFacet, default 100) is an index setting, not a per-search parameter, so the new config key can't be applied at query time. Consumers set it on the index.

glennjacobs:
@alecritson if we have a config, but it's not used by Meilisearch, is that not confusing? Surely Meilisearch should also honour the setting?

alecritson:
> @alecritson if we have a config, but it's not used by Meilisearch, is that not confusing? Surely Meilisearch should also honour the setting?

Meilisearch doesn't support it and trying to get Meilisearch to use the config option during index updating was a bit of a hack and we'd always be fighting against other code.

Instead opted for a  builder approach instead so you can do `->maxFacetValues(200)` when using Typesense or in the future any engine that supports it.

glennjacobs (approved):

PR #2598, opened 2026-08-08
Stop dropping media custom properties on create/edit

### LINKED ISSUE #2596: [2.x] Media custom properties are silently dropped because MediaRelationManager hardcodes `name`/`primary` on create and overwrites the whole JSON on edit
- Laravel version: `13.24.0`
- Filament version: `5.7.5`
- Livewire version: `4.3.5`
- Spatie Media Library version: `11.23.3`
- PHP version: `8.5.9`
- Database driver: MySQL

### Expected Behaviour

Spatie's `custom_properties` is an open key/value bag. Two things should follow from that in the admin media manager:

1. A field added to the media form through the documented extension seam — `RelationManagerExtension::extendForm()`, registered via `LunarPanel::extensions()` — should persist when the media item is created, exactly as fields added to any other extendable Lunar form do.
2. Editing a media item should not remove custom properties that the form does not render.

### Actual Behaviour

Neither holds. Both failures are silent: no exception, no validation error, no notification.

**1. Custom properties added by an extension are discarded on create.** The field renders, validates and submits, and the media item is created — but the value is never written:

```php
$brand->getFirstMedia('images')->getCustomProperty('credits'); // null
```

**2. Editing a media item overwrites the entire `custom_properties` JSON.** Any key without a corresponding form field is dropped on save:

```php
// before editing the item in the admin panel
$media->getCustomProperty('sha1'); // 'abc123'

// after opening the edit modal, changing the name and saving
$media->getCustomProperty('sha1'); // null
```

Scope of the second defect: a vanilla installation loses nothing, because Lunar only ever writes `name` and `primary`, and both have form fields. It affects installations that store anything else on a media item — imported metadata, dedup markers, sort keys, alt text, captions. `generated_conversions` and `responsive_images` are separate columns, so conversions and responsive images are not affected.

The same code is present on `1.x`.

### Steps To Reproduce

**Defect 1 — create drops the property**

2. Add a field to the media form through the documented extension seam:

    ```php
    class MediaCreditsExtension extends RelationManagerExtension
    {
        public function extendForm(Schema $schema): Schema
        {
            $schema->components([
                ...$schema->getComponents(withHidden: true),
                TextInput::make('custom_properties.credits'),
            ]);

            return $schema;
        }
    }
    ```

    ```php
    LunarPanel::extensions([
        MediaRelationManager::class => [MediaCreditsExtension::class],
    ]);
    ```

3. Open *Manage Brand Media* (or product/collection media), upload an image and fill in the new field.
4. Save. The item is created without an error.
5. Observe that the property was never written:

    ```php
    Brand::first()->getFirstMedia('images')->getCustomProperty('credits'); // null
    ```

**Defect 2 — edit wipes existing properties**

1. Attach a media item carrying a custom property that the form does not render:

    ```php
    $brand->addMedia($path)
        ->withCustomProperties(['sha1' => 'abc123', 'primary' => true])
        ->toMediaCollection('images');
    ```

2. Open *Manage Brand Media*, edit that item, change the name and save.
3. Observe that `sha1` is gone:

    ```php
    $brand->getFirstMedia('images')->getCustomProperty('sha1'); // null
    ```

### Root Cause

Both defects come from the same assumption: `MediaRelationManager` treats `custom_properties` as a closed set of exactly two keys.

**Create.** The `CreateAction` handler enumerates the two keys it knows about and discards the rest of the submitted array:

```php
->withCustomProperties([
    'name' => $data['custom_properties']['name'],
    'primary' => $data['custom_properties']['primary'],
])
```

Because the handler is defined inside `getDefaultTable()`, this happens *after* `extendTable()`/`extendForm()` have run, so an extension has no supported way to get its own key persisted — the extension API can add the field, but not save it.

**Edit.** `EditAction` is used unmodified, so Filament writes the submitted form data straight onto the model. `custom_properties` is an `array` cast on Spatie's `Media` model, so assigning the form's two keys replaces the whole JSON value rather than merging into it. `name` and `primary` survive only because they happen to be the two keys the form renders; everything else is lost.

Relevant source:

- https://github.com/lunarphp/admin/blob/9da7134fc7b8b27cc2f73e6cdb447d96caaa7fa5/src/Support/RelationManagers/MediaRelationManager.php#L92-L110
- https://github.com/lunarphp/core/blob/c0c6a81f4c9cc429fb99a8b9c4030828e1f29c18/src/Observers/MediaObserver.php

Note that `MediaObserver` maintains the `primary` flag with `setCustomProperty()`, which merges correctly — the observer is not affected, but it does mean the two write paths in the codebase disagree about whether `custom_properties` is open or closed.

### Suggested Fix

Both are small and independent.

1. **Create** — pass the submitted array through instead of enumerating keys:

    ```php
    ->withCustomProperties($data['custom_properties'] ?? [])
    ```

    This is also more robust than the current code, which assumes both keys are present and would raise an undefined-key warning if an extension hid either field.

2. **Edit** — merge the stored properties underneath the submitted ones, so unrendered keys survive while submitted keys still win:

    ```php
    EditAction::make()
        ->mutateDataUsing(function (array $data, Media $record): array {
            $data['custom_properties'] = array_merge(
                $record->custom_properties,
                $data['custom_properties'] ?? [],
            );

            return $data;
        }),
    ```

Say the word and I'll provide a PR for either or both.

### Application Workaround

We worked around both defects in a single `RelationManagerExtension` registered against `MediaRelationManager`, which reconfigures the two cached actions in `extendTable()`:

```php
public function extendTable(Table $table): Table
{
    $table->getAction('edit')?->mutateDataUsing(function (array $data, Media $record): array {
        $data['custom_properties'] = array_merge(
            $record->custom_properties,
            $data['custom_properties'] ?? [],
        );

        return $data;
    });

    $table->getAction('create')?->using(
        fn (array $data, MediaRelationManager $livewire): Model => $livewire->getOwnerRecord()
            ->addMediaFromString($data['media']->get())
            ->usingFileName($data['media']->getClientOriginalName())
            ->withCustomProperties($data['custom_properties'] ?? [])
            ->preservingOriginal()
            ->toMediaCollection($livewire->mediaCollection),
    );

    return $table;
}
```

`Table::getAction()` returns the live instances, because `headerActions()`/`recordActions()` cache them while `getDefaultTable()` runs, and `extendTable()` runs afterwards. Re-calling `->using()` replaces the stored closure.

With this workaround:

- Custom properties added by an extension persist on create.
- Editing a media item no longer drops keys the form does not render.
- Behaviour is unchanged for installations that only use `name` and `primary`.
- No vendor files are modified.

The `->using()` override silently shadows the packaged closure, so it has to be re-checked on every Lunar upgrade — which is the main reason we would prefer the fix upstream.

wakqasahmed:
Closes #2596

## Summary

`MediaRelationManager` treated `custom_properties` as a closed set of exactly two keys (`name`, `primary`), which silently broke Spatie Media Library's open key/value contract in two ways:

1. **Create** — the `CreateAction` handler enumerated only `name`/`primary` from the submitted form data and discarded everything else, so any field added via the documented `RelationManagerExtension::extendForm()` seam rendered, validated, and submitted — but its value was never persisted.
2. **Edit** — `EditAction` wrote the submitted form data straight onto the model. Because `custom_properties` is an `array`-cast column, this replaced the whole JSON value instead of merging into it, so any key not rendered by the edit form (imported metadata, dedup markers, alt text, etc.) was silently dropped on save.

## Fix

- **Create**: pass the full submitted `custom_properties` array through to `withCustomProperties()` instead of hand-picking `name`/`primary`.
- **Edit**: added `mutateDataUsing()` to the `EditAction` that merges the submitted `custom_properties` on top of the record's existing `custom_properties`, so unrendered keys survive while submitted keys still win.

## Testing

- Added a feature test asserting a custom property contributed via a `RelationManagerExtension::extendForm()` field is persisted when a media item is created.
- Added a feature test asserting editing a media item preserves an existing custom property key that the edit form does not render, while still applying the change to the field the form does render.
- Ran `vendor/bin/pest tests/admin/Feature/Support` (built a PHP 8.3 + intl/exif/bcmath/gd toolchain via Docker to install the monorepo's Composer dependencies): all 16 tests in that directory pass, including the 2 new tests and the pre-existing `PriceRelationManagerTest`/`RelationManagerExtensionTest`/`CreatePageExtensionTest` suites.
- Verified the new "persists custom properties added by an extension when creating media" test fails against the pre-fix code (`null` instead of `'Jane Doe'`), confirming it actually exercises bug 1.
- `php -l` on both changed files (no syntax errors).

### DIFF
--- a/packages/admin/src/Support/RelationManagers/MediaRelationManager.php
+++ b/packages/admin/src/Support/RelationManagers/MediaRelationManager.php
@@ -99,10 +99,7 @@ class MediaRelationManager extends BaseRelationManager
                             ->usingFileName(
                                 $data['media']->getClientOriginalName()
                             )
-                            ->withCustomProperties([
-                                'name' => $data['custom_properties']['name'],
-                                'primary' => $data['custom_properties']['primary'],
-                            ])
+                            ->withCustomProperties($data['custom_properties'] ?? [])
                             ->preservingOriginal()
                             ->toMediaCollection($this->mediaCollection);
                     })->after(
@@ -112,11 +109,20 @@ class MediaRelationManager extends BaseRelationManager
                     ),
             ])
             ->recordActions([
-                EditAction::make()->after(
-                    fn () => ModelMediaUpdated::dispatch(
-                        $this->getOwnerRecord()
-                    )
-                ),
+                EditAction::make()
+                    ->mutateDataUsing(function (array $data, Media $record): array {
+                        $data['custom_properties'] = array_merge(
+                            $record->custom_properties ?? [],
+                            $data['custom_properties'] ?? [],
+                        );
+
+                        return $data;
+                    })
+                    ->after(
+                        fn () => ModelMediaUpdated::dispatch(
+                            $this->getOwnerRecord()
+                        )
+                    ),
                 DeleteAction::make(),
                 Action::make('view_open')
                     ->label(__('lunarpanel::relationmanagers.medias.actions.view.label'))
--- a/tests/admin/Feature/Support/RelationManagers/MediaRelationManagerTest.php
+++ b/tests/admin/Feature/Support/RelationManagers/MediaRelationManagerTest.php
@@ -1,8 +1,15 @@
 <?php
 
+use Filament\Actions\CreateAction;
+use Filament\Actions\EditAction;
+use Filament\Forms\Components\TextInput;
+use Filament\Schemas\Schema;
+use Illuminate\Http\UploadedFile;
 use Livewire\Livewire;
 use Lunar\Admin\Filament\Resources\BrandResource\Pages\ManageBrandMedia;
 use Lunar\Admin\Filament\Resources\ProductResource\Pages\ManageProductMedia;
+use Lunar\Admin\Support\Extending\RelationManagerExtension;
+use Lunar\Admin\Support\Facades\LunarPanel;
 use Lunar\Admin\Support\RelationManagers\MediaRelationManager;
 use Lunar\Models\Brand;
 use Lunar\Models\Language;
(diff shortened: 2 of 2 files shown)


wakqasahmed:
Hi @glennjacobs — noticed this PR doesn't have a reviewer assigned yet — it's been about 10 days, CI is green and it's mergeable. Would you (or whoever's best placed) be able to take a look when you get a chance, or point me to who should? Thanks!

glennjacobs (approved):

PR #2599, opened 2026-08-08
Include non-shippable (digital) lines in order PDF invoice (#2409)

### LINKED ISSUE #2409: Non shippable lines not appearing on order invoices
If a product relating to an order line is set to not be shippable then it is not returned via the `physicalLines` relation and is not included in some exports such as the order pdf.

wakqasahmed:
Closes #2409

## Problem
The order PDF invoice view (`packages/admin/resources/views/pdf/order.blade.php`) iterates `$record->physicalLines`, which is scoped to `OrderLine::whereType('physical')`. `ProductVariant::getLineTypeAttribute()` sets an order line's type to `digital` whenever the underlying product variant is not shippable. As a result, order lines for non-shippable products (digital goods, services) are silently omitted from the generated invoice PDF.

## Fix
Use the existing `productLines` relation (`lines()->where('type', '!=', 'shipping')`) instead of `physicalLines`, so both physical and digital product lines appear on the invoice while shipping lines remain excluded (shipping is already rendered separately in the PDF).

## Test plan
- Added a Pest test to `tests/core/Unit/Models/OrderTest.php` asserting `physicalLines`, `digitalLines`, and `productLines` each return the expected line counts for a mix of physical, digital, and shipping order lines.
- Ran `vendor/bin/pest --testsuite core --filter=OrderTest` — all 16 tests pass (48 assertions), including the new test.


wakqasahmed:
Hi @glennjacobs — noticed this PR doesn't have a reviewer assigned yet — it's been about 9 days, CI is green and it's mergeable. Would you (or whoever's best placed) be able to take a look when you get a chance, or point me to who should? Thanks!

glennjacobs (approved):

PR #2600, opened 2026-08-08
fix(core): queue-safe catalog invalidation events (#2562)

### LINKED ISSUE #2562: [2.0-alpha4] Deleted catalog invalidation events fail in queued listeners
- Laravel Version: 13.19.0
- PHP Version: 8.5.7
- Database Driver & Version: 11.8.6

### What I found

I found that the catalog invalidation contract describes its captured scalar identity as authoritative for deleted targets and queued listeners. However, the model-bearing invalidation events use`SerializesModels` and expose the affected Eloquent model as a required property.

After a hard delete commits, a queued listener deserializing the event asks Laravel to restore that model with `firstOrFail()`. Because the row no longer exists, deserialization can throw `ModelNotFoundException` before the listener can read `morphType()`, `cacheKey()`, or `cacheTags()`. Retrying cannot restore a row that has been deleted.

The same required-model pattern is present in the Product, Collection, Brand, and ProductOption invalidation events.

### How to reproduce

1. Register a queued listener for a catalog invalidation event.
2. Hard-delete the corresponding catalog model and commit.
3. Process the queued listener.
4. Observe model restoration fail before the handler reads the captured scalar identity.

### Suggested resolution

The supported queued-listener surface should be scalar-first and must not require model rehydration. One backward-compatible direction is an additive scalar-only event or entity-reference DTO dispatched alongside the existing model-bearing events. Synchronous consumers that require a live model could continue using the existing events while the scalar companion becomes the documented queued contract.

This is not a safe one-line change: removing `SerializesModels` can serialize a larger, stale model graph, while removing or making `cacheModel()` nullable changes the existing public contract.

### Regression coverage

Please add real queue-serialization tests after committed hard deletes for Product, Collection, Brand, and ProductOption invalidation events. The queued handler should be able to read the captured scalar identity without restoring the deleted row.

### Pull request

I can provide a pull request implementing the agreed backward-compatible queue-safe surface and the deletion/serialization regression tests above. I would first align the implementation with the maintainers' preferred compatibility direction for model-based listeners.

Relevant source:
`packages/core/src/Contracts/CacheInvalidationEvent.php`,
`packages/core/src/Events/Catalog/`, and Laravel's
`Illuminate/Queue/SerializesAndRestoresModelIdentifiers.php`.

wakqasahmed:
Closes #2562

## Problem

`ProductInvalidated`, `CollectionInvalidated`, `BrandInvalidated`, and `ProductOptionInvalidated` use `Illuminate\Queue\SerializesModels` on their model property. That trait's `__unserialize` restores the model by re-querying the database with `firstOrFail()`. For a `Deleted` invalidation, the row is already gone by the time a queued listener processes the job, so deserialization throws `ModelNotFoundException` before the listener can read `morphType()`, `cacheKey()`, `cacheTags()`, or `reason()` — the scalar identity the contract promises is captured at record time specifically so it survives a deleted row.

## Fix

Added `Lunar\Core\Concerns\SerializesInvalidatedModel`, a queue-safe drop-in for `SerializesModels` on these four events. Instead of a `ModelIdentifier` that gets re-queried on unserialize, it captures the model's raw attributes at serialize time and rehydrates an in-memory instance via `newFromBuilder()` on unserialize — no query, so it can't fail even when the row no longer exists. `cacheModel()` still returns a real, fully-attributed `Model` instance; the public contract (`CacheInvalidationEvent`) is unchanged.

This is narrower than the additive scalar-only event / DTO the issue floated as an option: no new public surface, no change to the documented queued contract, and the model-bearing events become safe to queue as-is.

## Test plan

- Added `tests/core/Feature/CacheInvalidationTest.php` coverage that constructs each of the four invalidation events, hard-deletes the underlying row, round-trips the event through PHP's real `serialize()`/`unserialize()` (the same mechanism a queued job uses), and asserts the restored event's scalar identity (`cacheKey()`, `cacheTags()`, `morphType()`, `reason()`) and `cacheModel()` are readable without hitting the database.
- Ran the full `tests/core/Feature/CacheInvalidationTest.php` suite (24 tests, 41 assertions) — all green, including the pre-existing coverage.

```
Tests:    24 passed (41 assertions)
Duration: 260.88s
```


wakqasahmed:
Hi @glennjacobs — noticed this PR doesn't have a reviewer assigned yet — it's been about 9 days, CI is green and it's mergeable. Would you (or whoever's best placed) be able to take a look when you get a chance, or point me to who should? Thanks!

glennjacobs (changes requested):
Thanks @wakqasahmed — this is a well-scoped fix and the approach is right. I like that it keeps `cacheModel()` returning a real model rather than widening the contract with a scalar-only DTO, and the test coverage doing a real `serialize()`/`unserialize()` round-trip against a hard-deleted row is exactly what I'd want here. I've run the suite, PHPStan and Pint locally and everything's green.

Two things before merge:

1. **Move the trait to `Events/Concerns/`.** This adds a new top-level `src/Concerns/` folder, which isn't part of the core package layout — the existing convention is `Models/Concerns/` for model traits, and since this one belongs to the events, `packages/core/src/Events/Concerns/SerializesInvalidatedModel.php` (namespace `Lunar\Core\Events\Concerns`) is the right home. Worth getting right now since the namespace becomes public surface once it ships.

2. **Guard the rehydration.** `(new $value['__lunarModel'])` will instantiate whatever class name is in the payload. Queue payloads are already an object-injection surface so it's no worse than `SerializesModels`, but a one-line `is_a($class, Model::class, true)` check (falling back to the raw value) is cheap hardening, and it also covers the theoretical case of a plain array property that happens to contain a `__lunarModel` key.

One docblock suggestion while you're in there: worth noting that the restored model's attributes are a dispatch-time snapshot (not re-fetched), and that `newFromBuilder` fires the `retrieved` event on unserialize. Both are the right semantics for invalidation events, but future readers shouldn't have to discover that from the implementation.

wakqasahmed:
Thanks for the thorough review, @glennjacobs — pushed all three:

1. Moved the trait to `packages/core/src/Events/Concerns/SerializesInvalidatedModel.php` (namespace `Lunar\Core\Events\Concerns`), matching the `Models/Concerns/` convention.
2. `__unserialize()` now checks `is_a($class, Model::class, true)` before instantiating the payload's class name, closing that off as an object-injection surface.
3. Added the docblock note that the restored model is a dispatch-time snapshot (not re-fetched) and that `newFromBuilder()` fires `retrieved` on unserialize.

Ready for another pass whenever you get a chance.

PR #2602, opened 2026-08-08
Resolve staff model_type against config('lunar.staff.model')

### LINKED ISSUE #2569: [2.0-alpha4] Staff role assignment via Admin Panel stores wrong model_type, breaking permission checks
- Laravel Version: 13.20.0
- PHP Version: 8.4
- Database Driver & Version: SQLite (reproducible independent of driver)

### Expected Behaviour:
Roles assigned to a staff member via the Admin Panel's Access Control page should be stored against the same model class the panel's auth guard authenticates with, so that hasPermissionTo() / navigation gating works for that staff member after login.

### Actual Behaviour:
Roles assigned via the Staff AccessControl page are stored in model_has_roles with model_type = 'Lunar\Core\Models\Staff'. The panel guard, however, authenticates users as Lunar\Filament\Models\Staff (registered under the staff morph-map alias). Since model_type doesn't match, hasPermissionTo() returns false for every staff member who received a role through the UI — only the Dashboard and any resource with $permission = null remain visible in navigation, even though the DB and UI both appear to show the role assigned correctly.

Root cause:
packages/admin/src/Filament/Resources/StaffResource.php:22 hardcodes:
```
phpuse Lunar\Core\Models\Staff;
...
protected static ?string $model = Staff::class;
```

packages/filament/src/Schemas/Staff/StaffForm.php:116 calls syncRoles() directly on that resolved $record:
```
php->saveRelationshipsUsing(fn ($state, $record) => $record->syncRoles($state))
```

Because $record is a Lunar\Core\Models\Staff instance, Eloquent's getMorphClass() can't resolve it via the morph map — only Lunar\Filament\Models\Staff (set via packages/filament/src/LunarFilamentServiceProvider.php:22, config('lunar.staff.model')) is registered under the 'staff' alias in packages/core/src/LunarServiceProvider.php:322-324. syncRoles() therefore falls back to the raw class name Lunar\Core\Models\Staff instead of 'staff'.
The auth guard (packages/core/src/LunarServiceProvider.php:330-348) authenticates against config('lunar.staff.model'), i.e. Lunar\Filament\Models\Staff, whose morph class is 'staff' — so it never matches what got stored.

### Steps To Reproduce:

1. Install Lunar 2.x with the standard admin panel.
2. Create a new staff member via the Admin Panel.
3. Assign them a role via the Staff edit page's Access Control / Roles field, save.
4. Inspect the DB: SELECT * FROM model_has_roles ORDER BY model_id DESC LIMIT 1; → model_type is Lunar\Core\Models\Staff, not staff.
5. Log in as that staff member.
6. Navigation only shows Dashboard / permission-less resources, despite the role appearing correctly assigned in the UI.

Suggested Fix:
In packages/admin/src/Filament/Resources/StaffResource.php, resolve $model from config('lunar.staff.model', \Lunar\Core\Models\Staff::class) instead of hardcoding Lunar\Core\Models\Staff, so the resource operates on the same class the auth guard and morph map use. No changes needed in core or filament — both are already correct.

Workaround:
```
phpDB::table('model_has_roles')->where('model_type', 'Lunar\Core\Models\Staff')->update(['model_type' => 'staff']);
DB::table('model_has_permissions')->where('model_type', 'Lunar\Core\Models\Staff')->update(['model_type' => 'staff']);
Artisan::call('permission:cache-reset');
```

Only fixes existing rows — recurs on every new role assignment through the panel until fixed upstream.

wakqasahmed:
Closes #2569

## Summary

`StaffResource` hardcoded `$model` to `Lunar\Core\Models\Staff`, so role/permission assignments made via the Admin Panel's Access Control page were synced onto that raw class instead of the guard-resolved, morph-map-aliased staff model (`config('lunar.staff.model')`, aliased as `'staff'` in `LunarServiceProvider::morphMap()`).

This meant `model_has_roles.model_type` (and `model_has_permissions.model_type`) got stored as the literal `Lunar\Core\Models\Staff` string, which never matches the `'staff'` morph alias the auth guard's model resolves to. `hasPermissionTo()` therefore returned false for every staff member who received a role through the panel UI, even though the DB and UI both appeared correct.

### Fix

- `StaffResource::getModel()` now resolves `config('lunar.staff.model', Staff::class)`, matching the class the auth guard, morph map, and `LunarFilamentServiceProvider` already register/use.
- While reproducing this, found a second, previously-masked bug: `PermissionSelector::getAuthorizeRecord()` used `trait_uses_recursive($record)`, which only inspects the exact class's own traits, not inherited ones. This worked by accident only because the resource always resolved to the class that declares `Spatie\Permission\Traits\HasRoles` directly (`Lunar\Core\Models\Staff`). Once the resource correctly resolves the subclass that *inherits* `HasRoles` (`Lunar\Filament\Models\Staff`), the check needs `class_uses_recursive()` to walk the class hierarchy — otherwise every panel role/permission save started throwing `Not implemented \Spatie\Permission\Traits\HasRoles`. Fixed alongside the primary fix so the panel keeps working end-to-end.

## Test plan

- Updated `tests/admin/Feature/Filament/Resources/StaffResource/Pages/EditStaffTest.php`:
  - `can assign staff role and permissions` now re-fetches through the guard-resolved model (mirroring runtime auth) instead of the raw factory instance, so it actually exercises the real code path.
  - Added `stores role assignments against the guard-resolved staff model type`, which asserts `model_has_roles.model_type` matches `(new config('lunar.staff.model'))->getMorphClass()` directly — independent of `StaffResource` itself. Verified this test fails against the pre-fix code (`'Lunar\Core\Models\Staff'` vs expected `'staff'`) and passes with the fix.
- `vendor/bin/pest tests/admin/Feature/Filament/Resources/StaffResource --group=resource.staff` — 13 passed (55 assertions).
- `vendor/bin/pest --testsuite=admin` — 236 passed (791 assertions), no regressions.
- `vendor/bin/pest tests/panel/Feature/Settings/StaffTest.php` — 13 passed (98 assertions), unaffected (separate Inertia-based staff settings flow).


wakqasahmed:
Hi @glennjacobs — noticed this PR doesn't have a reviewer assigned yet — it's been about 6 days, CI is green and it's mergeable. Would you (or whoever's best placed) be able to take a look when you get a chance, or point me to who should? Thanks!

wakqasahmed:
Hi @glennjacobs — circling back on this one — still no reviewer assigned, about 7 days since the last activity. No pressure, just don't want it to get lost — happy to make any changes needed.

glennjacobs (approved):
Thanks @wakqasahmed — great diagnosis and an honest, reproducible test plan. I verified the bug independently (reverted the resource against the new regression test — it fails with the FQCN exactly as described) and ran the full admin suite, PHPStan and Pint locally, all green. The `class_uses_recursive` fix in `PermissionSelector` is genuinely required too, since a resolved staff model can inherit `HasRoles` rather than declare it directly.

I've pushed one small commit to the branch: `getModel()` now returns `static::$model ?? config('lunar.staff.model', Staff::class)` so a downstream subclass that sets the `$model` property keeps Filament's normal override semantics.

Two non-blocking notes for future reference:

- The new test hardcodes the `model_has_roles` table name — `config('permission.table_names.model_has_roles')` would track the Spatie config if it's ever customised.
- The `PermissionSelector` fix is only covered transitively via the admin EditStaff flow; a small `getAuthorizeRecord()` unit test in the filament suite with a record that inherits `HasRoles` would pin it in the package that owns it.

PR #2603, opened 2026-08-08
fix(core): scope CacheInvalidator lifetime and clear rolled-back pending entries

### LINKED ISSUE #2561: [2.0-alpha4] CacheInvalidator can retain rolled-back entries across later work
- Laravel Version: 13.19.0
- PHP Version: 8.5.7
- Database Driver & Version: 11.8.6

I found a lifecycle mismatch in `CacheInvalidator`. The class describes itself as request-scoped, but `LunarServiceProvider` binds it as a singleton.

More importantly, recording inside a transaction adds the target to `$pending` and registers an `afterCommit` callback. A full rollback discards that callback, but the pending entry is not cleared. A later autocommit mutation or committed transaction can therefore flush an invalidation for a change that was rolled back. A later transaction for the same target can also skip callback registration because the stale key is still considered present.

The singleton lifetime increases the impact in long-lived workers and Octane, where abandoned state may survive beyond the request or job that created it.

### How to reproduce

1. Begin a transaction.
2. Update a model that participates in cache invalidation.
3. Roll back the transaction.
4. Update another participating model outside a transaction.
5. Observe that the second update calls `flush()` and can emit the entry retained from step 2.

The existing rollback test checks only that no event is emitted immediately after rollback. It does not perform a later mutation that exposes the retained entry.

### Suggested resolution

The binding should match the documented lifecycle, and pending state should be associated with its connection and transaction frame so a rollback discards only the affected entries.

This does not appear to have a safe one-block fix. Changing `singleton` to `scoped` limits cross-request or cross-job leakage but does not fix rollback followed by later work in the same scope. Clearing the entire pending array in a generic rollback callback can also remove valid outer-transaction entries during nested transaction handling.

### Regression coverage

Please add tests for:

- root rollback followed by an autocommit mutation;
- root rollback followed by a later committed transaction;
- a later transaction for the same target;
- nested rollback followed by outer commit;
- the same target appearing in outer and nested transaction frames;
- lifecycle isolation in a long-lived process;
- unchanged one-event-per-target commit deduplication.

### Pull request

I can provide a pull request fixing the recorder lifetime and rollback-state handling and adding the regression coverage above. I would first align the implementation with the maintainers' preferred nested-transaction semantics. 

I wanted to run this by you all first before I submit a PR to make sure this is accurate.

Relevant source:
`packages/core/src/Cache/CacheInvalidator.php`,
`packages/core/src/LunarServiceProvider.php`, and
`tests/core/Feature/CacheInvalidationTest.php`.

wakqasahmed:
Closes #2561

## Summary

`CacheInvalidator` is documented as request-scoped but was bound as a `singleton`, letting pending state outlive a request/job under Octane or long-lived workers. It's now bound with `$app->scoped(...)`.

Separately, `record()` added targets to `$pending` and registered an `afterCommit` callback, but a rollback discarded the callback while leaving the `$pending` entry in place. A later autocommit mutation, a later committed transaction, or a later transaction touching the same target could then flush an invalidation event for a change that was rolled back.

The fix tracks the transaction level each pending entry was first (or most shallowly) touched at, and listens for `Illuminate\Database\Events\TransactionRolledBack` per connection to drop only the entries whose shallowest touch happened inside the frame being discarded. An entry also touched by a surviving outer frame is kept (accepted over-invalidation, consistent with the class's existing reliability bias documented in its docblock).

## Changes

- `packages/core/src/LunarServiceProvider.php`: bind `CacheInvalidator` as `scoped` instead of `singleton`.
- `packages/core/src/Cache/CacheInvalidator.php`: record the transaction level per pending entry and register a rollback listener that prunes entries whose frame was discarded.
- `tests/core/Feature/CacheInvalidationTest.php`: added regression coverage for
  - root rollback followed by an autocommit mutation
  - root rollback followed by a later committed transaction
  - a later transaction for the same target after a rollback
  - nested rollback followed by outer commit
  - the same target touched in both outer and nested frames surviving a nested rollback

## Test plan

- [x] Reproduced the bug: reverting the source changes while keeping the new tests fails `a root rollback does not leak into a later autocommit mutation` (event dispatched 2 times instead of 1).
- [x] All 25 tests in `tests/core/Feature/CacheInvalidationTest.php` pass with the fix applied (ran via `vendor/bin/pest`, PHP 8.4, sqlite).


wakqasahmed:
Hi @glennjacobs @alecritson — noticed this PR doesn't have a reviewer assigned yet — it's been about 6 days, CI is green and it's mergeable. Would you (or whoever's best placed) be able to take a look when you get a chance, or point me to who should? Thanks!

glennjacobs:
@wakqasahmed on holiday at the moment. Will be hitting all the PRs on my return at the end of next week.

wakqasahmed:
Sounds good, thanks — no rush, I'll wait for your review whenever you're back.

PR #2612, opened 2026-08-23
fix invalid countries and invalid states

xlvis7:
Fixes Shipping Zone create and update validation error:
https://github.com/lunarphp/lunar/issues/2618

1. "The selected countries is invalid." error when "Limit to Countries".
2. "The selected states is invalid." error when "Limit to States".

image
image


wychoong:
Few points
- starting filament v4 Select also validates input with options, hence this error
- As brought up by the referred issue, removing `getOptionLabelsUsing` should solve it as it’s redundant with `options` and root of the issue
- Check for other similar issue

PR #2615, opened 2026-08-23
fix(upgrade): reconcile v1 soft-deleted records onto v2 state

### LINKED ISSUE #2613: [2.x] lunar:upgrade leaves v1 soft-deleted records visible after upgrade (dropped SoftDeletes not reconciled)
- Laravel Version: all
- PHP Version: all
- Database Driver & Version: all

### Expected Behaviour:

`lunar:upgrade` should reconcile the models that dropped the `SoftDeletes` trait in v2 — `Product`, `ProductVariant`, `Channel`, `Collection` (`Cart` and `Staff` keep it). v2 replaced soft-deletion with explicit state: products/collections retire via `status = archived`, channels via `status = inactive`, and variants hide via `enabled = false`. The upgrade should map every v1 soft-deleted row (`deleted_at IS NOT NULL`) onto that state and drop the now-orphaned `deleted_at` column, so an upgraded database matches a fresh v2 install and previously soft-deleted records stay hidden.

Rows must be kept, not hard-deleted: historical order lines reference retired variants and must still resolve their purchasable.

### Actual Behaviour:

The upgrade carries the v1 `deleted_at` column and every soft-deleted row forward untouched, and never maps them onto the v2 state:

- `2026_06_01_000012_add_product_variant_enabled` backfills `enabled = true` for **every** variant, including v1 soft-deleted ones.
- Products/collections keep their old `status` (often `published`); channels keep `active`.
- The `deleted_at` column remains on all four tables, which a fresh v2 install does not have.

As a result, records a merchant soft-deleted in v1 resurface as **live / enabled** catalogue entries after upgrading, and the upgraded schema diverges from a fresh install.

### Steps To Reproduce:

1. On a Lunar 1.x store, soft-delete a product and a product variant (each gets a `deleted_at`).
2. Run `php artisan lunar:upgrade`.
3. Inspect the upgraded database:
   - `product_variants.enabled = 1` for the soft-deleted variant, `deleted_at` still populated.
   - `products.status` unchanged (e.g. `published`) for the soft-deleted product, `deleted_at` still populated.
4. The soft-deleted product / variant now appears as available in the catalogue / storefront.

omarelshopky:
### Summary

Fixes #2613 . Reconciles the models that dropped the `SoftDeletes` trait in v2 during `lunar:upgrade`.

v2 removed `SoftDeletes` from `Product`, `ProductVariant`, `Channel` and `Collection` (`Cart` and `Staff` keep it), replacing soft-deletion with explicit state. But the upgrade left the v1 `deleted_at` column and its rows untouched — and `000012_add_product_variant_enabled` marks every variant `enabled = true` — so records a merchant soft-deleted in v1 resurface as live / enabled catalogue entries after upgrading, and the upgraded schema keeps a `deleted_at` column that a fresh install does not have.

### Changes

- Adds `packages/upgrade/database/migrations/2026_06_01_000013_reconcile_dropped_soft_deletes.php`. For each dropped-`SoftDeletes` model it maps every `deleted_at IS NOT NULL` row onto the v2 hidden state, then drops the orphaned `deleted_at` column so the upgraded schema matches a fresh install:

  | Model | v2 hidden state |
  |---|---|
  | `product_variants` | `enabled = false` |
  | `products` | `status = 'archived'` |
  | `collections` | `status = 'archived'` |
  | `channels` | `status = 'inactive'` |

- Guarded per table on the presence of `deleted_at`, so re-runs, already-v2 databases, and the `SoftDeletes`-keeping `Cart` / `Staff` tables are untouched.
- Rows are **kept, not hard-deleted**: historical order lines reference retired variants and must still resolve their purchasable.
- One-way, like the other upgrade data steps (no `down()`). Runs after `000012` (which adds the `enabled` column), auto-discovered by `DataMigrationStep`.

### Tests

Adds `tests/upgrade/Feature/ReconcileDroppedSoftDeletesTest.php`: asserts every model maps its soft-deleted rows onto the v2 state while leaving live rows untouched and dropping `deleted_at`, plus a fresh-schema no-op case.


glennjacobs (changes requested):
Thanks for this — the diagnosis in #2613 is spot on, and half of this PR is exactly right. The `products` and `product_variants` halves match what spec 0021's migration-impact section promised (trashed → `archived` / `enabled = false`, rows kept so order lines still resolve their purchasable, one-way with no `down()`, per-table guards, prefix handling). The mapped values all check out against the real state machines. But verifying against the actual v1 schema turned up a serious problem with the other half.

## Silent data loss on real upgrades — channels and collections

v1's create migrations give `channels` and `collections` `softDeletes()` but **no `status` column**, and nothing in the upgrade path adds one (spec 0021 promised stage-3 migrations for this, but only the orders part ever landed as `000009_backfill_order_statuses_and_fulfilments`). Since the ledger rewrite marks the v2 baseline as already-run, an upgraded database reaches your migration with no `status` on either table.

That means the `Schema::hasColumn($table, $column)` guard silently skips the mapping for both tables — and then drops `deleted_at` anyway, destroying the only record of which channels/collections were retired. Every v1 soft-deleted channel and collection resurfaces as live, unrecoverably.

The tests pass because `simulateV1SoftDeletableTables()` creates the tables *with* a `status` column — a schema the real upgrade path never produces.

## Suggested shape

Mirror the existing `000012`-then-reconcile pattern:

1. **New `2026_06_01_000013_add_collection_and_channel_status.php`** — add `status` to both tables matching the v2 baseline definitions (`string, indexed; default 'draft'` for collections, `'active'` for channels), then backfill existing rows: collections → `'published'` (v1 collections were not lifecycle-gated, so live ones must not land on the `'draft'` default — that would hide a merchant's entire collection tree post-upgrade), channels → `'active'`. Note spec 0021's "map v1 `channels.enabled`" line was written on a false premise — v1's `enabled` lives on the `channelables` pivot, not `channels`, so there is nothing to map from; trashed → `inactive`, everything else `active` is the whole story.
2. **This migration becomes `000014`** and can drop the `hasColumn` guard on the hidden column entirely — with the columns guaranteed by 000013, the map-then-drop hazard disappears structurally. (Keep the `deleted_at` presence guard for re-runs and fresh installs.)
3. **Tests simulate the true v1 shape** — no `status` on channels/collections — and assert the full 000013 → 000014 chain: live collections end `'published'`, trashed `'archived'`; live channels `'active'`, trashed `'inactive'`; `deleted_at` gone everywhere.

One out-of-scope note, no action needed here: v1 `products.status` was free-form, so custom values other than `published`/`draft` flow through unmapped into the `ProductState` cast — we'll track that separately.

Happy to re-review as soon as this is updated — the products/variants halves and the overall structure are good to go.

omarelshopky:
Thanks for the thorough review — you're right, and it's a serious one.

Fixed exactly along the shape you suggested:

- **`000013_add_collection_and_channel_status`** — adds `status` to both tables matching the v2 baseline definitions (string, indexed; default `draft` for collections, `active` for channels), then backfills existing rows: collections to `published` (so live ones don't inherit the `draft` default and hide the tree), channels to `active`. I took your note on the channel side — v1's enable/disable is on the `channelables` pivot, not `channels`, so there's nothing to map from; the reconcile step just marks trashed channels `inactive`.
- **`000014_reconcile_dropped_soft_deletes`** (renamed from 000013) — with the columns now guaranteed by 000013 (plus `products.status` from v1 and `product_variants.enabled` from 000012), it drops the per-column guard entirely, so a missing column fails loud instead of silently dropping `deleted_at`. The `deleted_at` presence guard stays for re-runs and fresh installs.
- **Tests** now stand up the true v1 shape (no `status` on channels/collections) and assert the full 000013 -> 000014 chain: live collections end `published` / trashed `archived`, live channels `active` / trashed `inactive`, live products unchanged / trashed `archived`, variants enabled / trashed disabled, `deleted_at` gone everywhere.

Verified: the `upgrade` suite is green, and a full `lunar:upgrade` run against a real v1 production snapshot now adds `status` to both tables, lands live rows on `published`/`active`, and drops `deleted_at` with no data loss.

glennjacobs (approved):

PR #2622, opened 2026-08-24
fix(admin): keep record-url resolvers out of config as closures

alecritson:
Any application running the admin shell cannot cache its config:

```
$ php artisan config:cache
Your configuration files could not be serialized because the value at
"lunar.filament.record_urls.order" is non-serializable.
```

`LunarPanelProvider::registerBridgeRecordUrls()` writes three closures into config at boot. A single closure anywhere in the tree makes the **entire** application config unserializable, so this disables config caching for the whole app, not just for Lunar's keys — and any deploy script running `config:cache` fails outright.

## The contract was already documented

`packages/filament/config/filament.php` describes the intended design:

> Each entry is a Filament resource or page class string. The bridge calls `{class}::getUrl(['record' => $record])` when present; an empty string disables the link.

But `RecordUrls::for()` only ever handled callables, and the admin provider only ever set closures. This makes the implementation match the documented contract.

## What `RecordUrls::for()` now accepts

| Config value | Behaviour |
| --- | --- |
| `null` | bridge skips the link |
| `Page::class` | `Page::getUrl([...$context, 'record' => $record])` |
| `[Resource::class, 'edit']` | `Resource::getUrl('edit', [...$context, 'record' => $record])` |
| `callable` | `$resolver($record, $context)` — unchanged |

The array form is matched **before** `is_callable`, because `[Resource::class, 'edit']` would satisfy `is_callable` if the resource happened to expose a static `edit()`.

Callable support is retained, so this is additive: any consumer already configuring their own closure keeps working, and only the shipped defaults change shape. Malformed values fall through to `null` rather than fataling.

## Tests

`tests/filament/Unit/Support/RecordUrlsTest.php` keeps its three existing cases (including the callable one) and adds four: page class string with context, resource class plus page name, malformed values returning null, and a `var_export` check — that being what `config:cache` uses, and what throws on a closure.

Suites run: `filament` 49 passed, `admin` 235 passed, `panel` 739 passed, `core` 1174 passed / 1 skipped. `phpstan analyse` no errors. Pint clean.

## Verified end to end

On a Lunar 2.x storefront that previously failed, `php artisan config:cache` completes and the panel's order, variant and collection links still resolve.

### DIFF
--- a/packages/filament/src/Support/RecordUrls.php
+++ b/packages/filament/src/Support/RecordUrls.php
@@ -7,20 +7,49 @@ namespace Lunar\Filament\Support;
  *
  * Bridge components link out to per-record pages that live in the consuming
  * panel — `lunarphp/admin` by default, or any Filament panel the consumer
- * builds. Each key under `lunar.filament.record_urls.{key}` is a closure
- * receiving the record and returning a URL, or null when the bridge should
- * skip the link.
+ * builds. Each key under `lunar.filament.record_urls.{key}` says how to reach
+ * that page:
+ *
+ *   null                        the bridge skips the link
+ *   Page::class                 Page::getUrl([...$context, 'record' => $record])
+ *   [Resource::class, 'edit']   Resource::getUrl('edit', [...$context, 'record' => $record])
+ *   callable                    $resolver($record, $context)
+ *
+ * The class-string forms are the ones to reach for: config has to survive
+ * `config:cache`, and a closure anywhere in the tree makes the whole of the
+ * application's config non-serializable, not just this key. The callable form
+ * stays supported for consumers already passing one, and for the rare resolver
+ * that genuinely cannot be expressed as a class and a page name.
  */
 class RecordUrls
 {
+    /**
+     * @param  array<string, mixed>  $context
+     */
     public static function for(string $key, mixed $record, array $context = []): ?string
     {
         $resolver = config('lunar.filament.record_urls.'.$key);
 
-        if (! is_callable($resolver)) {
-            return null;
+        $parameters = [...$context, 'record' => $record];
+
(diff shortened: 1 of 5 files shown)


packages/filament/src/Support/RecordUrls.php
```diff
@@ -7,20 +7,49 @@
...
+        $parameters = [...$context, 'record' => $record];
+
+        if (is_string($resolver) && method_exists($resolver, 'getUrl')) {
+            return $resolver::getUrl($parameters);
+        }
+
+        // Checked before is_callable: [Resource::class, 'edit'] would satisfy
+        // is_callable if the resource happened to expose a static edit().
+        if (is_array($resolver) && count($resolver) === 2) {
+            [$class, $page] = $resolver;
```
glennjacobs: A two-element **associative** array — e.g. `['class' => Foo::class, 'page' => 'edit']`, a plausible consumer mistake — passes the `count($resolver) === 2` check but then fatals on the list destructure with `Undefined array key 0`, rather than falling through to `null` as described.

```suggestion
        if (is_array($resolver) && array_is_list($resolver) && count($resolver) === 2) {
            [$class, $page] = $resolver;
```

With the guard in place an associative array falls through to the `is_callable` check (false for an assoc array) and returns `null`. Might be worth adding `['class' => 'Foo', 'page' => 'edit']` to the "keeps working when the configured value is nonsense" test to cover it.

glennjacobs (commented):
Nice fix — implementing the documented contract in `RecordUrls` rather than just reshaping the shipped defaults means third-party panels get config-cache safety too. One small gap in the "malformed values fall through to null" guarantee, flagged inline.

glennjacobs:
@alecritson good to go?

Issue #2517, opened 2026-06-20
Several core relationships lack deterministic ordering — row order shifts on PostgreSQL


alecgarcia:
### Describe the bug

Several core relationships are plain `hasMany`/`belongsToMany` with no `ORDER BY`. Per the SQL standard, row order is then undefined. MySQL/InnoDB happens to return clustered primary-key order, so this has gone unnoticed — but PostgreSQL (listed as officially supported, 9.4+) returns heap/tuple order, which **changes after an `UPDATE`**. The result is user-visible instability when running on PostgreSQL.

Affected relationships:

- **`Cart::lines()`** — after `Cart::updateLine()` (which runs `refresh()->recalculate()`), lines reload in a different order. This is more than cosmetic: `GenerateFingerprint` reduces `$cart->lines` *in iteration order*, so an unstable order can produce a **different cart fingerprint** for the same logical cart after an unrelated update, affecting draft-order matching.
- **`Order::lines()`** — invoice/order-history display sequence (and the derived `physicalLines`/`digitalLines`/`shippingLines`/`productLines`) becomes nondeterministic.
- **`ProductVariant::values()`** — `getOption()` joins these values into the variant option label (e.g. `"Red, Large"`), which is snapshotted onto order lines in `CreateOrderLines`. The persisted label can therefore differ across engines and shift after updates on PostgreSQL.

### To reproduce (PostgreSQL)

1. Create a cart with ≥2 lines; call `updateLine()` on an earlier line; observe `$cart->lines` ordering (and `$cart->fingerprint()`) change.
2. Create a variant with multiple option values; observe `getOption()` ordering vary.

### Expected behavior

Line and option-value order is stable and deterministic across database engines.

### Environment

- Lunar: 1.x
- Database: PostgreSQL 9.4+ (officially supported)
- PHP: 8.x

An accompanying PR with failing tests + fixes is attached.

PR #2633, opened 2026-08-25
Include the selected shipping option in the cart fingerprint

kha333n:
A shopper reaches payment, switches from express to standard delivery in
another tab, and pays. `checkFingerprint()` reports the cart as unchanged,
because the fingerprint does not cover the delivery choice.

### What happens now

- Two carts that differ only in their selected shipping option hash
  identically.
- `Cart::checkFingerprint()` — the documented "have the contents changed?"
  guard — returns `true` after a shipping change, so a checkout that trusts it
  proceeds on a stale amount.
- Every other input to the order total *is* covered: lines, quantities,
  user, currency and coupon. Shipping is the one gap.
- It goes unnoticed because `currentDraftOrder()` also compares `total`, so the
  draft-order reuse path happens to be protected. The exposed path is any
  storefront using `checkFingerprint()` directly.

### What should happen

- Changing the shipping option changes the fingerprint.
- `checkFingerprint()` reports a change whenever anything that moves the order
  total has moved.

### Why

`GenerateFingerprint` hashes the lines, then appends the remaining cart-level
inputs:

```php
$value .= $cart->user_id.$cart->currency_id.$cart->coupon_code;
```

The selected shipping option lives on the cart's shipping address
(`cart_addresses.shipping_option`) and is never read. `ApplyShipping` resolves
the cart's shipping cost from exactly that column, so it is the input that
decides the shipping charge — and it is the only total-affecting field the
fingerprint omits.

### The fix

Append the shipping address's `shipping_option` to the hashed value. It is a
persisted column, so the fingerprint stays a pure function of stored cart
state — deliberately *not* the shipping total, which is transient and only
exists after a calculate. Carts with no shipping address are unaffected, since
the null-safe read contributes nothing.

One compatibility note: fingerprints already stored on orders will not match a
freshly generated one for carts that have a shipping address. That is inherent
to changing the hash inputs, and only affects an in-flight checkout at the
moment of upgrade.

Not related to #2517 / #2518, which also concern the fingerprint but through
line *ordering* on PostgreSQL; that fix is in `Cart::lines()` and does not
touch this action.

### Tests

`tests/core/Unit/Actions/Carts/GenerateFingerprintTest.php`:

- **can generate a different cart fingerprint when the shipping option
  changes** — one cart, one line, shipping option moved from `BASDEL` to
  `EXPDEL`. Fails on `1.x` with
  `Expecting '5411acac202fe57320451f405edc3…a9d85d' not to be
  '5411acac202fe57320451f405edc3…a9d85d'`.

The existing `can generate cart fingerprint` test is unchanged and still
passes, so the covered inputs still behave as before.

### DIFF
--- a/packages/core/src/Actions/Carts/GenerateFingerprint.php
+++ b/packages/core/src/Actions/Carts/GenerateFingerprint.php
@@ -21,7 +21,10 @@ class GenerateFingerprint
                 $line->subTotal;
         });
 
-        $value .= $cart->user_id.$cart->currency_id.$cart->coupon_code;
+        $value .= $cart->user_id.
+            $cart->currency_id.
+            $cart->coupon_code.
+            $cart->shippingAddress?->shipping_option;
 
         return sha1($value);
     }
--- a/tests/core/Unit/Actions/Carts/GenerateFingerprintTest.php
+++ b/tests/core/Unit/Actions/Carts/GenerateFingerprintTest.php
@@ -3,6 +3,7 @@
 use Illuminate\Foundation\Testing\RefreshDatabase;
 use Lunar\Actions\Carts\GenerateFingerprint;
 use Lunar\Models\Cart;
+use Lunar\Models\CartAddress;
 use Lunar\Models\Channel;
 use Lunar\Models\Currency;
 use Lunar\Models\Price;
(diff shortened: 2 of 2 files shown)


glennjacobs:
Thanks — the analysis of what's hashed is accurate, but this changes the fingerprint's contract. It exists to detect whether the cart *contents* changed underneath a checkout (see the docblock on `Cart::fingerprint()`), and the shipping option is checkout-managed state: including it would make `checkFingerprint()` throw when the checkout itself sets or changes the option mid-flow. The stale-total case is covered on the draft-order path, which compares `total` as well. Stores that want stricter semantics can swap the generator via `lunar.cart.fingerprint_generator`. Appreciate the well-tested PR regardless — the BuyXGetY and tax-zone fixes have been great.

PR #2701, opened 2026-08-29
chore(panel): bump npm packages to 0.1.1

glennjacobs:
## What

Bumps `@lunarphp/panel` and `@lunarphp/panel-vite-plugin` from `0.1.0` to `0.1.1`. No content changes.

## Why

`0.1.0` was published manually today; the OIDC trusted-publishing path in `publish_npm.yml` has never actually published. This bump makes the next release tag (`2.0.0-alpha.5`) run the full automated publish for both packages — provenance included — so the pipeline gets its first end-to-end validation on a version where nothing is at stake.

The example add-on's `^0.1.0` ranges resolve to `0.1.1` without changes.

### DIFF
--- a/packages/panel/resources/package/package.json
+++ b/packages/panel/resources/package/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@lunarphp/panel-vite-plugin",
-    "version": "0.1.0",
+    "version": "0.1.1",
     "type": "module",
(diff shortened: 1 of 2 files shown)


glennjacobs:
Superseded by #2703 — the drift guard's first run revealed the 0.1.1 bump is genuinely required (stale FlashMessage.vue.d.ts in the published 0.1.0), so the bump moved there to keep the fix and its version bump atomic.

PR #2703, opened 2026-08-29
feat(repo): CI guard against unpublished npm package drift

glennjacobs:
## What

A CI check that makes it impossible to forget bumping `@lunarphp/panel` / `@lunarphp/panel-vite-plugin` versions when their published content changes.

`scripts/check-npm-drift.mjs` (run via `npm run check-npm-drift` from the monorepo root) is stateless — the registry is the record of what has shipped:

1. Read the version from the package's `package.json`.
2. If that version is **not** on the registry → pass (it publishes on the next tag).
3. If it **is** published → `npm pack` the current tree and compare it file-by-file (sha256) against the published tarball. Any added, removed, or changed file fails with a message naming the file(s) and asking for a version bump.

Runs in the existing `panel-js` job after the panel build (which generates the `@lunarphp/panel` types that `npm pack` includes), so it executes on every PR, every push to `2.x`, and the daily sweep — the daily run doubles as drift detection for toolchain-induced `.d.ts` changes, since `package-lock.json` is unversioned.

## Verified locally

- Clean state: both packages **pass** — current tree byte-matches the published `0.1.0` tarballs (incidentally confirming the manual first publish shipped the READMEs).
- Tampered `vite-plugin.js`: **fails** with `differs: vite-plugin.js` and exit 1.
- Version set to unpublished `0.9.9`: **passes** via the not-yet-on-registry path.

Together with `publish_npm.yml`'s skip-if-already-published guard, the invariant is now machine-checked from both sides: publishes never clobber an existing version, and an existing version's content can never silently drift from the repo.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -186,3 +186,9 @@ jobs:
       # preset (IIFE output, externalised vue/inertia/panel globals) still works.
       - name: Build example add-on bundle
         run: npm run build --workspace @lunarphp/panel-addon-example
+
+      # Fails when a publishable npm package's content no longer matches the
+      # tarball published under its current version — the reminder to bump.
+      # Relies on the panel build above for fresh @lunarphp/panel types.
+      - name: Check npm packages for unpublished drift
+        run: npm run check-npm-drift
--- a/package.json
+++ b/package.json
@@ -6,5 +6,8 @@
         "packages/panel/resources/package",
         "packages/panel/resources/panel-package",
         "packages/panel-addon-example"
-    ]
+    ],
+    "scripts": {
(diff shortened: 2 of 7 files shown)


glennjacobs:
The guard's first CI run caught a real bug — see the follow-up commits: `build:types` never cleaned `dist/`, so the published 0.1.0 ships a stale `FlashMessage.vue.d.ts` for a component removed in #2566, and the example add-on still imported/rendered `FlashMessage` (undefined at runtime; vite builds don't type-check). Both fixed here, and the 0.1.1 bump from #2701 is folded in since the tarball content now legitimately changes.

PR #2713, opened 2026-09-03
fix(upgrade): drop morph indexes before columns and baseline 000071

wychoong:
Stacked on #2712.

The new upgrade-path workflow currently fails on every driver. This PR is the package-side fix so that workflow can go green.

## Attribute morph indexes

v1 indexes `attribute_groups.attributable_type` and `attributes.attribute_type` (plus unique `(attribute_type, handle)`). `2026_06_01_000003` drops those columns without dropping the indexes first.

- SQLite: `DROP COLUMN` fails while the index remains
- MariaDB: error 1072 (`Key column doesn't exist in table`)
- MySQL 8 / Postgres: the data step already succeeds

Drop the indexes (when present) immediately before `dropColumn`.

## Product-types tax-class FK

`2026_06_01_000011` adds `product_types.default_tax_class_id` **and** its foreign key. Core `2026_01_01_000071` only adds the constraint, for fresh installs whose create migration cannot reference `tax_classes` yet.

`000071` is missing from `ledger.v2_baseline`, so after ledger-rewrite `php artisan migrate` runs it again and MySQL/Postgres error on a duplicate FK. Same treatment `000059` already gets. This is the one-line change from #2605.

## Test plan

- [x] Parent workflow in #2712 reproduces both failures on a vanilla Laravel 13 + `lunarphp/core` 1.5.0 app
- [ ] Re-run that workflow on this branch (sqlite / mysql / postgres / mariadb)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/upgrade/database/migrations/2026_06_01_000003_convert_attribute_data_keys_to_ids.php
+++ b/packages/upgrade/database/migrations/2026_06_01_000003_convert_attribute_data_keys_to_ids.php
@@ -24,9 +24,9 @@ use Lunar\Core\Facades\DB;
  * 6. Convert `attributes.validation_rules` from the v1 pipe-delimited string
  *    (`min:1|max:10`) to the v2 json list (`["min:1", "max:10"]`) — split on
  *    `|` exactly as Laravel's validator parses string rules.
- * 7. Add `attribute_groups.system`, drop the morph columns and unused
- *    `section` / `default_value`, make `attributes.attribute_group_id`
- *    nullable.
+ * 7. Add `attribute_groups.system`, drop the morph columns (and the v1
+ *    indexes that still sit on them) and unused `section` / `default_value`,
+ *    make `attributes.attribute_group_id` nullable.
  *
  * One-way — `down()` is intentionally absent. Restore from backup to reverse.
  */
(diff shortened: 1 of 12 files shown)


packages/upgrade/src/Steps/LedgerRewriteStep.php
```diff
@@ -107,6 +110,23 @@ public function run(StepContext $context): void
...
+     */
+    protected function baselineSchemaPresent(string $migration, ?string $connection): bool
+    {
+        if ($migration !== '2026_01_01_900000_create_staff_table') {
+            return true;
+        }
+
+        return Schema::connection($connection)->hasTable(
+            config('lunar.database.table_prefix').'staff',
+        );
```
wychoong, line 127: in v1 staff is in panel package, so there is a chance someone using lunarphp for core without panel

packages/upgrade/database/migrations/2026_06_01_000003_convert_attribute_data_keys_to_ids.php
```diff
@@ -406,6 +415,35 @@ protected function reshapeAttributesSchema(): void
         }
     }
 
+    /**
+     * Drop a v1 index by its live name. `Schema::hasIndex(..., 'index')` misses
+     * SQLite, which reports the type as `btree` rather than `index`.
```
wychoong: seems a few more sqlite specific issue, not sure worth handling for upgrade command as unlikely anyone using it for prod

PR #2723, opened 2026-09-09
fix(core): register a model observer once for a replaced model

### LINKED ISSUE #2722: Observers registered by a trait boot run twice for a replaced model
### Which release line(s) does this affect?

1.x — confirmed. 2.x — not reproduced, see the note at the end.

### Lunar version

lunarphp/core 1.5.0

### PHP version

8.4.25

### Laravel version

12.69.2

### Database driver & version

MySQL 8 in development, MariaDB 10.6 in production. Reproduced in the core test suite on SQLite — not database-related.

### Expected behaviour

A single save on a model replaced through `ModelManifest::replace()` runs each observer once: one `MakeSearchable` job per save with Laravel Scout, one `activity_log` row per save with spatie/laravel-activitylog.

### Actual behaviour

Every observer registered by a trait boot runs twice per save.

`HasModelExtending::fireModelEvent()` dispatches each event under the concrete class *and* under the Lunar model it replaces. Both classes boot their traits — the Lunar model gets booted by `HasModelExtending::observe()` (`new static`), which `LunarServiceProvider::registerObservers()` calls on the Lunar classes — so Scout's `Searchable::bootSearchable()` registers `ModelObserver` for both class names, and nothing deduplicates the two.

Listener state after a replaced order model has booted:

```
eloquent.saved: Lunar\Models\Order          => 1 Laravel\Scout\ModelObserver@saved
eloquent.saved: App\Models\Order            => 1 Laravel\Scout\ModelObserver@saved
```

Nothing is corrupted — the same work is done twice, and it scales with write volume:

- **Scout**: two `MakeSearchable` jobs per save. In our shop a single Mollie checkout queued ten of them for one unpaid draft order: the order-creation pipeline saves the order five times (`FillOrderFromCart` twice, `MapDiscountBreakdown`, `MapAppliedCoupons`, plus a project-specific step), each save doubled.
- **spatie/laravel-activitylog**: two `activity_log` rows per save, visible as consecutive id pairs for one subject. `Lunar\Models\Product` sidesteps the symptom with `$enableLoggingModelsEvents = false`.

### Minimal reproduction steps

1. Register a replacement for the order model — any subclass of `Lunar\Models\Order` will do; the core test suite's `Lunar\Tests\Core\Stubs\Models\Order` is enough:
   ```php
   ModelManifest::replace(
       Lunar\Models\Contracts\Order::class,
       Lunar\Tests\Core\Stubs\Models\Order::class
   );
   ```
2. Enable queued Scout syncing and fake the queue: `config(['scout.queue' => true]); Queue::fake();`
3. Create one order from the replacement: `Lunar\Tests\Core\Stubs\Models\Order::factory()->create();`
4. Observe `Queue::assertPushed(MakeSearchable::class, 1)` failing: *pushed 2 times instead of 1 time*. The same `create()` on a model that is **not** replaced pushes one job, so the duplication is specific to replacement. Deduplicating the observer registration brings the replaced model back to one job, which also rules out the factory simply saving twice.

---

**Note on `2.x`** — `main` rewrote the trait: `registerModelEvent()` and `fireModelEvent()` both canonicalise the event name through `ModelManifest::getMorphClassBaseModel()`. Both boots then register onto the *same* event name, so by reading the code the duplication looks unchanged rather than fixed. I have not reproduced it there — please treat that as a hypothesis, not a report.

**Suggested fix** — make `observe()` idempotent: skip an observer already listening to the resolved model's events, under either its own class name or the Lunar model's, for every event it would be registered for. That keeps the double dispatch, and with it the documented behaviour that observers referencing the Lunar model keep working. A pull request against `1.x` with a failing test and this fix follows.

YannikFirre:
Closes #2722

### The bug

When a model is replaced through `ModelManifest::replace()`, an observer registered by a trait
boot is attached to both the replacement and the Lunar model it replaces, and therefore runs
twice for a single save.

`fireModelEvent()` dispatches each event under the concrete class *and* under the Lunar model.
Both classes boot their traits — the Lunar model is booted by `observe()` itself (`new static`),
which `LunarServiceProvider::registerObservers()` calls on the Lunar classes — so Laravel
Scout's `Searchable::bootSearchable()` registers `ModelObserver` under both names, and nothing
deduplicates them.

Measured: one `create()` on a replaced model pushes two `MakeSearchable` jobs. In our shop a
single Mollie checkout queued ten of them for one unpaid draft order, because the
order-creation pipeline saves the order five times. `spatie/laravel-activitylog` writes two
`activity_log` rows per save for the same reason (`Lunar\Models\Product` sidesteps it with
`$enableLoggingModelsEvents = false`).

### The fix

`observe()` skips an observer class that already listens, under either the resolved model's
class name or the Lunar model's, to *every* event `registerObserver()` would attach it to (same
`method_exists()` filter). The double dispatch is left alone, so observers registered against
the Lunar model keep working as documented. An observer with no matching event is handed to
`registerObserver()` untouched, so an unknown observer still throws Eloquent's *Unable to find
observer* exception.

The expression that resolves the Lunar model for a replacement already lived in
`fireModelEvent()`; it is now `lunarModelClass()` and shared by both call sites.

Two points I'd rather you decided than assumed:

- `Model::observe(SomeObserver::class)` called twice on the same class now registers once. That
  is the same duplicate this fixes, but it is a behaviour change against vanilla Eloquent.
- Should the fix instead prevent the Lunar model from booting at all when a replacement is
  registered? That removes the duplicate at its source, but only when `ModelManifest::replace()`
  runs before any `observe()` call on those models, so it does not hold in every boot order.

### Tests

`tests/core/Unit/Base/Traits/HasModelExtendingTest.php`, both failing on `1.x` today:

- *observers registered by a trait boot only run once for a replaced model* — one `create()` on
  a replaced model must queue one `MakeSearchable` job (currently *pushed 2 times instead of 1*).
  The same `create()` on a model that is not replaced already queues one, so the duplication is
  specific to replacement.
- *an observer registered on both the lunar model and its replacement runs once* — covers the
  failure mode the fix could introduce (an observer silently never registered) together with the
  documented forwarding of observers registered against the Lunar model. Uses a new
  `tests/core/Stubs/TestOrderObserver.php`.

Pint and PHPStan pass on the changed files; `tests/core` is green.

### Forward port

`main` rewrote the trait so that `registerModelEvent()` and `fireModelEvent()` both canonicalise
the event name; the two boots then register onto the *same* event name, so from reading the code
the `2.x` line looks affected too. I have not reproduced it there — flagging it for
`needs-forward-port` rather than claiming it.

### DIFF
--- a/packages/core/src/Base/Traits/HasModelExtending.php
+++ b/packages/core/src/Base/Traits/HasModelExtending.php
@@ -5,6 +5,7 @@ namespace Lunar\Base\Traits;
 use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\Relations\Relation;
+use Illuminate\Events\Dispatcher;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Str;
 use Lunar\Base\BaseModel;
@@ -166,8 +167,66 @@ trait HasModelExtending
         }
 
         foreach (Arr::wrap($classes) as $class) {
+            if (static::observerIsAlreadyRegistered($instance, $class)) {
+                continue;
+            }
+
             $instance->registerObserver($class);
         }
+    }
+
+    /**
+     * Whether the given observer already listens to every event it would be registered for.
+     *
+     * Model events are dispatched twice for a replaced model: under its own class name, and
+     * under the Lunar model it replaces (see fireModelEvent()). Both classes boot their traits,
+     * so an observer registered from a trait boot - Scout's ModelObserver, for one - would be
+     * attached to both names and run twice for a single save.
+     */
+    protected static function observerIsAlreadyRegistered(Model $instance, object|string $class): bool
+    {
+        if (! static::$dispatcher instanceof Dispatcher) {
+            return false;
+        }
+
+        $observer = is_object($class) ? $class::class : $class;
+
+        $events = array_filter(
+            $instance->getObservableEvents(),
+            fn (string $event) => method_exists($class, $event)
+        );
+
+        if ($events === []) {
+            return false;
+        }
+
+        $modelClasses = array_unique([
+            $instance::class,
+            static::lunarModelClass($instance::class),
+        ]);
+
+        $listeners = static::$dispatcher->getRawListeners();
+
+        foreach ($events as $event) {
+            $registered = array_merge(...array_map(
+                fn (string $modelClass) => (array) ($listeners["eloquent.{$event}: {$modelClass}"] ?? []),
+                $modelClasses
+            ));
+
+            if (! in_array("{$observer}@{$event}", $registered, true)) {
+                return false;
+            }
+        }
+
+        return true;
+    }
+
+    /**
+     * The Lunar model a replacement re-dispatches its events under.
+     */
+    protected static function lunarModelClass(string $modelClass): string
+    {
+        return str_replace('Contracts\\', '', ModelManifest::guessContractClass($modelClass));
     }
 
     /**
@@ -178,7 +237,7 @@ trait HasModelExtending
         // Fire the actual models events
         $result = parent::fireModelEvent($event, $halt);
 
-        $lunarClass = str_replace('Contracts\\', '', ModelManifest::guessContractClass(static::class));
+        $lunarClass = static::lunarModelClass(static::class);
 
         if ($lunarClass == static::class) {
             return $result;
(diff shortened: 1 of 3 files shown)


wychoong:
1. firing twice is a valid issue
2. however only registering for the extended model is a breaking change for those extended the models and having custom observer that is built on top instead replacing lunar's
3. `2.x` is not affected because model replacement is removed

